PR #117, opened 2021-11-11
Fix device side contains implementation and add tests

vyasr:
The `__device__` implementation of `static_map::device_view::contains` is currently broken because it relies on accessing private parent members. This PR fixes that problem (as well as adding a necessary indirection via `this` since we're in a template context) and adds a test for this function. This PR also makes `contains` a const function.


PointKernel:
Such minor issues can be identified much earlier if we have a comprehensive unit test. Probably another PR updating unit tests to exercise all device-view and host-bulk functions.

vyasr:
Yes of course. In this instance I just added unit tests for the feature whose error I found and I think that's suitable scope here, but we can definitely add more tests to catch additional issues going forward.

jrhemstad (approved):

PR #126, opened 2021-12-11
Make sentinel getters part of public device_view API.

vyasr:
`static_multimap::device_view` currently exposes public `__device__` APIs to access the sentinel values in device code, but the implementation relies on calling corresponding private methods of the `device_view_impl` class. As a result, when actually invoked, these calls fail. This PR makes those APIs public.


PointKernel (approved):
Looks like it's the same issue with other public APIs e.g. `get_slots()`, `get_capacity()`, etc, and these APIs should be exposed to `device_mutable_view` also.

I will make another PR to fix them all.

jrhemstad (approved):

PR #129, opened 2021-12-18
Add clang tidy

PointKernel:
TBD


CMakeLists.txt
```diff
@@ -29,6 +29,9 @@ include(rapids-find)
 # * Enable the CMake CUDA language
 rapids_cuda_init_architectures(CUCO)
 
+# this is needed for clang-tidy runs
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
sleeepyjack, line 33: This also enables us to use other development tools that use the `clangd` language server. Awesome!

benchmarks/synchronization.hpp
```diff
@@ -81,7 +81,9 @@ class cuda_event_timer {
    *                            every iteration.
    * @param[in] stream_ The CUDA stream we are measuring time on.
    */
-  cuda_event_timer(benchmark::State &state, bool flush_l2_cache = false, cudaStream_t stream = 0)
+  cuda_event_timer(benchmark::State& state,
+                   bool flush_l2_cache = false,
+                   cudaStream_t stream = nullptr)
```
sleeepyjack, line 86: The default stream is commonly represented by `(int)0` but I guess `nullptr` is fine, too.

sleeepyjack (approved):
This PR satisfies my inner Monk. Very good!

tests/static_multimap/static_multimap_test.cu
```diff
@@ -83,7 +83,7 @@ __inline__ void test_custom_key_value_type(Map& map,
                                            KeyIt key_begin,
                                            size_t num_pairs)
 {
-  constexpr cudaStream_t stream = 0;
+  constexpr cudaStream_t stream{nullptr};
```
karthikeyann, line 86: I am curious about if we can use `nullptr` instead of `0`.
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#default-stream
> Kernel launches and host <-> device memory copies that do not specify any stream parameter, or equivalently that set the stream parameter to zero, are issued to the default stream. They are therefore executed in order.

benchmarks/synchronization.hpp
PointKernel: `cudaStream_t` is an alias of a pointer. `clang-tidy` will complain if we initialize it with 0.

tests/static_multimap/static_multimap_test.cu
PointKernel: Right, `0` is the common way to initialize CUDA stream while according to the CUDA user guide, `cudaStream_t` is a type alias of `CUstream_st *` (see [here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1ge15d9c8b7a240312b533d6122558085a)) and `clang-tidy` will complain if it's initialized with `0`. @karthikeyann @sleeepyjack For your reference, `rmm` is doing the same in https://github.com/rapidsai/rmm/pull/857.

sleeepyjack:
What's the state of this PR? This could be used to automatically catch bugs such as #510.

PointKernel:
> What's the state of this PR? This could be used to automatically catch bugs such as #510.

I vaguely recall that some upstream libraries couldn't be built with clang. Let me prioritize this work.

PR #131, opened 2022-01-05
Reduce compilation time

### LINKED ISSUE #93: [FEA] Reduce building time
After introducing `rapids.cmake` into the project, building cuco becomes unreasonably expensive for such a small project. More precisely, it takes 11 mins to build the code with 6 concurrent threads. Linking `STATIC_MAP_TEST` and `DYNAMIC_MAP_TEST` are the major time killers.

**Describe the solution you'd like**
Get rid of the dynamic initialization warnings in these two tests and reduce building time.

PointKernel:
Close https://github.com/NVIDIA/cuCollections/issues/93

This PR splits tests/benchmarks into multiple files to reduce build time. It also replaces `thrust` algorithms with user-defined ones. In the end, for one GPU architecture, it reduced the build time from ~265 seconds to about 38 seconds.

Build time (baseline): 
baseline


PointKernel:
<details><summary>Splitting unit tests into multiple files</summary>
<p>


split

</p>
</details>

<details><summary>Using custom logical algorithms instead of thrust</summary>

Replaced `all_of`, `none_of`, `any_of` and `equal`

<p>

custom_func

</p>
</details>

<details><summary>Splitting multimap benchmarks into multiple files</summary>
<p>

split-bench

</p>
</details>

<details><summary>Using custom sort function</summary>
<p>

custom_sort

</p>
</details>

tests/utils.cuh
```diff
@@ -0,0 +1,79 @@
...
+  while (tid < n) {
+    auto cmp = begin1 + tid;
+    auto ref = begin2 + tid;
+    atomicAdd(count, static_cast<int>(p(*cmp, *ref)));
+    tid += gridDim.x * blockDim.x;
+  }
+}
+
+template <typename T, typename Operator>
+__global__ void sort(T* data, int const size, Operator op)
```
jrhemstad: This makes me uneasy. I don't like the idea of having to write and maintain our own sort implementation just for tests.

PointKernel: Right, I was reluctant to do this as well. Despite the downside, using our own sort reduces the build time from 40 seconds to 30 seconds. So we stick with `thrust::sort`?

jrhemstad: What happens if we just use CUB sort directly instead?

PointKernel: Using `cub::DeviceRadixSort::SortKeys` is as expensive as `thrust::sort` (in terms of build time).

elstehle: Saw this by chance. Wondering if the main issue is CUB/thrust's sort template being instantiated for all `<KeyT, ValueT>` combinations or whether even just a single sort template instantiation was bad enough for you. I would have some thoughts if the former was the issue but I'm afraid that just a single instantiation may be bad enough already. Can you try to have it just for a single instance and let me know if that compilation time was still acceptable?

PointKernel: I just tested explicit instantiation of cub sort and didn't see any impact on build time. :(

elstehle: Sorry, I was not very clear. What I actually meant was that **_we_** are instantiating the sort for multiple types (because we test for multiple types). Basically, what I was wondering: how much would it help us if instead of instantiating the sort for all those types, we would have to instantiate it just once (just assuming). For the sake of measuring compile time, you could just hardcode all sort invocations to the same type, e.g., `sort(reinterpret_cast<uint32_t>(probe_keys.data().get(), ...)`.

PointKernel: OK, I see what you mean now. In this case, one single instantiation is already bad enough.

vyasr:
This PR is awesome, but once this is done as a next step we really need to investigate the underlying issue of why the cuco data structures are so slow to compile. The improvements in this PR make building cuco tests more palatable, but any code aiming to incorporate cuco will still find compile times prohibitively slow. We're running into this in cudf now. This PR should make those investigations much easier though!

PointKernel:
@jrhemstad Revert back to use `thrust::sort` for now. 

I figured out another thing we can do is to move benchmarks/tests to `cpp` whenever possible. Will be in a follow-up PR.

PointKernel:
> we really need to investigate the underlying issue of why the cuco data structures are so slow to compile.

@vyasr Probably just because we are passing so many template parameters

jrhemstad (approved):

PR #133, opened 2022-01-18
Move cuCollections to libcudacxx 1.7

robertmaynard:
Previously was on 1.7.0-ea, but can use rapids-cmake now that https://github.com/rapidsai/rapids-cmake/pull/143 has been merged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 #=============================================================================
 cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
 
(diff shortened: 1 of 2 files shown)


PointKernel:
We want to update the `rapids-cmake` version [here](https://github.com/robertmaynard/cuCollections/blob/fea/move_to_libcudacxx_1.7/CMakeLists.txt#L18-L20) also since the old version is still fetching `libcudacxx-1.6`

robertmaynard:
> We want to update the `rapids-cmake` version [here](https://github.com/robertmaynard/cuCollections/blob/fea/move_to_libcudacxx_1.7/CMakeLists.txt#L18-L20) also since the old version is still fetching `libcudacxx-1.6`

Done!

PointKernel (approved):
LGTM

PR #138, opened 2022-01-27
Get rid of `std::move` when using `cuco::make_pair`

PointKernel:
This PR fixes multiple improper uses of `cuco::make_pair` where types were explicitly specified and `std::move` were used. Accordingly, it adds a `cuco::pair_converter` constructor taking universal references and a copy constructor to `cuco::pair`.

The related cudf implementation should be updated as well:
https://github.com/rapidsai/cudf/blob/57ff6f55b9fd44e8a8e10282d3f95d5f38e299ef/cpp/src/join/hash_join.cuh#L71

### DIFF
--- a/include/cuco/detail/pair.cuh
+++ b/include/cuco/detail/pair.cuh
@@ -132,7 +132,10 @@ union pair_converter {
   packed_type packed;
   pair_type pair;
 
-  __device__ pair_converter(pair_type _pair) : pair{_pair} {}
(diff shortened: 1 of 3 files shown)


vyasr:
This looks good. I'm assuming that CI won't pass until #130 is merged though.

PointKernel:
pre-commit.ci run

jrhemstad (approved):

PR #141, opened 2022-03-02
Make all headers self-contained

### LINKED ISSUE #140: [BUG] Headers are not completely self-contained
**Describe the bug**
While working on https://github.com/NVIDIA/cuCollections/pull/139, I noticed that some of the `cuco` headers are not self-contained i.e. being dependent upon other headers being included first. 

**Steps/Code to reproduce bug**
Order the header groups from local to global will introduce compilation errors.

**Additional context**
To fix the issue, we should:
1. reorder the header groups from local to global to uncover potential self-contained issues
2. include what's used and do not rely on transitive inclusions

PointKernel:
Closes #140 

This PR reorders header groups in `cuco` and add missing headers to make sure all headers are self-contained. 

To be merged after #155 and #156.

### DIFF
--- a/benchmarks/hash_table/dynamic_map_bench.cu
+++ b/benchmarks/hash_table/dynamic_map_bench.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,11 +14,14 @@
  * limitations under the License.
  */
 
-#include <benchmark/benchmark.h>
+#include <synchronization.hpp>
+
 #include <cuco/dynamic_map.cuh>
+
+#include <benchmark/benchmark.h>
+
 #include <iostream>
 #include <random>
-#include <synchronization.hpp>
 
 enum class dist_type { UNIQUE, UNIFORM, GAUSSIAN };
 
--- a/benchmarks/hash_table/static_map_bench.cu
+++ b/benchmarks/hash_table/static_map_bench.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,13 +14,16 @@
  * limitations under the License.
  */
 
-#include "cuco/static_map.cuh"
+#include <cuco/static_map.cuh>
+
+#include <thrust/device_vector.h>
+#include <thrust/for_each.h>
+
 #include <benchmark/benchmark.h>
+
 #include <fstream>
 #include <iostream>
 #include <random>
-#include <thrust/device_vector.h>
-#include <thrust/for_each.h>
 
 enum class dist_type { UNIQUE, UNIFORM, GAUSSIAN };
 
--- a/benchmarks/hash_table/static_map_bench.cu
+++ b/benchmarks/hash_table/static_map_bench.cu
@@ -202,63 +202,61 @@ static void BM_static_map_erase_all(::benchmark::State& state)
   state.SetBytesProcessed((sizeof(Key) + sizeof(Value)) * int64_t(state.iterations()) *
                           int64_t(state.range(0)));
 }
-/*
+
 BENCHMARK_TEMPLATE(BM_static_map_insert, int32_t, int32_t, dist_type::UNIQUE)
   ->Unit(benchmark::kMillisecond)
   ->Apply(generate_size_and_occupancy)
   ->UseManualTime();
-  */
 
 BENCHMARK_TEMPLATE(BM_static_map_search_all, int32_t, int32_t, dist_type::UNIQUE)
   ->Unit(benchmark::kMillisecond)
   ->Apply(generate_size_and_occupancy);
-/*
 
 BENCHMARK_TEMPLATE(BM_static_map_insert, int32_t, int32_t, dist_type::UNIFORM)
-->Unit(benchmark::kMillisecond)
-->Apply(generate_size_and_occupancy)
-->UseManualTime();
+  ->Unit(benchmark::kMillisecond)
+  ->Apply(generate_size_and_occupancy)
+  ->UseManualTime();
 
 BENCHMARK_TEMPLATE(BM_static_map_search_all, int32_t, int32_t, dist_type::UNIFORM)
-->Unit(benchmark::kMillisecond)
-->Apply(generate_size_and_occupancy);
+  ->Unit(benchmark::kMillisecond)
+  ->Apply(generate_size_and_occupancy);
 
 BENCHMARK_TEMPLATE(BM_static_map_insert, int32_t, int32_t, dist_type::GAUSSIAN)
-->Unit(benchmark::kMillisecond)
-->Apply(generate_size_and_occupancy)
-->UseManualTime();
+  ->Unit(benchmark::kMillisecond)
+  ->Apply(generate_size_and_occupancy)
+  ->UseManualTime();
 
 BENCHMARK_TEMPLATE(BM_static_map_search_all, int32_t, int32_t, dist_type::GAUSSIAN)
-->Unit(benchmark::kMillisecond)
-->Apply(generate_size_and_occupancy);
+  ->Unit(benchmark::kMillisecond)
+  ->Apply(generate_size_and_occupancy);
 
 BENCHMARK_TEMPLATE(BM_static_map_insert, int64_t, int64_t, dist_type::UNIQUE)
-->Unit(benchmark::kMillisecond)
-->Apply(generate_size_and_occupancy)
-->UseManualTime();
+  ->Unit(benchmark::kMillisecond)
+  ->Apply(generate_size_and_occupancy)
+  ->UseManualTime();
 
 BENCHMARK_TEMPLATE(BM_static_map_search_all, int64_t, int64_t, dist_type::UNIQUE)
-->Unit(benchmark::kMillisecond)
-->Apply(generate_size_and_occupancy);
+  ->Unit(benchmark::kMillisecond)
+  ->Apply(generate_size_and_occupancy);
 
 BENCHMARK_TEMPLATE(BM_static_map_insert, int64_t, int64_t, dist_type::UNIFORM)
-->Unit(benchmark::kMillisecond)
-->Apply(generate_size_and_occupancy)
-->UseManualTime();
+  ->Unit(benchmark::kMillisecond)
+  ->Apply(generate_size_and_occupancy)
+  ->UseManualTime();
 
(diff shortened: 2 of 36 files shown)


jrhemstad:
rerun tests

benchmarks/hash_table/dynamic_map_bench.cu
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
```
sleeepyjack, line 2: I see these kinds of line changes very frequently. How about we inject the year automatically using a `sed` script in the CI?

benchmarks/hash_table/dynamic_map_bench.cu
```diff
@@ -14,11 +14,14 @@
...
 
-#include <benchmark/benchmark.h>
+#include <synchronization.hpp>
+
 #include <cuco/dynamic_map.cuh>
+
+#include <benchmark/benchmark.h>
+
 #include <iostream>
 #include <random>
```
sleeepyjack, line 24: This seems to be a permutation of the same includes. I think `clang_format` is already sorting these in lexicographic order for us if they are defined in a contiguous block, which is nice and consistent. However, lexicographic order might not always be correct or desired. Maybe we can customize the `clang_format` file to always produce an ordering that satisfies our needs and then just let the CI take care of it.

benchmarks/hash_table/static_map_bench.cu
```diff
@@ -202,11 +208,55 @@ BENCHMARK_TEMPLATE(BM_static_map_insert, int32_t, int32_t, dist_type::UNIQUE)
   ->Apply(generate_size_and_occupancy)
   ->UseManualTime();
 
-BENCHMARK_TEMPLATE(BM_static_map_erase_all, int32_t, int32_t, dist_type::UNIQUE)
+BENCHMARK_TEMPLATE(BM_static_map_search_all, int32_t, int32_t, dist_type::UNIQUE)
```
sleeepyjack: Not specific to this PR, but I think we are using `BM_` only for `gbench` benchmarks but not for the `nvbench` ones. We should aim for a consistent naming scheme in the future.

benchmarks/hash_table/static_multimap/count_bench.cu
```diff
@@ -14,13 +14,13 @@
  * limitations under the License.
  */
 
-#include <random>
```
sleeepyjack, line 17: This is already included by the `key_generator.hpp`. Good!

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+#include <cub/cub.cuh>
```
sleeepyjack: We are just using the `cub::BlockReduce` in this header, so we could just include `<cub/block/block_reduce.cuh>`.

include/cuco/detail/pair.cuh
```diff
@@ -19,6 +19,7 @@
 #include <thrust/device_reference.h>
 #include <thrust/pair.h>
 #include <thrust/tuple.h>
+
```
sleeepyjack, line 22: I thought we would also need `<algorithm>` since we are using `std::min` in line L54 but CE didn't complain we pasting the file into it.

include/cuco/detail/static_map_kernels.cuh
```diff
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+#include <cub/cub.cuh>
```
sleeepyjack: We are just using the cub::BlockReduce in this header, so we could just include <cub/block/block_reduce.cuh>.

include/cuco/detail/static_multimap/kernels.cuh
```diff
@@ -14,10 +14,15 @@
  * limitations under the License.
  */
 
-#include <cooperative_groups/memcpy_async.h>
+#include <cuco/detail/pair.cuh>
+
 #include <thrust/type_traits/is_contiguous_iterator.h>
 
-#include <cuco/detail/pair.cuh>
+#include <cub/cub.cuh>
```
sleeepyjack: We are just using the cub::BlockReduce in this header, so we could just include <cub/block/block_reduce.cuh>.

include/cuco/detail/static_multimap/kernels.cuh
```diff
@@ -14,10 +14,15 @@
...
+
+#include <cuda/std/atomic>
```
sleeepyjack, line 23: We access the atomic types through a `typedef atomicT`, so we don't need to include `<cuda/std/atomic>` here... At least I think that's the case.

include/cuco/detail/static_multimap/static_multimap.inl
```diff
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
+#include <cuco/detail/utils.hpp>
+
 #include <thrust/count.h>
 #include <thrust/iterator/transform_iterator.h>
 #include <thrust/tuple.h>
```
sleeepyjack, line 22: In L472 we instantiate a `cuda::barrier` which needs a declaration. We also use `std::distance` and `std::make_pair` which also may need their declarations. Correct me if I'm wrong on this.

include/cuco/dynamic_map.cuh
```diff
@@ -16,13 +16,19 @@
...
+
 #include <thrust/device_vector.h>
+#include <thrust/functional.h>
+
+#include <cuda/std/atomic>
+
+#include <cstddef>
+#include <memory>
+#include <type_traits>
+#include <vector>
```
sleeepyjack, line 31: Took me a while to verify but looks good.

include/cuco/detail/static_multimap/static_multimap.inl
sleeepyjack: Ah, forget about it. I just realized we are using the `CUCO_HAS_CUDA_BARRIER`macro here.

tests/dynamic_map/unique_sequence_test.cu
```diff
@@ -14,12 +14,13 @@
...
-#include <catch2/catch.hpp>
-#include <thrust/device_vector.h>
+#include <utils.hpp>
 
 #include <cuco/dynamic_map.cuh>
 
-#include <utils.hpp>
+#include <thrust/device_vector.h>
+
+#include <catch2/catch.hpp>
```
sleeepyjack, line 23: btw [Catch2 v3](https://github.com/catchorg/Catch2/blob/devel/docs/release-notes.md#301) is out and now packaged as a statically compiled library which we can link against 🎉

tests/static_map/custom_type_test.cu
```diff
@@ -14,12 +14,15 @@
...
+#include <utils.hpp>
 
 #include <cuco/static_map.cuh>
 
-#include <utils.hpp>
+#include <thrust/device_vector.h>
+
+#include <catch2/catch.hpp>
+
+#include <tuple>
```
sleeepyjack, line 25: I'm not seeing it. Why do we need `<tuple` here?

sleeepyjack (commented):
Thanks @PointKernel for this PR! I made some comments targeting the automation of some reoccurring tasks regarding the code structure. This would also cause fewer diffs in future PRs. Happy to hear your thoughts on this.

tests/static_map/custom_type_test.cu
PointKernel: Needed by [std::tie](https://github.com/NVIDIA/cuCollections/blob/5bca985018e652d44f5d9d9a5da67fc5ca39e502/tests/static_map/custom_type_test.cu#L88)

tests/dynamic_map/unique_sequence_test.cu
PointKernel: Filed an issue to track this: #153

include/cuco/detail/static_multimap/kernels.cuh
PointKernel: It's still needed due to atomic operations like this: https://github.com/NVIDIA/cuCollections/blob/5bca985018e652d44f5d9d9a5da67fc5ca39e502/include/cuco/detail/static_multimap/kernels.cuh#L251

Removing it will cause build failure:
```
/home/yunsongw/Work/cuCollections/include/cuco/detail/static_multimap/kernels.cuh(256): error: name followed by "::" must be a class or namespace name
          detected during instantiation of "std::size_t cuco::static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::count_outer(InputIt, InputIt, cudaStream_t, KeyEqual) const [with Key=int, Value=int, Scope=cuda::std::__4::__detail::thread_scope_device, Allocator=cuco::cuda_allocator<char>, ProbeSequence=cuco::double_hashing<8U, cuco::detail::MurmurHash3_32<int>, cuco::detail::MurmurHash3_32<int>>, InputIt=thrust::detail::normal_iterator<thrust::device_ptr<int>>, KeyEqual=thrust::equal_to<int>]" 

```

include/cuco/detail/pair.cuh
PointKernel: Good catch.

benchmarks/hash_table/static_map_bench.cu
PointKernel: Right, `gbench` will soon be replaced by `nvbench` so this shouldn't be a big issue.

benchmarks/hash_table/dynamic_map_bench.cu
PointKernel: We can copy-paste the cudf [script](https://github.com/rapidsai/cudf/blob/branch-22.06/ci/checks/copyright.py) and make some changes. It's on my to-do list as well.

benchmarks/hash_table/dynamic_map_bench.cu
PointKernel: The idea here is to separate different header groups and order them from "near" to "far". e.g. `synchronization.hpp` is a bench-local header thus it's placed before the library header `cuco/dynamic_map.cuh`. gbench header is even further but considered closer than std headers.

This grouping method seems a bit awkward with only one file in each group but will show its advantage with more headers involved.

include/cuco/detail/static_multimap/static_multimap.inl
PointKernel: Added `iterator` header for `std::distance`

benchmarks/hash_table/dynamic_map_bench.cu
sleeepyjack: Ah ok, this makes sense. I wonder however if we should add a CI script for that. Basically extract all includes, check in the include tree where this file originates, group and then reorder.

include/cuco/detail/static_multimap/kernels.cuh
sleeepyjack: Yes but in L242 we call view.count() which is also not declared (only after the template is instantiated).

tests/static_map/custom_type_test.cu
sleeepyjack: I am blind :D

jrhemstad (approved):
Thanks @PointKernel!

PR #144, opened 2022-03-24
[FIX] Get Thrust via `rapids_cpm_thrust`

trxcllnt:
* Get Thrust via `rapids_cpm_thrust`
* Don't immediately create the `cuco::Thrust` target, as this will be done by `cuco-config.cmake`

Fix an issue where including two projects that both use CPM to load `cuco` fails to configure with this error:
```cmake
CMake Error at thrust-config.cmake:176 (add_library):
  add_library cannot create imported target "cuco::Thrust" because another
  target with the same name already exists.
```


PointKernel:
@robertmaynard Could you review this PR?

robertmaynard (approved):

Issue #65, opened 2020-11-16
[FEA] CUDA stream support


seunghwak:
**Is your feature request related to a problem? Please describe.**
static_map/dynamic_map currently does not take cudaStream_t stream parameter. This often requires additional synchronization and limits speedup we can get when we wish to concurrently run multiple cuCollection operations using multiple CUDA streams.

**Describe the solution you'd like**
Add CUDA stream support

**Additional context**
cuGraph needs this to run multiple graph kernels concurrently using multiple CUDA streams (for batch processing).

chirayuG-nvidia:
Stream support for static_map added in commit https://github.com/NVIDIA/cuCollections/commit/76a6ba1d08761f7e14f27c6794ca2d469301b54f

PointKernel:
Completed.

All new cuco hash tables take a stream parameter.

PR #149, opened 2022-04-06
Erase Functionality for dynamic_map

Nicolas-Iskos:
Erase support added for `cuco::dynamic_map`, based on `cuco::static_map` with erase support. See #142. More details to follow soon. 

Some benchmark data:

Insert throughput before_after erase support (unique 4B_4B)
Search throughput before_after erase-support (unique 4B_4B)
Erase-all and Insert Throughput (unique 4B_4B)
Erase-none and Search-none Throughput (unique 4B_4B)


GPUtester:

PointKernel:
ok to test

PointKernel:
ok to test

PointKernel:
add to allowlist

PointKernel:
ok to test

PointKernel:
add to allowlist

PointKernel:
ok to test

PointKernel:
add to allowlist

tests/dynamic_map/erase_test.cu
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright (c) 2020-2021, NVIDIA CORPORATION.
```
PointKernel:
```suggestion
 * Copyright (c) 2022, NVIDIA CORPORATION.
```

tests/dynamic_map/erase_test.cu
```diff
@@ -0,0 +1,138 @@
...
+ */
+
+#include <catch2/catch.hpp>
+#include <thrust/device_vector.h>
+#include <thrust/execution_policy.h>
+#include <thrust/sequence.h>
+
+#include <cuco/dynamic_map.cuh>
+
+#include <utils.hpp>
```
PointKernel:
```suggestion
#include <utils.hpp>

#include <cuco/dynamic_map.cuh>

#include <thrust/device_vector.h>
#include <thrust/execution_policy.h>
#include <thrust/sequence.h>

#include <catch2/catch.hpp>
```
Based on [cudf include guidelines](https://github.com/rapidsai/cudf/blob/46c5e90bf1565ac4df87949a806597160ffbc22f/cpp/docs/DEVELOPER_GUIDE.md#includes), we include header groups from the "nearest" to "furthest".

PointKernel:
@Nicolas-Iskos Can you please check the doxygen warnings [here](https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/nvidia%2Fgpuci%2FcuCollections%2Fprb%2FcuCollections-style/detail/cuCollections-style/241/pipeline) (in the `check-style` CI test) and fix them?

benchmarks/hash_table/static_map_bench.cu
```diff
@@ -200,60 +244,105 @@ static void BM_static_map_erase_all(::benchmark::State& state)
...
+    map.erase(d_keys.begin(), d_keys.end());
+
+    // state.PauseTiming();
+  }
+
+  state.SetBytesProcessed((sizeof(Key) + sizeof(Value)) * int64_t(state.iterations()) *
+                          int64_t(state.range(0)));
+}
+
+/*
```
PointKernel:
```suggestion
```
Unless it's intended, please do not remove existing benchmarks.

benchmarks/hash_table/static_map_bench.cu
```diff
@@ -200,60 +244,105 @@ static void BM_static_map_erase_all(::benchmark::State& state)
...
   ->Unit(benchmark::kMillisecond)
-  ->Apply(generate_size_and_occupancy)
-  ->UseManualTime();
+  ->Apply(generate_size_and_occupancy);
 
-BENCHMARK_TEMPLATE(BM_static_map_search_all, int32_t, int32_t, dist_type::GAUSSIAN)
+BENCHMARK_TEMPLATE(BM_static_map_search_none, int64_t, int64_t, dist_type::UNIQUE)
   ->Unit(benchmark::kMillisecond)
   ->Apply(generate_size_and_occupancy);
+*/
```
PointKernel:
```suggestion
```

benchmarks/hash_table/static_map_bench.cu
```diff
@@ -200,60 +244,105 @@ static void BM_static_map_erase_all(::benchmark::State& state)
...
+BENCHMARK_TEMPLATE(BM_static_map_search_all, int32_t, int32_t, dist_type::GAUSSIAN)
   ->Unit(benchmark::kMillisecond)
   ->Apply(generate_size_and_occupancy);
 
-BENCHMARK_TEMPLATE(BM_static_map_erase_all, int32_t, int32_t, dist_type::UNIQUE)
+BENCHMARK_TEMPLATE(BM_static_map_insert, int64_t, int64_t, dist_type::UNIQUE)
   ->Unit(benchmark::kMillisecond)
-  ->Apply(generate_size_and_occupancy);
+  ->Apply(generate_size_and_occupancy)
+  ->UseManualTime();
```
PointKernel: I noticed there are several existing benchmarks being removed with this PR. Is the removal intentional?

benchmarks/hash_table/dynamic_map_bench.cu
```diff
@@ -137,62 +137,140 @@ static void BM_dynamic_search_all(::benchmark::State& state)
...
   ->Unit(benchmark::kMillisecond)
   ->Apply(gen_final_size)
   ->UseManualTime();
 
-BENCHMARK_TEMPLATE(BM_dynamic_search_all, int64_t, int64_t, dist_type::GAUSSIAN)
+BENCHMARK_TEMPLATE(BM_dynamic_erase_all, int32_t, int32_t, dist_type::UNIQUE)
   ->Unit(benchmark::kMillisecond)
   ->Apply(gen_final_size)
-  ->UseManualTime();
+  ->UseManualTime();
```
PointKernel: We might want to keep the existing benchmarks as they are right now.

If you want, similar to multimap benchmarks, we can split the current file into several smaller ones and relocate them into a new `hash_table/dynamic_map` folder.

include/cuco/detail/static_map_kernels.cuh
```diff
@@ -163,6 +163,26 @@ __global__ void insert(
...
+ * If the key `*(first + i)` exists in the map, its slot is erased and made available for future
+   insertions.
+ * Else, no effect.
+ * @tparam block_size The size of the thread block
+ * @tparam InputIt Device accessible input iterator whose `value_type` is
+ * convertible to the map's `key_type`
+ * @tparam atomicT Type of atomic storage
+ * @tparam viewT Type of device view allowing access of hash map storage
+ * @tparam Hash Unary callable type
+ * @tparam KeyEqual Binary callable type
```
PointKernel, line 183:
```suggestion
 * @tparam KeyEqual Binary callable type
 
```
blank line between `param` and `tparam`

include/cuco/detail/static_map_kernels.cuh
```diff
@@ -163,6 +163,26 @@ __global__ void insert(
   if (threadIdx.x == 0) { *num_successes += block_num_successes; }
 }
 
+/**
+ * @brief Erases the key/value pairs corresponding to all keys in the range `[first, last)`.
+ *
+ * If the key `*(first + i)` exists in the map, its slot is erased and made available for future
+   insertions.
+ * Else, no effect.
```
PointKernel, line 175:
```suggestion
 * Else, no effect.
 
```

include/cuco/detail/static_map_kernels.cuh
```diff
@@ -192,6 +212,27 @@ __global__ void erase(
   }
 }
 
+/**
+ * @brief Erases the key/value pairs corresponding to all keys in the range `[first, last)`.
+ *
+ * If the key `*(first + i)` exists in the map, its slot is erased and made available for future
+   insertions.
+ * Else, no effect.
```
PointKernel, line 226:
```suggestion
 * Else, no effect.
 
```

include/cuco/detail/static_map_kernels.cuh
```diff
@@ -192,6 +212,27 @@ __global__ void erase(
...
+ * @tparam block_size The size of the thread block
+ * @tparam tile_size The number of threads in the Cooperative Groups used to perform erase
+ * @tparam InputIt Device accessible input iterator whose `value_type` is
+ * convertible to the map's `key_type`
+ * @tparam atomicT Type of atomic storage
+ * @tparam viewT Type of device view allowing access of hash map storage
+ * @tparam Hash Unary callable type
+ * @tparam KeyEqual Binary callable type
```
PointKernel, line 235:
```suggestion
 * @tparam KeyEqual Binary callable type
 
```

include/cuco/dynamic_map.cuh
```diff
@@ -16,11 +16,13 @@
...
 #pragma once
 
+#include <cooperative_groups.h>
+#include <cub/cub.cuh>
 #include <cuco/detail/dynamic_map_kernels.cuh>
 #include <cuco/detail/error.hpp>
 #include <cuco/sentinel.cuh>
 #include <cuco/static_map.cuh>
-
+#include <cuda/std/atomic>
```
PointKernel:
```suggestion

```
Duplicated. Already included at line 29

include/cuco/dynamic_map.cuh
```diff
@@ -99,16 +101,24 @@ class dynamic_map {
...
-  using value_type      = cuco::pair_type<Key, Value>;       ///< Type of key/value pairs
-  using key_type        = Key;                               ///< Key type
-  using mapped_type     = Value;                             ///< Type of mapped values
-  using atomic_ctr_type = cuda::atomic<std::size_t, Scope>;  ///< Type of atomic counters
-  using view_type = typename static_map<Key, Value, Scope>::device_view;  ///< Device view type
+  using value_type        = cuco::pair_type<Key, Value>;
+  using key_type          = Key;
+  using mapped_type       = Value;
+  using atomic_ctr_type   = cuda::atomic<std::size_t, Scope>;
+  using view_type         = typename static_map<Key, Value, Scope>::device_view;
```
PointKernel: Removing docs here will cause doxygen warnings in CI style-check.

include/cuco/dynamic_map.cuh
```diff
@@ -16,11 +16,13 @@
 
 #pragma once
 
+#include <cooperative_groups.h>
+#include <cub/cub.cuh>
```
PointKernel: One simple rule we are following is to include what's used. It seems cub and CG are not used in the current file thus we can defer the inclusion.
```suggestion
```

include/cuco/dynamic_map.cuh
```diff
@@ -172,6 +212,40 @@ class dynamic_map {
...
+   *  - `get_size()` is reduced by the total number of erased keys
+   *
+   * This function synchronizes `stream`.
+   *
+   * Keep in mind that `erase` does not cause the map to shrink its memory allocation.
+   *
+   * @tparam InputIt Device accessible input iterator whose `value_type` is
+   * convertible to the map's `value_type`
+   * @tparam Hash Unary callable type
+   * @tparam KeyEqual Binary callable type
```
PointKernel, line 239:
```suggestion
   * @tparam KeyEqual Binary callable type
   
```

include/cuco/dynamic_map.cuh
```diff
@@ -99,16 +101,24 @@ class dynamic_map {
...
 
   dynamic_map(dynamic_map const&) = delete;
   dynamic_map(dynamic_map&&)      = delete;
+
+  template <typename T1, typename T2>
+  dynamic_map(std::size_t, T1, T2, Allocator const& = Allocator{}) = delete;
+
+  template <typename T1, typename T2, typename T3>
+  dynamic_map(std::size_t, T1, T2, T3, Allocator const& = Allocator{}) = delete;
+
```
PointKernel:
```suggestion
```
I think this is no longer needed since sentinel ctors are explicit now.

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -186,6 +186,189 @@ __global__ void insert(InputIt first,
...
+    }
+  }
+}
+
+/**
+ * @brief Erases the key/value pairs corresponding to all keys in the range `[first, last)`.
+ *
+ * If the key `*(first + i)` exists in the map, its slot is erased and made available for future
+   insertions.
+ * Else, no effect.
```
PointKernel, line 270:
```suggestion
 * Else, no effect.
 
```

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -186,6 +186,189 @@ __global__ void insert(InputIt first,
...
+ * perform `erase` operations on each underlying `static_map`
+ * @param num_successes The number of successfully erased key/value pairs
+ * @param submap_num_successes The number of successfully erased key/value pairs
+ * in each submap
+ * @param num_submaps The number of submaps in the map
+ * @param hash The unary function to apply to hash each key
+ * @param key_equal The binary function to compare two keys for equality
+ */
+template <uint32_t block_size,
+          typename pair_type,
```
PointKernel: `pair_type` seems unused

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -186,6 +186,189 @@ __global__ void insert(InputIt first,
...
+ * Else, no effect.
+ * @tparam block_size The size of the thread block
+ * @tparam tile_size The number of threads in the Cooperative Groups used to perform erase
+ * @tparam pair_type Type of the pairs contained in the map
+ * @tparam InputIt Device accessible input iterator whose `value_type` is
+ * convertible to the map's `key_type`
+ * @tparam mutableViewT Type of device view allowing modification of hash map storage
+ * @tparam atomicT Type of atomic storage
+ * @tparam Hash Unary callable type
+ * @tparam KeyEqual Binary callable type
```
PointKernel, line 279:
```suggestion
 * @tparam KeyEqual Binary callable type
 
```

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -186,6 +186,189 @@ __global__ void insert(InputIt first,
...
+ * @param num_successes The number of successfully erased key/value pairs
+ * @param submap_num_successes The number of successfully erased key/value pairs
+ * in each submap
+ * @param num_submaps The number of submaps in the map
+ * @param hash The unary function to apply to hash each key
+ * @param key_equal The binary function to compare two keys for equality
+ */
+template <uint32_t block_size,
+          uint32_t tile_size,
+          typename pair_type,
```
PointKernel: marker for unused `pair_type`.

include/cuco/detail/dynamic_map.inl
```diff
@@ -128,6 +171,64 @@ void dynamic_map<Key, Value, Scope, Allocator>::insert(InputIt first,
...
+  auto const tile_size  = 4;
+  auto const grid_size  = (tile_size * num_keys + stride * block_size - 1) / (stride * block_size);
+
+  // TODO: memset an atomic variable is unsafe
+  static_assert(sizeof(std::size_t) == sizeof(atomic_ctr_type));
+  CUCO_CUDA_TRY(cudaMemset(num_successes_, 0, sizeof(atomic_ctr_type)));
+
+  // zero out submap success counters
+  if (submaps_.size() > 1) {
+    static_assert(sizeof(std::size_t) == sizeof(atomic_ctr_type));
```
PointKernel:
```suggestion
```
Seems redundant

include/cuco/detail/dynamic_map.inl
```diff
@@ -128,6 +171,64 @@ void dynamic_map<Key, Value, Scope, Allocator>::insert(InputIt first,
...
+  }
+
+  auto const temp_storage_size = submaps_.size() * sizeof(unsigned long long);
+
+  detail::erase<block_size, tile_size, cuco::pair_type<key_type, mapped_type>>
+    <<<grid_size, block_size, temp_storage_size>>>(first,
+                                                   first + num_keys,
+                                                   submap_mutable_views_.data().get(),
+                                                   num_successes_,
+                                                   d_submap_num_successes_.data().get(),
```
PointKernel: Passing `d_submap_num_successes` looks fishy. 

We set all elements in `submap_num_successes_` to 0 before the kernel and memcpy `submap_num_successes_` back to the host after the kernel while the parameter actually passed to the kernel is `d_submap_num_successes`.

include/cuco/detail/dynamic_map.inl
```diff
@@ -37,14 +39,43 @@ dynamic_map<Key, Value, Scope, Allocator>::dynamic_map(
...
+    initial_capacity,
+    sentinel::empty_key<Key>{empty_key_sentinel_},
+    sentinel::empty_value<Value>{empty_value_sentinel_},
+    sentinel::erased_key<Key>{erased_key_sentinel_},
+    alloc));
+  submap_views_.push_back(submaps_[0]->get_device_view());
+  submap_mutable_views_.push_back(submaps_[0]->get_device_mutable_view());
+  submap_num_successes_.push_back(submaps_[0]->get_num_successes());
+  d_submap_num_successes_ = submap_num_successes_;
+  num_successes_ = std::allocator_traits<counter_allocator_type>::allocate(counter_allocator_, 1);
```
PointKernel: Init with initializer list when possible

include/cuco/detail/dynamic_map.inl
```diff
@@ -37,14 +39,43 @@ dynamic_map<Key, Value, Scope, Allocator>::dynamic_map(
     alloc));
   submap_views_.push_back(submaps_[0]->get_device_view());
   submap_mutable_views_.push_back(submaps_[0]->get_device_mutable_view());
+  num_successes_ = std::allocator_traits<counter_allocator_type>::allocate(counter_allocator_, 1);
```
PointKernel: Init with initializer list when possible

include/cuco/detail/dynamic_map.inl
```diff
@@ -128,6 +171,64 @@ void dynamic_map<Key, Value, Scope, Allocator>::insert(InputIt first,
...
+  std::size_t num_keys = std::distance(first, last);
+
+  auto const block_size = 128;
+  auto const stride     = 1;
+  auto const tile_size  = 4;
+  auto const grid_size  = (tile_size * num_keys + stride * block_size - 1) / (stride * block_size);
+
+  // TODO: memset an atomic variable is unsafe
+  static_assert(sizeof(std::size_t) == sizeof(atomic_ctr_type));
+  CUCO_CUDA_TRY(cudaMemset(num_successes_, 0, sizeof(atomic_ctr_type)));
```
PointKernel: Can we get rid of `num_successes_` in dynamic map? It's just the sum of all counters in `submap_num_successes` and we don't have to do the atomic add twice.

include/cuco/static_map.cuh
```diff
@@ -1414,6 +1414,8 @@ class static_map {
                                sentinel::erased_key<Key>{erased_key_sentinel_});
   }
 
+  atomic_ctr_type* get_num_successes() const noexcept { return num_successes_; }
+
```
PointKernel: This shouldn't be public. `num_successes_` is for internal use only and it's reset to zero each time before using.

include/cuco/dynamic_map.cuh
```diff
@@ -172,6 +212,40 @@ class dynamic_map {
...
+   * @param key_equal The binary function to compare two keys for equality
+   * @param stream Stream used for executing the kernels
+   *
+   * @throw std::runtime_error if a unique erased key sentinel value was not
+   * provided at construction
+   */
+  template <typename InputIt,
+            typename Hash     = cuco::detail::MurmurHash3_32<key_type>,
+            typename KeyEqual = thrust::equal_to<key_type>>
+  void erase(InputIt first, InputIt last, Hash hash = Hash{}, KeyEqual key_equal = KeyEqual{});
```
PointKernel: Can you please add a stream parameter w.r.t. https://github.com/NVIDIA/cuCollections/issues/65?

PointKernel (changes requested):
The first round of review.

include/cuco/dynamic_map.cuh
Nicolas-Iskos: Yep, added to all `dynamic_map` functions

include/cuco/detail/dynamic_map.inl
Nicolas-Iskos: Good point. Removed `num_successes_`.

benchmarks/hash_table/static_map_bench.cu
Nicolas-Iskos: The old benchmarks have been added back for simplicity, although they were pretty random to begin with.

tests/dynamic_map/erase_test.cu
Nicolas-Iskos: Should be fixed

include/cuco/detail/dynamic_map.inl
```diff
@@ -18,12 +18,12 @@ namespace cuco {
...
-                                                       Key empty_key_sentinel,
-                                                       Value empty_value_sentinel,
+                                                       sentinel::empty_key<Key> empty_key_sentinel,
+                                                       sentinel::empty_value<Value> empty_value_sentinel,
                                                        Allocator const& alloc)
-  : empty_key_sentinel_(empty_key_sentinel),
-    empty_value_sentinel_(empty_value_sentinel),
-    erased_key_sentinel_(empty_key_sentinel),
+  : empty_key_sentinel_(empty_key_sentinel.value),
+    empty_value_sentinel_(empty_value_sentinel.value),
```
sleeepyjack: Let's enable copy/move construction of sentinels.

include/cuco/detail/dynamic_map.inl
```diff
@@ -94,15 +93,22 @@ void dynamic_map<Key, Value, Scope, Allocator>::reserve(std::size_t n)
     // if the submap does not exist yet, create it
     else {
       submap_capacity = capacity_;
-      submaps_.push_back(std::make_unique<static_map<Key, Value, Scope, Allocator>>(
-        submap_capacity,
-        sentinel::empty_key<Key>{empty_key_sentinel_},
-        sentinel::empty_value<Value>{empty_value_sentinel_},
-        sentinel::erased_key<Key>{erased_key_sentinel_},
-        alloc_));
+      if(erased_key_sentinel_ != empty_key_sentinel_) {
```
sleeepyjack: If the caller accidentally initializes `erased_key_sentinel_` with the same value as `empty_key_sentinel_`, deletions are disabled as this is the default behavior here. It would be nice to detect this case and throw a runtime error.
However, I think with the current design this is not possible.

include/cuco/dynamic_map.cuh
```diff
@@ -131,26 +144,60 @@ class dynamic_map {
...
    * @param empty_value_sentinel The reserved mapped value for empty slots
    * @param alloc Allocator used to allocate submap device storage
+   * @param stream Stream used for executing the kernels
    */
   dynamic_map(std::size_t initial_capacity,
               sentinel::empty_key<Key> empty_key_sentinel,
               sentinel::empty_value<Value> empty_value_sentinel,
-              Allocator const& alloc = Allocator{});
+              Allocator const& alloc = Allocator{},
+              cudaStream_t stream    = 0);
```
sleeepyjack:
```suggestion
              cudaStream_t stream    = nullptr);
```

include/cuco/dynamic_map.cuh
```diff
@@ -131,26 +144,60 @@ class dynamic_map {
...
+   *
+   * @throw std::runtime error if the empty key sentinel and erased key sentinel
+   * are the same value
+   */
+  dynamic_map(std::size_t initial_capacity,
+              sentinel::empty_key<Key> empty_key_sentinel,
+              sentinel::empty_value<Value> empty_value_sentinel,
+              sentinel::erased_key<Key> erased_key_sentinel,
+              Allocator const& alloc = Allocator{},
+              cudaStream_t stream    = 0);
```
sleeepyjack:
```suggestion
              cudaStream_t stream    = nullptr);
```

sleeepyjack: clang-tidy might complain otherwise

include/cuco/dynamic_map.cuh
```diff
@@ -131,26 +144,60 @@ class dynamic_map {
...
   /**
    * @brief Grows the capacity of the map so there is enough space for `n` key/value pairs.
    *
    * If there is already enough space for `n` key/value pairs, the capacity remains the same.
    *
    * @param n The number of key value pairs for which there must be space
+   * @param stream Stream used for executing the kernels
    */
-  void reserve(std::size_t n);
+  void reserve(std::size_t n, cudaStream_t stream = 0);
```
sleeepyjack:
```suggestion
  void reserve(std::size_t n, cudaStream_t stream = nullptr);
```

include/cuco/dynamic_map.cuh
```diff
@@ -166,11 +213,54 @@ class dynamic_map {
...
+   * @param stream Stream used for executing the kernels
+   */
+  template <typename InputIt,
+            typename Hash     = cuco::detail::MurmurHash3_32<key_type>,
+            typename KeyEqual = thrust::equal_to<key_type>>
+  void insert(InputIt first,
+              InputIt last,
+              Hash hash           = Hash{},
+              KeyEqual key_equal  = KeyEqual{},
+              cudaStream_t stream = 0);
```
sleeepyjack:
```suggestion
              cudaStream_t stream = nullptr);
```

include/cuco/dynamic_map.cuh
```diff
@@ -166,11 +213,54 @@ class dynamic_map {
...
    */
   template <typename InputIt,
             typename Hash     = cuco::detail::MurmurHash3_32<key_type>,
             typename KeyEqual = thrust::equal_to<key_type>>
-  void insert(InputIt first, InputIt last, Hash hash = Hash{}, KeyEqual key_equal = KeyEqual{});
+  void erase(InputIt first,
+             InputIt last,
+             Hash hash           = Hash{},
+             KeyEqual key_equal  = KeyEqual{},
+             cudaStream_t stream = 0);
```
sleeepyjack:
```suggestion
             cudaStream_t stream = nullptr);
```

include/cuco/dynamic_map.cuh
```diff
@@ -197,8 +288,9 @@ class dynamic_map {
   void find(InputIt first,
             InputIt last,
             OutputIt output_begin,
-            Hash hash          = Hash{},
-            KeyEqual key_equal = KeyEqual{});
+            Hash hash           = Hash{},
+            KeyEqual key_equal  = KeyEqual{},
+            cudaStream_t stream = 0);
```
sleeepyjack:
```suggestion
            cudaStream_t stream = nullptr);
```

include/cuco/dynamic_map.cuh
```diff
@@ -224,8 +317,9 @@ class dynamic_map {
   void contains(InputIt first,
                 InputIt last,
                 OutputIt output_begin,
-                Hash hash          = Hash{},
-                KeyEqual key_equal = KeyEqual{});
+                Hash hash           = Hash{},
+                KeyEqual key_equal  = KeyEqual{},
+                cudaStream_t stream = 0);
```
sleeepyjack:
```suggestion
                cudaStream_t stream = nullptr);
```

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -183,7 +184,153 @@ __global__ void insert(InputIt first,
...
+  __syncthreads();
+
+  while (it < last) {
+    auto erased = false;
+    int i;
+    for (i = 0; i < num_submaps; ++i) {
+      erased = submap_mutable_views[i].erase(tile, *it, hash, key_equal);
+      if (erased) { break; }
+    }
+    if (erased && tile.thread_rank() == 0) { atomicAdd(&submap_block_num_successes[i], 1); }
```
sleeepyjack, line 326: Minor suggestion: Can we switch to `cuda::atomic_ref` here?

include/cuco/detail/dynamic_map.inl
```diff
@@ -80,25 +121,24 @@ void dynamic_map<Key, Value, Scope, Allocator>::reserve(std::size_t n)
...
     // only if we meet the minimum insert size.
+
     if (capacity_remaining >= min_insert_size_) {
-      *num_successes_ = 0;
-      int device_id;
-      CUCO_CUDA_TRY(cudaGetDevice(&device_id));
-      CUCO_CUDA_TRY(cudaMemPrefetchAsync(num_successes_, sizeof(atomic_ctr_type), device_id));
+      // TODO: memset an atomic variable is unsafe
+      static_assert(sizeof(std::size_t) == sizeof(atomic_ctr_type));
+      CUCO_CUDA_TRY(cudaMemset(submap_num_successes_[submap_idx], 0, sizeof(atomic_ctr_type)));
```
sleeepyjack:
```suggestion
      CUCO_CUDA_TRY(cudaMemsetAsync(submap_num_successes_[submap_idx], 0, sizeof(atomic_ctr_type), stream));
```
Although I don't know if this plays well with the `CUCO_CUDA_TRY` macro.

include/cuco/detail/dynamic_map.inl
```diff
@@ -128,34 +171,85 @@ void dynamic_map<Key, Value, Scope, Allocator>::insert(InputIt first,
...
+  auto const block_size = 128;
+  auto const stride     = 1;
+  auto const tile_size  = 4;
+  auto const grid_size  = (tile_size * num_keys + stride * block_size - 1) / (stride * block_size);
+
+  // TODO: memset an atomic variable is unsafe
+  static_assert(sizeof(std::size_t) == sizeof(atomic_ctr_type));
+
+  // zero out submap success counters
+  static_assert(sizeof(std::size_t) == sizeof(atomic_ctr_type));
```
sleeepyjack: Duplicate check.
Can we also emit a suitable error message?

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -183,7 +184,153 @@ __global__ void insert(InputIt first,
...
+
+  for (int i = threadIdx.x; i < num_submaps; i += block_size)
+    submap_block_num_successes[i] = 0;
+  __syncthreads();
+
+  while (it < last) {
+    int i;
+    for (i = 0; i < num_submaps; ++i) {
+      if (submap_mutable_views[i].erase(*it, hash, key_equal)) {
+        atomicAdd(&submap_block_num_successes[i], 1);
```
sleeepyjack, line 249: We could also use `cuda::atomic_ref` here.

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -183,7 +184,153 @@ __global__ void insert(InputIt first,
...
+          typename KeyEqual>
+__global__ void erase(InputIt first,
+                      InputIt last,
+                      mutableViewT* submap_mutable_views,
+                      atomicT** submap_num_successes,
+                      const uint32_t num_submaps,
+                      Hash hash,
+                      KeyEqual key_equal)
+{
+  typedef cub::BlockReduce<std::size_t, block_size> BlockReduce;
```
sleeepyjack: What's our standpoint on the `using` vs. `typedef` front?

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -183,7 +184,153 @@ __global__ void insert(InputIt first,
...
+  extern __shared__ unsigned long long submap_block_num_successes[];
+
+  auto tile = cg::tiled_partition<tile_size>(cg::this_thread_block());
+  auto tid  = block_size * blockIdx.x + threadIdx.x;
+  auto it   = first + tid / tile_size;
+
+  for (int i = threadIdx.x; i < num_submaps; i += block_size)
+    submap_block_num_successes[i] = 0;
+  __syncthreads();
```
sleeepyjack:
```suggestion
  auto block = cg::this_thread_block();
  auto tile  = cg::tiled_partition<tile_size>(block);
  auto tid   = block_size * block.group_index().x  block.thread_rank();
  auto it    = first + tid / tile_size;

  for (int i = block.thread_rank(); i < num_submaps; i += block_size)
    submap_block_num_successes[i] = 0;
  block.sync();
```
Just a suggestion to use the `cooperative_groups` API whenever possible for better readability.

include/cuco/detail/static_map_kernels.cuh
```diff
@@ -192,6 +212,27 @@ __global__ void erase(
...
+ * @param first Beginning of the sequence of keys
+ * @param last End of the sequence of keys
+ * @param num_successes The number of successfully erased key/value pairs
+ * @param view Device view used to access the hash map's slot storage
+ * @param hash The unary function to apply to hash each key
+ * @param key_equal The binary function to compare two keys for equality
+ */
 template <std::size_t block_size,
           uint32_t tile_size,
```
sleeepyjack, line 245: Why is `block_size` a `size_t` but `tile_size` an `uint32_t`?
I think we can get away with declaring both as `int` but since this occurs throughout the whole library it's out of scope for this PR.
We'll keep it in mind for the refactor.

include/cuco/detail/dynamic_map_kernels.cuh
Nicolas-Iskos: I don't really mind either way. Do you have a preference?

include/cuco/detail/dynamic_map_kernels.cuh
Nicolas-Iskos: That's a good point, but there are a lot of places in the library for which this would have to be changed for consistency, right?

include/cuco/detail/dynamic_map.inl
Nicolas-Iskos: Yep, added that in.

include/cuco/detail/dynamic_map.inl
Nicolas-Iskos: I added this check into the `dynamic_map` constructor. Does that accomplish what you were thinking of?

include/cuco/detail/dynamic_map.inl
Nicolas-Iskos: Is this not enabled by default? I didn't delete those constructors in `sentinel.cuh`.

include/cuco/detail/dynamic_map_kernels.cuh
Nicolas-Iskos: Good idea, added.

include/cuco/detail/static_map_kernels.cuh
Nicolas-Iskos: Yeah, probably there are a few areas where this would need to be addressed.

include/cuco/detail/dynamic_map_kernels.cuh
PointKernel: `using` is preferred throughout the project.

include/cuco/detail/static_map_kernels.cuh
PointKernel: > I think we can get away with declaring both as int but since this occurs throughout the whole library it's out of scope for this PR.

:+1: We should use `int32_t` consistently during refactoring.

include/cuco/detail/dynamic_map_kernels.cuh
Nicolas-Iskos: Sounds good. Switched these to `using`.

include/cuco/detail/dynamic_map.inl
sleeepyjack: Right, those ctors do exists by default, we just haven't used them here. The change looks good. Resolved.

sleeepyjack: Right, these ctors do exists so we can rewrite these lines as
```cpp
  : empty_key_sentinel_(empty_key_sentinel),
    empty_value_sentinel_(empty_value_sentinel),
```

include/cuco/detail/dynamic_map_kernels.cuh
sleeepyjack: Yeah, it's probably out of scope for this PR. We can leave it as is.

include/cuco/detail/dynamic_map.inl
Nicolas-Iskos: But `empty_key_sentinel_` has type `Key`, not `sentinel::empty_key<Key>`, so we need to retrieve the `value` from the sentinel before assigning it to `empty_key_sentinel_`, right?

include/cuco/detail/dynamic_map.inl
Nicolas-Iskos: `submap_num_successes_` and `d_submap_num_successes` are both arrays of `atomic_ctr_type*`. They contain the same device pointers. The only difference is  `submap_num_successes_` is on the host while `d_submap_num_successes` is on the device. Just different ways to access the same pointers.

benchmarks/hash_table/dynamic_map_bench.cu
Nicolas-Iskos: I added the rest of the benchmarks back in. Although even before, there wasn't any particular reason why that subset of benchmarks was being executed.

include/cuco/detail/dynamic_map.inl
sleeepyjack: Oh, whups, my bad. I had the new design for the grand refactor still in mind which holds the actual sentinel type as a member + the sentinel provides an implicit conversion operator for convenience.
Nevermind, we can leave it as is then.

include/cuco/detail/dynamic_map.inl
Nicolas-Iskos: Hm, why wouldn't it play well with `CUCO_CUDA_TRY`? I think the only way the call can return an error is if the memory address is invalid, which is independent of whether or not the memory has actually been written yet, right?

sleeepyjack: You're right. The async memcpy should return a value immediately which is then consumed by the error macro.

ajschmidt8:
rerun tests

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -147,13 +149,14 @@ __global__ void insert(InputIt first,
                        InputIt last,
                        viewT* submap_views,
                        mutableViewT* submap_mutable_views,
-                       atomicT* num_successes,
+                       // atomicT* num_successes,
```
PointKernel:
```suggestion
```

PointKernel (approved):

PR #151, opened 2022-05-02
Fix a bug in `static_multimap`: add early exit for empty input

PointKernel:
Related to https://github.com/rapidsai/cudf/issues/10572

This PR adds an early exit for all bulk APIs in `static_multimap`. It can get rid of the `cudaErrorInvalidConfiguration` error since the current code will try to launch a `gridDim.x == 0` kernel if the input is empty.

### DIFF
--- a/include/cuco/detail/static_multimap/static_multimap.inl
+++ b/include/cuco/detail/static_multimap/static_multimap.inl
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2022, NVIDIA CORPORATION.
  *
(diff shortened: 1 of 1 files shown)


vyasr (approved):
LGTM. Errors seem to be unrelated CI issues with sccache configuration.

PointKernel:
rerun tests

jrhemstad (approved):
One of the many things to add to my list for cuco refactoring is a kernel launch wrapper that will automate skipping kernel launches for 0 threads so we can avoid having to put these checks everywhere.

PointKernel:
> One of the many things to add to my list for cuco refactoring is a kernel launch wrapper that will automate skipping kernel launches for 0 threads so we can avoid having to put these checks everywhere.

I will file an issue for this.

PR #161, opened 2022-05-25
Add missing Thrust includes.

### LINKED ISSUE #146: Clean up `thrust` include groups
Similar to https://github.com/rapidsai/rmm/pull/1011, we should clean up include groups to adopt thrust 1.16 by using the Python fixer written by @bdice. Note this should be done before `rapids-cmake` adopts thrust 1.16.

bdice:
Closes #146 

## Description

This PR cleans up some `#include`s for Thrust. This is meant to help ease the transition to Thrust 1.16 / 1.17 when that is updated in rapids-cmake.

(I'm going to test this locally with Thrust 1.16 and will open the draft PR once I confirm it works.)

## Context

Version 1.16 of Thrust reduced the number of internal header inclusions:
> [#1572](https://github.com/NVIDIA/thrust/pull/1572) Removed several unnecessary header includes. Downstream projects may need to update their includes if they were relying on this behavior.

I am making similar changes across all RAPIDS libraries to clean up includes ("include what we use," in essence) to make sure we have compatibility with future versions of Thrust.


bdice:
I checked the logs and the new thrust was used: `CPM: adding package Thrust@1.17.0.0-rc2 (1.17.0-rc2)`.

All CI passed for commit 97044cc.

Next: wait for final Thrust 1.17.0 release, retest, then revert commits altering rapids-cmake behavior. Then this can be merged. We can update rapids-cmake once all repos are ready.

PointKernel:
@bdice Thanks!

Any ETA for the thrust upgrade in `rapids-cmake`? I need to make a cudf PR to resolve the conflicts with newer version of cuco since all sentinel values are required to be passed via `sentinel` namespace (added in #142 ) for `static_map`. This would break all use cases of `static_map` in libcudf.

bdice:
The upstream change in rapids-cmake is probably a week away from ready. The Thrust 1.17.0 release should come out next week, and we'll need to finalize some testing across all RAPIDS repos. This PR's changes to `#include`s can be merged, if we believe the Thrust 1.17.0-rc2 ➡️ 1.17.0 final update won't break includes. I think that's a safe bet. I will revert the changes to pull and test Thrust 1.17.0-rc2, then this can be reviewed and merged.

edit: It's ready for review 🚀 

Tracking PR for rapids-cmake: https://github.com/rapidsai/rapids-cmake/pull/199

PointKernel (approved):
Looks GREAT to me!

include/cuco/static_multimap.cuh
```diff
@@ -23,6 +23,7 @@
 #include <cuco/probe_sequences.cuh>
 #include <cuco/traits.hpp>
 
+#include <thrust/for_each.h>
```
PointKernel:
```suggestion
```
Unnecessary since it's only appeared in docs: https://github.com/NVIDIA/cuCollections/blob/4d36c90621d5a4165a7f61788a53b77db767f431/include/cuco/static_multimap.cuh#L649-L657

PointKernel (changes requested):

include/cuco/static_multimap.cuh
bdice: I tried not to insert includes for algorithms mentioned in docstrings but it was a manual process to verify each one, so I am not surprised that some slipped through. 😬

PointKernel (approved):

jrhemstad (approved):

PR #174, opened 2022-06-04
Make device `contains` take a template key parameter

### LINKED ISSUE #172: [FEA] Support `contains` with template key types
Currently, the `contains` APIs of various maps (in particular, `static_map` and `static_multimap`) require that their input iterators must be dereferenced into the maps' key_type. As such, we cannot insert keys of one type and check the existence of keys from another (convertible) type.

We should support `contains` with template key types.

PointKernel:
Contributes to #26

Closes #172 

This PR makes the device `contains` APIs take a template key parameter. It also includes

- update CI script to unblock CI failure
- clean up tests by using `thrust::identity`

### DIFF
--- a/include/cuco/detail/static_map.inl
+++ b/include/cuco/detail/static_map.inl
@@ -716,9 +716,11 @@ static_map<Key, Value, Scope, Allocator>::device_view::find(CG g,
 }
 
 template <typename Key, typename Value, cuda::thread_scope Scope, typename Allocator>
-template <typename Hash, typename KeyEqual>
-__device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
-  Key const& k, Hash hash, KeyEqual key_equal) const noexcept
+template <typename ProbeKey, typename Hash, typename KeyEqual>
+__device__ std::enable_if_t<std::is_convertible_v<ProbeKey, Key>, bool>
+static_map<Key, Value, Scope, Allocator>::device_view::contains(ProbeKey const& k,
+                                                                Hash hash,
+                                                                KeyEqual key_equal) const noexcept
 {
   auto current_slot = initial_slot(k, hash);
 
@@ -734,9 +736,12 @@ __device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
 }
 
 template <typename Key, typename Value, cuda::thread_scope Scope, typename Allocator>
-template <typename CG, typename Hash, typename KeyEqual>
-__device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
-  CG g, Key const& k, Hash hash, KeyEqual key_equal) const noexcept
+template <typename CG, typename ProbeKey, typename Hash, typename KeyEqual>
+__device__ std::enable_if_t<std::is_convertible_v<ProbeKey, Key>, bool>
+static_map<Key, Value, Scope, Allocator>::device_view::contains(CG const& g,
+                                                                ProbeKey const& k,
+                                                                Hash hash,
+                                                                KeyEqual key_equal) const noexcept
 {
   auto current_slot = initial_slot(g, k, hash);
 
--- a/include/cuco/detail/static_map.inl
+++ b/include/cuco/detail/static_map.inl
@@ -719,15 +719,6 @@ template <typename ProbeKey, typename Hash, typename KeyEqual>
 __device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
   ProbeKey const& k, Hash hash, KeyEqual key_equal) const noexcept
 {
-  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
-                "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
-  static_assert(std::is_invocable_r_v<bool, KeyEqual, Key, ProbeKey>,
-                "KeyEqual(Key{}, ProbeKey{}) must be a valid callable.");
-  static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, Key>,
-                "Hash(Key{}) must be a valid callable.");
-  static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, ProbeKey>,
-                "Hash(ProbeKey{}) must be a valid callable.");
-
   auto current_slot = initial_slot(k, hash);
 
   while (true) {
@@ -749,15 +740,6 @@ static_map<Key, Value, Scope, Allocator>::device_view::contains(CG const& g,
                                                                 Hash hash,
                                                                 KeyEqual key_equal) const noexcept
 {
-  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
-                "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
-  static_assert(std::is_invocable_r_v<bool, KeyEqual, Key, ProbeKey>,
-                "KeyEqual(Key{}, ProbeKey{}) must be a valid callable.");
-  static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, Key>,
-                "Hash(Key{}) must be a valid callable.");
-  static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, ProbeKey>,
-                "Hash(ProbeKey{}) must be a valid callable.");
-
   auto current_slot = initial_slot(g, k, hash);
 
   while (true) {
--- a/include/cuco/detail/static_map.inl
+++ b/include/cuco/detail/static_map.inl
@@ -716,11 +716,16 @@ static_map<Key, Value, Scope, Allocator>::device_view::find(CG g,
 
 template <typename Key, typename Value, cuda::thread_scope Scope, typename Allocator>
 template <typename ProbeKey, typename Hash, typename KeyEqual>
-__device__ std::enable_if_t<std::is_convertible_v<ProbeKey, Key>, bool>
-static_map<Key, Value, Scope, Allocator>::device_view::contains(ProbeKey const& k,
-                                                                Hash hash,
-                                                                KeyEqual key_equal) const noexcept
+__device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
+  ProbeKey const& k, Hash hash, KeyEqual key_equal) const noexcept
 {
+  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
+                "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
+  static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, Key>,
+                "Hash(Key{}) must be a valid callable.");
+  static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, ProbeKey>,
+                "Hash(ProbeKey{}) must be a valid callable.");
+
   auto current_slot = initial_slot(k, hash);
 
   while (true) {
@@ -736,12 +741,19 @@ static_map<Key, Value, Scope, Allocator>::device_view::contains(ProbeKey const& 
 
 template <typename Key, typename Value, cuda::thread_scope Scope, typename Allocator>
 template <typename CG, typename ProbeKey, typename Hash, typename KeyEqual>
-__device__ std::enable_if_t<std::is_convertible_v<ProbeKey, Key>, bool>
+__device__ std::enable_if_t<std::is_invocable_v<KeyEqual, ProbeKey, Key>, bool>
 static_map<Key, Value, Scope, Allocator>::device_view::contains(CG const& g,
                                                                 ProbeKey const& k,
                                                                 Hash hash,
                                                                 KeyEqual key_equal) const noexcept
 {
+  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
+                "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
+  static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, Key>,
+                "Hash(Key{}) must be a valid callable.");
+  static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, ProbeKey>,
+                "Hash(ProbeKey{}) must be a valid callable.");
+
   auto current_slot = initial_slot(g, k, hash);
 
   while (true) {
--- a/include/cuco/detail/static_map.inl
+++ b/include/cuco/detail/static_map.inl
@@ -721,6 +721,8 @@ __device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
 {
   static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
                 "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
+  static_assert(std::is_invocable_r_v<bool, KeyEqual, Key, ProbeKey>,
+                "KeyEqual(Key{}, ProbeKey{}) must be a valid callable.");
   static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, Key>,
                 "Hash(Key{}) must be a valid callable.");
   static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, ProbeKey>,
@@ -749,6 +751,8 @@ static_map<Key, Value, Scope, Allocator>::device_view::contains(CG const& g,
 {
   static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
                 "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
+  static_assert(std::is_invocable_r_v<bool, KeyEqual, Key, ProbeKey>,
+                "KeyEqual(Key{}, ProbeKey{}) must be a valid callable.");
   static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, Key>,
                 "Hash(Key{}) must be a valid callable.");
   static_assert(std::is_invocable_r_v<cuco::hash_value_type, Hash, ProbeKey>,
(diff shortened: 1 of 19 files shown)


PointKernel:
rerun tests

include/cuco/detail/static_map.inl
```diff
@@ -716,9 +716,11 @@ static_map<Key, Value, Scope, Allocator>::device_view::find(CG g,
 }
 
 template <typename Key, typename Value, cuda::thread_scope Scope, typename Allocator>
-template <typename Hash, typename KeyEqual>
-__device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
-  Key const& k, Hash hash, KeyEqual key_equal) const noexcept
+template <typename ProbeKey, typename Hash, typename KeyEqual>
+__device__ std::enable_if_t<std::is_convertible_v<ProbeKey, Key>, bool>
+static_map<Key, Value, Scope, Allocator>::device_view::contains(ProbeKey const& k,
```
ttnghia: Changing type of `k` may not be enough, as I see that `initial_slot` relies on `Key` type.

tests/static_map/custom_type_test.cu
```diff
@@ -89,10 +89,10 @@ struct hash_custom_key {
 
 // User-defined device key equality
 struct custom_key_equals {
-  template <typename custom_type>
-  __device__ bool operator()(custom_type lhs, custom_type rhs)
+  template <typename lhs_type, typename rhs_type>
```
ttnghia, line 93: This is not enough yet. For the places that call `map.contains`, we should create a new key array of different type (but copied the same content of the original keys array). The new key type share implementation with the old type. So we may have something similar to:
```
template<typename T>
struct key_pair_type_base {.....};

template<typename T>
struct key_pair_type_lhs : struct key_pair_type_base {};

template<typename T>
struct key_pair_type_rhs : struct key_pair_type_base {};
```

ttnghia: In addition, the test that calls `map.find` makes me wonder if we also need a template key version for `find()`?

include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -569,17 +569,19 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
    *
    * @tparam uses_vector_load Boolean flag indicating whether vector loads are used
    * @tparam CG Cooperative Group type
+   * @tparam ProbeKey Probe key type that is convertible to the map's `key_type`
```
jrhemstad: It does not need to be convertible to `key_type`. The generic definition would be that `std::invoke_result<KeyEqual, ProbeKey, Key>` must be well-formed. In other words, you just need to be able to call `key_equal(ProbeKey{}, Key{})`

jrhemstad: `invoke_result<Hash, ProbeKey>` must also be well-formed. And if `key_equal(ProbeKey, Key)` returns true, then `hash(ProbeKey) == hash(Key)` must also be true. 

See https://www.cppstories.com/2021/heterogeneous-access-cpp20/#the-performance-gains-with-heterogeneous-lookup

PointKernel: > It does not need to be convertible to key_type

Do we need `ProbeKey` being convertible to `key_type` for template `initial_slot(CG const& g, T const& k)` (required by https://github.com/NVIDIA/cuCollections/pull/174#discussion_r890385352)? My idea was to cast `T` to `key_type` when invoking the base implementation:
```cpp
  template <typename CG, typename T>
  __device__ __forceinline__ iterator initial_slot(CG const& g, T const& k) noexcept
  {
    return probe_sequence_.initial_slot(g, static_cast<Key>(k));
  }
```

> The generic definition would be that std::invoke_result<KeyEqual, ProbeKey, Key> must be well-formed

Something like this?
```cpp
static_assert(std::is_same_v<std::invoke_result_t<KeyEqual, ProbeKey, Key>, bool>);
```

> And if key_equal(ProbeKey, Key) returns true, then hash(ProbeKey) == hash(Key) must also be true.

Do we want a runtime check for this or just a statement in the doc?

ttnghia: > Do we need ProbeKey being convertible to key_type for template initial_slot(CG const& g, T const& k)

I think that is reasonable. We want to preserve `k` in `ProbeKey` type for calling to key hashing and key equal comparison. Beyond that we can cast to `Key` type without any issue.

jrhemstad: There should be no requirement on conversion from `ProbeKey` to `Key` (or vice versa). The only thing that matters is being able to invoke the hash and key_equal functions.

tests/static_map/custom_type_test.cu
PointKernel: > In addition, the test that calls `map.find` makes me wonder if we also need a template key version for `find()`?

I would prefer to make this PR dedicated to solving issue #172 unless template `find` is also a blocker for your current work. Fully supporting heterogeneous lookups #26 will be a separate PR that addresses all lookup functions.

include/cuco/detail/static_multimap/device_view_impl.inl
jrhemstad: > Do we want a runtime check for this or just a statement in the doc?

That would just be a stated pre-condition in the docs.

jrhemstad: `std::is_invocable_v<KeyEqual, ProbeKey, Key>` and
`std::is_invocable_v<Hash, ProbeKey>` and `std::is_invocable_v<Hash,Key>` is what you want.

https://en.cppreference.com/w/cpp/types/is_invocable

PointKernel: Maybe `is_invocable_r_v`?
```cpp
// Determines whether Fn can be invoked with the arguments
// ArgTypes... to yield a result that is convertible to R
template <class R, class Fn, class... ArgTypes>
struct is_invocable_r;
```

jrhemstad: Yeah, that would be even better.

tests/static_map/custom_type_test.cu
PointKernel: @ttnghia I think the new test should be sufficient. Please confirm.

include/cuco/detail/static_map.inl
```diff
@@ -715,10 +715,17 @@ static_map<Key, Value, Scope, Allocator>::device_view::find(CG g,
...
 
 template <typename Key, typename Value, cuda::thread_scope Scope, typename Allocator>
-template <typename Hash, typename KeyEqual>
+template <typename ProbeKey, typename Hash, typename KeyEqual>
 __device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
-  Key const& k, Hash hash, KeyEqual key_equal) const noexcept
+  ProbeKey const& k, Hash hash, KeyEqual key_equal) const noexcept
 {
+  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
+                "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
```
ttnghia: Let's add this for symmetry:
```suggestion
  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
                "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
  static_assert(std::is_invocable_r_v<bool, KeyEqual, Key, ProbeKey>,
                "KeyEqual(Key{}, ProbeKey{}) must be a valid callable.");
```

include/cuco/detail/static_map.inl
```diff
@@ -733,10 +740,20 @@ __device__ bool static_map<Key, Value, Scope, Allocator>::device_view::contains(
...
-  CG g, Key const& k, Hash hash, KeyEqual key_equal) const noexcept
+template <typename CG, typename ProbeKey, typename Hash, typename KeyEqual>
+__device__ std::enable_if_t<std::is_invocable_v<KeyEqual, ProbeKey, Key>, bool>
+static_map<Key, Value, Scope, Allocator>::device_view::contains(CG const& g,
+                                                                ProbeKey const& k,
+                                                                Hash hash,
+                                                                KeyEqual key_equal) const noexcept
 {
+  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
+                "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
```
ttnghia: Similar to above.

include/cuco/detail/static_multimap/static_multimap.inl
```diff
@@ -536,13 +536,37 @@ template <typename Key,
...
+template <typename ProbeKey, typename KeyEqual>
 __device__ __forceinline__ bool
 static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view::contains(
   cooperative_groups::thread_block_tile<ProbeSequence::cg_size> const& g,
-  Key const& k,
+  ProbeKey const& k,
   KeyEqual key_equal) noexcept
 {
+  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
+                "KeyEqual(ProbeKey{}, Key{}) must be a valid callable.");
```
ttnghia: Similar to above.

include/cuco/static_map.cuh
```diff
@@ -1247,12 +1265,14 @@ class static_map {
...
      */
     template <typename CG,
+              typename ProbeKey,
               typename Hash     = cuco::detail::MurmurHash3_32<key_type>,
               typename KeyEqual = thrust::equal_to<key_type>>
-    __device__ bool contains(CG g,
-                             Key const& k,
-                             Hash hash          = Hash{},
-                             KeyEqual key_equal = KeyEqual{}) const noexcept;
+    __device__ std::enable_if_t<std::is_invocable_v<KeyEqual, ProbeKey, Key>, bool> contains(
```
ttnghia, line 1280:
```suggestion
    __device__ std::enable_if_t<std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key> && std::is_invocable_r_v<bool, KeyEqual, Key, ProbeKey>, bool> contains(
```

tests/static_map/heterogeneous_lookup_test.cu
```diff
@@ -0,0 +1,122 @@
...
+  auto const sentinel_key   = Key{-1};
+  auto const sentinel_value = Value{-1};
+
+  constexpr std::size_t num      = 100;
+  constexpr std::size_t capacity = num * 2;
+  cuco::static_map<Key, Value> map{capacity,
+                                   cuco::sentinel::empty_key<Key>{sentinel_key},
+                                   cuco::sentinel::empty_value<Value>{sentinel_value}};
+
+  auto insert_keys = thrust::make_transform_iterator(
```
ttnghia: We insert keys and values at the same time, not just keys.
```suggestion
  auto insert_kv = thrust::make_transform_iterator(
```

tests/static_multimap/heterogeneous_lookup_test.cu
```diff
@@ -0,0 +1,125 @@
...
+  cuco::static_multimap<Key,
+                        Value,
+                        cuda::thread_scope_device,
+                        cuco::cuda_allocator<char>,
+                        cuco::linear_probing<1, custom_hasher>>
+    map{capacity,
+        cuco::sentinel::empty_key<Key>{sentinel_key},
+        cuco::sentinel::empty_value<Value>{sentinel_value}};
+
+  auto insert_keys = thrust::make_transform_iterator(
```
ttnghia:
```suggestion
  auto insert_kv = thrust::make_transform_iterator(
```

include/cuco/static_map.cuh
PointKernel: The condition here is not a requirement check. It's just to differentiate CG API and non-CG API cause if we pass three template parameters to `contains`, the compiler cannot know which one we are referring to. e.g. https://github.com/PointKernel/cuCollections/blob/3b0adf597ed828f3813030452fb39f9b1735c90b/tests/static_map/custom_type_test.cu#L217

ttnghia:
This looks good to me. I will try to test and will update.

ttnghia:
Update: I have successfully adopted this: https://github.com/rapidsai/cudf/pull/10656 ([code](https://github.com/rapidsai/cudf/pull/10656/files#diff-8d714835115e07e9f7ac9cc8a3e095fbda10768b15a86be65116388f1a4bfd46R95)).

Thanks again for working on it.

include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -69,13 +71,16 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
...
+   *
    * @param g the Cooperative Group for which the initial slot is needed
    * @param k The key to get the slot for
    * @return Pointer to the initial slot for `k`
    */
-  template <typename CG>
-  __device__ __forceinline__ iterator initial_slot(CG const& g, Key const& k) noexcept
+  template <typename ProbeKey>
+  __device__ __forceinline__ iterator
+  initial_slot(cooperative_groups::thread_block_tile<ProbeSequence::cg_size> const& g,
```
jrhemstad, line 82: Don't we already have a namespace alias for `cg = cooperative_groups` somewhere? If not, let's add it to make this less verbose.

tests/static_map/heterogeneous_lookup_test.cu
```diff
@@ -0,0 +1,122 @@
...
+  constexpr std::size_t capacity = num * 2;
+  cuco::static_map<Key, Value> map{capacity,
+                                   cuco::sentinel::empty_key<Key>{sentinel_key},
+                                   cuco::sentinel::empty_value<Value>{sentinel_value}};
+
+  auto insert_pairs = thrust::make_transform_iterator(
+    thrust::counting_iterator<int>(0),
+    [] __device__(auto i) { return cuco::pair_type<Key, Value>(i, i); });
+  auto probe_keys = thrust::make_transform_iterator(
+    thrust::counting_iterator<int>(0), [] __device__(auto i) { return key_triplet<Value>(i); });
```
jrhemstad: Shouldn't `key_triplet` bet templated on `Key` not `Value`?

tests/static_map/heterogeneous_lookup_test.cu
```diff
@@ -0,0 +1,122 @@
...
+      contained.begin(), contained.end(), [] __device__(bool const& b) { return b; }));
+  }
+
+  SECTION("Non-inserted keys-value pairs should not be contained")
+  {
+    thrust::device_vector<bool> contained(num);
+    map.contains(
+      probe_keys, probe_keys + num, contained.begin(), custom_hasher{}, custom_key_equal{});
+    REQUIRE(cuco::test::none_of(
+      contained.begin(), contained.end(), [] __device__(bool const& b) { return b; }));
```
jrhemstad:
```suggestion
      contained.begin(), contained.end(), thrust::identity{});
```

tests/static_map/heterogeneous_lookup_test.cu
```diff
@@ -0,0 +1,122 @@
...
+    thrust::counting_iterator<int>(0), [] __device__(auto i) { return key_triplet<Value>(i); });
+
+  SECTION("All inserted keys-value pairs should be contained")
+  {
+    thrust::device_vector<bool> contained(num);
+    map.insert(insert_pairs, insert_pairs + num, custom_hasher{}, custom_key_equal{});
+    map.contains(
+      probe_keys, probe_keys + num, contained.begin(), custom_hasher{}, custom_key_equal{});
+    REQUIRE(cuco::test::all_of(
+      contained.begin(), contained.end(), [] __device__(bool const& b) { return b; }));
```
jrhemstad:
```suggestion
      contained.begin(), contained.end(), thrust::identity{});
```

include/cuco/detail/static_multimap/static_multimap.inl
```diff
@@ -536,13 +536,39 @@ template <typename Key,
...
-template <typename KeyEqual>
+template <typename ProbeKey, typename KeyEqual>
 __device__ __forceinline__ bool
 static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view::contains(
   cooperative_groups::thread_block_tile<ProbeSequence::cg_size> const& g,
-  Key const& k,
+  ProbeKey const& k,
   KeyEqual key_equal) noexcept
 {
+  static_assert(std::is_invocable_r_v<bool, KeyEqual, ProbeKey, Key>,
```
jrhemstad: tbh, all these static_asserts are overkill. If someone passes in an invalid callable, it will still fail to compile, just with a different error message. This feels like a ton of complexity to maintain.

jrhemstad: I pointed to `is_invocable` just for the purposes of documentation. I didn't intend that we need to `static_assert` all of the requirements of the function.

tests/static_map/heterogeneous_lookup_test.cu
PointKernel: The naming was a bit misleading. Corrected.

include/cuco/detail/static_multimap/static_multimap.inl
PointKernel: I've removed those `static_assert` as requested and updated corresponding docs.

I agree that so many asserts are painful to maintain now but the situation will be improved after the grand refactor where we only need one set of those static asserts in the base class as opposed to putting them in every involved function for all map types.

include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -637,9 +634,11 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
    * @return A boolean indicating whether the key/value pair
    * containing `k` was inserted
    */
-  template <bool uses_vector_load, typename CG, typename ProbeKey, typename KeyEqual>
+  template <bool uses_vector_load, typename ProbeKey, typename KeyEqual>
   __device__ __forceinline__ std::enable_if_t<not uses_vector_load, bool> contains(
-    CG const& g, ProbeKey const& k, KeyEqual key_equal) noexcept
+    detail::cg::thread_block_tile<ProbeSequence::cg_size> const& g,
```
jrhemstad: Wait, `detail::cg`?

PointKernel: Reverted back to the `cooperative_groups` naming based on offline discussion.

jrhemstad (approved):

PR #203, opened 2022-08-08
Always download RAPIDS.cmake to ensure dependencies don't overwrite

robertmaynard:
When you have dependencies that use RAPIDS.cmake they could overwrite the file and cause a different version to be used when cmake is executed a second time. Instead always download the file to ensure you have the correct version.

Fixes #https://github.com/rapidsai/rapids-cmake/issues/240

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,10 +15,8 @@
 #=============================================================================
 cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
 
-if(NOT EXISTS ${CMAKE_BINARY_DIR}/RAPIDS.cmake)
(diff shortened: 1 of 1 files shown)


sleeepyjack:
The problem is that nvbench downloads the `RAPIDS.cmake` file every time:

https://github.com/NVIDIA/nvbench/blob/9630a081e6ea508d310c661d91d31aa71d672b82/cmake/NVBenchRapidsCMake.cmake#L3-#L6

Forcing the download every time should fix this.
Thanks @robertmaynard !

sleeepyjack (approved):

PointKernel (approved):

PR #204, opened 2022-08-08
Use rapids-cmake 22.10 best practice for RAPIDS.cmake location

robertmaynard:
Removes possibility of another projects RAPIDS.cmake being used, and removes need to always download a version.


robertmaynard:
With https://github.com/NVIDIA/cuCollections/pull/203 merged I had more time to look at what the best way to setup RAPIDS.cmake downloads. This is the current solution I am deploying to all rapids-cmake consumers ( that I know about ).

PR #225, opened 2022-08-30
Upgrade rapids-cmake

PointKernel:
This PR upgrades `rapids-cmake` to `branch-22.10` so `cuda/std/array` from libcudacxx-1.8 can be used in cuco.


tests/static_map/insert_and_find_test.cu
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2022, Jonas Hahnfeld, CERN.
+ * Copyright (c) 2022, NVIDIA CORPORATION.
```
sleeepyjack, line 3: ;)

sleeepyjack (approved):

PR #234, opened 2022-09-20
Enable `memcpy_async` execution path

sleeepyjack:
The following snippet uses an incorrect `CUDART_VERSION` scheme:
```cpp
#if defined(CUDART_VERSION) && (CUDART_VERSION >= 11100)
#define CUCO_HAS_CG_MEMCPY_ASYNC
#endif
```
`11100` maps to the non-existing CUDA CTK version 11._10_ instead of 11.1.

Reproducer: https://godbolt.org/z/WzfsPcPoE

### DIFF
--- a/include/cuco/detail/static_multimap/device_view_impl.inl
+++ b/include/cuco/detail/static_multimap/device_view_impl.inl
@@ -498,8 +498,13 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
     }
     offset = g.shfl(offset, 0);
 
-    if constexpr (thrust::is_contiguous_iterator_v<OutputIt>) {
 #if defined(CUCO_HAS_CG_MEMCPY_ASYNC)
+    constexpr bool uses_memcpy_async = thrust::is_contiguous_iterator_v<OutputIt>;
+#else
+    constexpr bool uses_memcpy_async = false;
+#endif // end CUCO_HAS_CG_MEMCPY_ASYNC
+
+    if constexpr (uses_memcpy_async) {
 #if defined(CUCO_HAS_CUDA_BARRIER)
       cooperative_groups::memcpy_async(
         g,
@@ -512,9 +517,9 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
                                        output_buffer,
                                        sizeof(value_type) * num_outputs);
 #endif  // end CUCO_HAS_CUDA_BARRIER
-#endif  // end CUCO_HAS_CG_MEMCPY_ASYNC
     }
-    if constexpr (not thrust::is_contiguous_iterator_v<OutputIt>) {
+
+    if constexpr (not uses_memcpy_async) {
       for (auto index = lane_id; index < num_outputs; index += g.size()) {
         *(output_begin + offset + index) = output_buffer[index];
       }
--- a/include/cuco/detail/static_multimap/device_view_impl.inl
+++ b/include/cuco/detail/static_multimap/device_view_impl.inl
@@ -502,7 +502,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
     constexpr bool uses_memcpy_async = thrust::is_contiguous_iterator_v<OutputIt>;
 #else
     constexpr bool uses_memcpy_async = false;
-#endif // end CUCO_HAS_CG_MEMCPY_ASYNC
+#endif  // end CUCO_HAS_CG_MEMCPY_ASYNC
 
     if constexpr (uses_memcpy_async) {
 #if defined(CUCO_HAS_CUDA_BARRIER)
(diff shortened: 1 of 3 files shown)


PointKernel:
Good catch. Looks like we need to include the missing header `#include <cooperative_groups/memcpy_async.h>` to make the code work with cuda-11.2.

sleeepyjack:
I tried including the missing header but the error still persists:
```
include/cuco/detail/static_multimap/device_view_impl.inl(504): error: no instance of overloaded function "cooperative_groups::memcpy_async" matches the argument list
            argument types are: (const cooperative_groups::__v1::thread_block_tile<8U, cooperative_groups::__v1::thread_block>, thrust::detail::normal_iterator<thrust::device_ptr<cuco::pair<int64_t, int64_t>>>, cuco::pair<int64_t, int64_t> *, cuda::__4::aligned_size_t<16UL>)
```

Seems like the iterator isn't convertible to `value_type*`.

include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -502,19 +503,23 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
...
+                                       sizeof(value_type) * num_outputs);
 #endif  // end CUCO_HAS_CUDA_BARRIER
       return;
 #endif  // end CUCO_HAS_CG_MEMCPY_ASYNC
     }
+#pragma nv_diag_suppress 128  // warning: unreachable
     for (auto index = lane_id; index < num_outputs; index += g.size()) {
       *(output_begin + offset + index) = output_buffer[index];
     }
+#pragma nv_diag_default 128
```
sleeepyjack: I thought I had fixed the following warning with the `nv_diag_suppress`:
```
include/cuco/detail/static_multimap/device_view_impl.inl(519): warning: loop is not reachable
```
but it still pops up.

I'll see if I can reorder the logic in order to fix that.

PointKernel:
```suggestion
#endif  // end CUCO_HAS_CUDA_BARRIER
#endif  // end CUCO_HAS_CG_MEMCPY_ASYNC
    }
    if constexpr (not thrust::is_contiguous_iterator_v<OutputIt>) {
      for (auto index = lane_id; index < num_outputs; index += g.size()) {
        *(output_begin + offset + index) = output_buffer[index];
      }
    }
```
Changing the logic a bit may fix the issue.

sleeepyjack: lol, yeah I was just about to type that.

sleeepyjack: Ah, wait, this way the function does nothing for the case `not defined(CUCO_CG_MEMCPY_ASYNC) and thrust::is_contiguous_iterator_v<OutputIt>`.

PointKernel: dang... you are right

sleeepyjack:
```cpp
#if defined(CUCO_HAS_CG_MEMCPY_ASYNC)
    constexpr bool uses_memcpy_async = thrust::is_contiguous_iterator_v<OutputIt>;
#else
    constexpr bool uses_memcpy_async = false;
#endif

    if constexpr (uses_memcpy_async) {
#if defined(CUCO_HAS_CUDA_BARRIER)
      cooperative_groups::memcpy_async(
        g,
        &thrust::raw_reference_cast(*(output_begin + offset)),
        output_buffer,
        cuda::aligned_size_t<alignof(value_type)>(sizeof(value_type) * num_outputs));
#else
      cooperative_groups::memcpy_async(g,
                                       &thrust::raw_reference_cast(*(output_begin + offset)),
                                       output_buffer,
                                       sizeof(value_type) * num_outputs);
#endif  // end CUCO_HAS_CUDA_BARRIER
    }

    if constexpr (not uses_memcpy_async) {
      for (auto index = lane_id; index < num_outputs; index += g.size()) {
        *(output_begin + offset + index) = output_buffer[index];
      }
    }
```

PointKernel (approved):

include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -497,23 +498,31 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
...
+#else
+    constexpr bool uses_memcpy_async = false;
+#endif  // end CUCO_HAS_CG_MEMCPY_ASYNC
+
+    if constexpr (uses_memcpy_async) {
 #if defined(CUCO_HAS_CUDA_BARRIER)
       cooperative_groups::memcpy_async(
         g,
-        output_begin + offset,
+        &thrust::raw_reference_cast(*(output_begin + offset)),
```
sleeepyjack, line 511: I'd like @jrhemstad to confirm that this line isn't too whack before I merge this PR.

PR #235, opened 2022-09-23
Document that minimum required CMake version is now 3.23.1

robertmaynard:
With rapids-cmake now requiring CMake 3.23.1 update consumers to correctly express this requirement

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #=============================================================================
-cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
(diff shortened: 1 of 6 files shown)


PointKernel:
@robertmaynard Just to make sure

https://github.com/NVIDIA/cuCollections/blob/77593b39aa72ad859d6196e0a04b40a79db5a794/CMakeLists.txt#L19

`22.10` is the version requiring `cmake-3.23.1`, right?

robertmaynard:
> @robertmaynard Just to make sure
> 
> https://github.com/NVIDIA/cuCollections/blob/77593b39aa72ad859d6196e0a04b40a79db5a794/CMakeLists.txt#L19
> 
> `22.10` is the version requiring `cmake-3.23.1`, right?

Yes

PointKernel (approved):

Issue #239, opened 2022-11-03
[ENHANCEMENT]: CRTP/mixins to encode concurrent device-side operations


PointKernel:
### Is your feature request related to a problem? Please describe.

```cpp
static_set_ref<insert,contains> map{...};

template <typename Mixins...>
struct static_set_ref : Mixins<static_set_ref>... {};
```
An individual mix-in type would need to query what other types are present in `Mixins...` and then statically dispatch to the appropriate implementation. E.g., `insert` for 8B/8B would use back2back CAS vs cas_dependent_write vs cast_dependent_write_spin depending on whether or not `find` or `insert_and_find` are present in the `Mixins...` pack. 

### Describe the solution you'd like

https://godbolt.org/z/EPvc3Ps88 (example of `static_set_ref`)
https://godbolt.org/z/enW78nP7a (simple mixin example)
https://godbolt.org/z/Mv9W7ooYf (Customizing behavior based on different mixin combinations)
https://godbolt.org/z/srGnPEb7c (Customizing behavior based on different mixin combinations, simpler version)

TODO:

- [ ] non-type template mixin where operators are a `enum class, e.g.
```cpp
enum class operator : int32_t {
  NONE = 0,
  INSERT,
  CONTAINS,
  ...
};
```

### Additional context

References:

https://www.modernescpp.com/index.php/mixins
https://www.fluentcpp.com/2017/12/12/mixin-classes-yang-crtp/
https://stackoverflow.com/questions/26486456/two-different-mixin-patterns-in-c-mixin-crtp

sleeepyjack:
Solved by #246

PR #246, opened 2022-11-23
Function mixins for static_set_ref

sleeepyjack:
The main part of this PR consists of adding the new function mixin design (#239) to the `static_set_ref` class.

I went with the following design:
- Each function is exposed as an empty tag struct that is available inside the cuco namespace, e.g., `cuco::(function::)insert`. We then pass a list of desired function tags to the `static_set_ref` class template. For each tag, `static_set_ref` inherits a function implementation that is specialized on the tag as well as the reference type: `: public function_impl<Tag, Reference>`.
- ~~Define a common base class `open_address_ref` for `static_set_ref`, `static_map_ref`, `static_multimap_ref`. Since functions such as `insert`/`find`/`contains` for these containers are more or less identical (differences are for instance if we need to continue probing after the first key hit, or if we insert a key or a key-value pair), we can implement these functions for the base class and then dispatch the correct behavior internally. This way we only need to maintain a single slightly more complex implementation -> less code duplication.~~ Removed for now.
- For functions that should only exist for one container type but not the others, we can still define a specialization that has nothing to do with the common base class.
- Add functionality to change the set of functions available to a ref object/type, e.g., `static_set_ref::make_with_functions<...>` and `static_set_ref::with_functions<...>`

What this design brings to the table:
- Functions can either be defined for all containers or just a single container.
- We can select the optimal code path given the current reference type and the set of other functions available to the reference object.
- ~~Less code duplication by unifying common implementations under the new base class.~~


sleeepyjack:
GH won't let me fix the typo in the title -.-

sleeepyjack:
ok to test

PointKernel:
rerun test

AyodeAwe:
rerun tests

include/cuco/sentinel.cuh
```diff
@@ -17,7 +17,7 @@
 #pragma once
 
 namespace cuco {
-namespace sentinel {
+inline namespace sentinel {
```
PointKernel: This could be a PR against `dev` branch.

sleeepyjack: /done #247

sleeepyjack:
I have removed the open address base class for now so this PR gets a bit easier to review.

include/cuco/static_set.cuh
```diff
@@ -195,20 +192,38 @@ class static_set {
...
 
+  /**
+   * @brief Get device reference (alias definition).
+   *
+   * @tparam Functions Set of `cuco::functions` to be provided by the reference
+   *
+   * @return Device reference of the current static_set
+   */
+  template <typename... Functions>
+  auto reference_with_functions() const noexcept
```
jrhemstad: Bikeshed: I'd opt to just call this `reference_with` or `ref_with`. I think it will read more concisely and still be intuitive:

```
auto ref = set.ref_with<insert, find>();
```

include/cuco/static_set_ref.cuh
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * @brief Create a reference with new functions from the current object.
+   *
+   * @tparam NewFunctions List of `cuco::function::*` types
+   *
+   * @return copy of `*this` with `newFunctions`
+   */
+  template <typename... NewFunctions>
+  [[nodiscard]] __host__ __device__ auto with_functions() const
```
jrhemstad: We'll want to provide solid documentation about the restriction against concurrent use of references to the same storage with different functions.

include/cuco/static_set_ref.cuh
```diff
@@ -0,0 +1,155 @@
...
+      static_set_ref<Key, Scope, KeyEqual, ProbingScheme, StorageRef, Functions...>>... {
+ public:
+  using key_type            = Key;                             ///< Key Type
+  using probing_scheme_type = ProbingScheme;                   ///< Type of probing scheme
+  using storage_ref_type    = StorageRef;                      ///< Type of slot storage ref
+  using window_type = typename storage_ref_type::window_type;  ///< Probing scheme element type
+  using value_type  = typename storage_ref_type::value_type;   ///< Probing scheme element type
+  using size_type   = typename storage_ref_type::size_type;    ///< Probing scheme size type
+  using key_equal =
+    detail::equal_wrapper<value_type, KeyEqual>;  ///< Type of key equality binary callable
```
jrhemstad: I don't think this is right. The `key_equal` typedef should just be `KeyEqual`. However, the data member stored for `key_compare_` can be `equal_wrapper<value_type, KeyEqual>`.

include/cuco/static_set_ref.cuh
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * @brief Device reference of static_set.
+ */
+template <typename Key,
+          cuda::thread_scope Scope,
+          typename KeyEqual,
+          typename ProbingScheme,
+          typename StorageRef,
+          typename... Functions>
```
jrhemstad: I don't know why, but "Operators" feels like a more appropriate name. "Functions" feels too generic like it can be any function.

sleeepyjack: I went with Operators for like half of all my prototypes then somehow switched to Functions. The naming struggle is real 😅

Wikipedia says:
> In mathematics, an operator is generally a mapping or function that acts on elements of a space to produce elements of another space (possibly the same space, sometimes required to be the same space). There is no general definition of an operator, but the term is often used in place of function when the domain is a set of functions or other structured objects.

This sounds to me like Operators is the right choice. Or not?

include/cuco/static_set_ref.cuh
sleeepyjack: Sure, otherwise this is a proper footgun. I would even suggest to add the op/op compatibility matrix that I'm working on to the documentation at some point. Could even be part of some sort of tuning guide.

include/cuco/detail/bitwise_compare.cuh
```diff
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <cuco/traits.hpp>
```
PointKernel: I just created #248 to fix this bug since the whole chain of refactoring PRs would take a while to merge into the default branch. Thanks for catching this issue!

examples/static_set/device_reference_example.cu
```diff
@@ -0,0 +1,116 @@
...
+  custom_cooperative_insert<<<128, 128>>>(
+    set.reference(), keys.begin() + num_keys / 2, num_keys / 2);
+
+  // Storage for result
+  thrust::device_vector<bool> found(num_keys);
+
+  // Check if all keys are now contained in the set. Note that we pass a reference that already has
+  // the `contains` functions
+  custom_contains<<<128, 128>>>(
+    set.template reference<cuco::experimental::contains>(), keys.begin(), num_keys, found.begin());
```
PointKernel: I feel like we remove `static_set::reference()` and exclusively use new `ref_with<...>` to create ref from a given data structure.

Then this example helps to show how multiple operators work together. e.g.:
```cpp
  auto my_set = set_type{capacity, cuco::empty_key{empty_key_sentinel}};
  auto d_ref  = my_set.ref_with<insert, contains>( ... );
  custom_cooperative_insert<<<128, 128>>>(d_ref, keys.begin() + num_keys, num_keys);
  custom_contains<<<128, 128>>>(d_ref, keys.begin(), num_keys, found.begin());
```

examples/static_set/device_reference_example.cu
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
```
PointKernel:
```suggestion
 * Copyright (c) 2022, NVIDIA CORPORATION.
```

examples/static_set/host_bulk_example.cu
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
```
PointKernel:
```suggestion
 * Copyright (c) 2022, NVIDIA CORPORATION.
```

include/cuco/detail/static_set/static_set.inl
```diff
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 
-#include <cuco/detail/defaults.cuh>
 #include <cuco/detail/error.hpp>
 #include <cuco/detail/prime.hpp>
 #include <cuco/detail/static_set/kernels.cuh>
-#include <cuco/reference.cuh>
+#include <cuco/detail/tuning.cuh>  // TODO .hpp?
```
PointKernel: So far those constexpr are accessed from host only so yeah, it could be a `hpp`

include/cuco/detail/static_set/static_set.inl
```diff
@@ -71,7 +71,7 @@ void static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>
 
   detail::insert<cg_size, detail::CUCO_DEFAULT_BLOCK_SIZE>
     <<<grid_size, detail::CUCO_DEFAULT_BLOCK_SIZE, 0, stream>>>(
-      first, first + num_keys, reference());
+      first, first + num_keys, reference_with_functions<function::insert>());
```
PointKernel: When users invoke host bulk functions, should we assume that they will combine any arbitrary (or all) operators together? For this case, what if the user launch bulk `insert` and `find` back to back with different streams and they expect to perform concurrent insert and find?

include/cuco/detail/static_set/static_set_ref.inl
```diff
@@ -0,0 +1,282 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cuco/detail/equal_wrapper.cuh>
+#include <cuco/detail/pair.cuh>
+#include <cuco/function.hpp>
+#include <cuco/sentinel.cuh>  // TODO .hpp
```
PointKernel: Good point. It seems sentinel wrappers are only used on the host side.

Would be a separate PR though.

include/cuco/detail/storage.cuh
```diff
@@ -191,21 +192,21 @@ class aos_storage_ref {
...
-   */
+   */ // TODO should be named data()
   __device__ inline window_type* windows() const noexcept { return windows_; }
 
   /**
    * @brief Gets the total number of slot windows in the current storage.
    *
    * @return The total number of slot windows
-   */
+   */ // TODO should be named windows()
```
PointKernel: I was thinking of using `size` instead of `num_windows` to denote the number of windows in the storage and using `capacity` to denote the number of slots (which is aligned with set/map::capacity). One concern of doing this is that `size` is not as descriptive as `num_windows` and maybe (or not) easy to confuse with `capacity`. Another issue is that `size` has a specific meaning in the map/set context which denotes the number of filled slots. In the end, I went with `num_windows`.

include/cuco/detail/storage.cuh
```diff
@@ -191,21 +192,21 @@ class aos_storage_ref {
    * @brief Gets windows array.
    *
    * @return Pointer to the first window
-   */
+   */ // TODO should be named data()
```
PointKernel: If we go with `data`, the storage ref will be exclusively an array of windows aka `aow_storage_ref`.

include/cuco/static_set.cuh
```diff
@@ -102,11 +95,15 @@ class static_set {
   using key_equal           = KeyEqual;                          ///< Key equality comparator type
   using allocator_type      = Allocator;                         ///< Allocator type
   using window_storage_type = Storage;                           ///< Window storage type
-  /// Window storage reference type
-  using window_reference_type = typename window_storage_type::reference_type;
-  using probing_scheme_type   = ProbingScheme;  ///< Probe scheme type
-  using counter_storage_type =
-    detail::counter_storage<size_type, Scope, Allocator>;  ///< Counter storage type
+  using window_reference_type =
```
PointKernel: maybe `storage_ref_type`

include/cuco/static_set_ref.cuh
PointKernel: Either `Operators` or `Operations` sounds good to me.

include/cuco/static_set_ref.cuh
PointKernel: Good point.

examples/static_set/device_reference_example.cu
sleeepyjack: I would even go one step further:
Define a free function something like `auto cuco::with_ops<Container, Functions...>(Container&& c)` and specialize this function for all containers.
The reason behind this is that I don't like the call syntax when the function is a member of the container: `ref.template with_ops<Functions...>()`.

examples/static_set/device_reference_example.cu
sleeepyjack: I wonder if we could automate this task with pre-commit: https://github.com/fwolf/copyright-year-updater.sh

include/cuco/detail/static_set/static_set.inl
sleeepyjack: We've been asked this question before: Do we want to provide plain cpp headers for the public interface so we can link to cuco in some separate compilation scenarios? I'm not sure how to make this work for all files. Let's discuss if we really want/need that feature. If not, I would stick with .cuh for sake of consistency.

include/cuco/detail/storage.cuh
sleeepyjack: I guess I hastily added those comments thinking we would switch to `cuda::std::array` as the underlying storage type. We should definitely talk about the storage class next but my comments are outdated.

include/cuco/detail/static_set/static_set.inl
sleeepyjack: This problem has been on my mind for weeks. What should the default behavior look like? All function or no functions? Or should we even implement a similar mixin function approach for the host API too? @jrhemstad What are your thoughts on this? We might want to discuss this in our next sync.

examples/static_set/device_reference_example.cu
PointKernel: Yes, cudf has a similar script and we can take it for our use.

include/cuco/detail/static_set/static_set.inl
PointKernel: There is an issue with this request: #232.

include/cuco/detail/static_set/static_set.inl
jrhemstad: I can't imagine a sane use case for concurrent bulk operations, so let's just defer this until when/if the need ever arises.

examples/static_set/device_reference_example.cu
```diff
@@ -34,8 +34,8 @@ __global__ void custom_cooperative_insert(SetRef raw_set, InputIterator keys, st
 
   constexpr auto cg_size = SetRef::cg_size;
 
-  // we haven't spcified any functions yet so we make a copy with the desired functionality
-  auto set = raw_set.template with_functions<cuco::experimental::insert>();
+  // we haven't spcified any operatorss yet so we make a copy with the desired functionality
+  auto set = raw_set.template with<cuco::experimental::insert>();
```
jrhemstad: Oh, gross. I didn't think about how this design would lead to needing a template disambiguator.

The usual solution to avoid this is to make the function a free function instead of a member function. That is why things like `std::get` are a free function and not a member of `tuple`.

I'm not sure if applies as well to our situation, but I really dislike that this will end up requiring most users to use a template disambiguator.

PointKernel: Yeah, @sleeepyjack also mentioned this issue https://github.com/NVIDIA/cuCollections/pull/246#discussion_r1038905868

sleeepyjack: Right. I'll switch to a free function.

sleeepyjack: Oh, this is going to be trickier than I thought. We would need to partially specialize `cuco::with<Reference, Operators...>` on the reference type, which is not possible with function templates. Let's see if I can make it work with a functor-like thing.

sleeepyjack: Adressed in [56086a5](https://github.com/NVIDIA/cuCollections/pull/246/commits/56086a5a29d0c531ccc54eab23d946a04a0f8f10)

sleeepyjack:
I just rebased with the `dev` branch. The only new commits are
[Refactor with() and ref_with()](https://github.com/NVIDIA/cuCollections/pull/246/commits/56086a5a29d0c531ccc54eab23d946a04a0f8f10)
and
[Use move semantics for with()](https://github.com/NVIDIA/cuCollections/pull/246/commits/55134fe3c8e963aee74ac3011f8b02ed11c1d590)

include/cuco/detail/static_set/static_set_ref.inl
```diff
@@ -0,0 +1,300 @@
...
+  template <typename ProbeKey>
+  [[nodiscard]] __device__ inline bool contains(ProbeKey const& key) const noexcept
+  {
+    // CRTP: cast `this` to the actual reference type
+    auto const& ref_ = static_cast<ref_type const&>(*this);
+
+    auto probing_iter = ref_.probing_scheme_(key, ref_.storage_ref_.num_windows());
+
+    while (true) {
+      // TODO do we need to use atomic_ref::load here?
```
sleeepyjack, line 234: @jrhemstad would this be needed in terms of correctness, in case a mutation operator like `cuco::insert` is enabled?

include/cuco/detail/static_set/static_set_ref.inl
```diff
@@ -0,0 +1,300 @@
...
+    auto probing_iter = ref_.probing_scheme_(group, value, ref_.storage_ref_.num_windows());
+
+    while (true) {
+      auto const window_slots = ref_.storage_ref_.window(*probing_iter);
+
+      auto const [state, intra_window_index] = [&]() {
+        for (auto i = 0; i < window_size; ++i) {
+          switch (ref_.predicate_(window_slots[i], value)) {
+            case detail::equal_result::EMPTY:
+              return cuco::pair<detail::equal_result, int32_t>{detail::equal_result::EMPTY, i};
```
jrhemstad: CTAD is your friend. 
```suggestion
              return cuco::pair{detail::equal_result::EMPTY, i};
```

include/cuco/detail/static_set/static_set_ref.inl
jrhemstad: Ideally we'd customize the implementation of `contains` such that it uses an atomic load when an `insert` operation is also present and otherwise do non-atomic loads.

sleeepyjack: That's what I had in mind 👍

PointKernel: For concurrent `insert` and `find`, I can see atomic loads are necessary but do we really need atomic loads for `contains` or `count`?

jrhemstad: concurrent `insert` and `contains` is no different than concurrent `insert` and `find`.

sleeepyjack: The bad thing about atomic loads is that we probably won't be able to use window probing as each window entry would translate to a single atomic load instruction.

We could always use relaxed loads + windows during probing and when we find a candidate slot we properly/atomically load it again to make sure we actually found a matching key. Not sure how this impacts performance.

include/cuco/detail/static_set/static_set_ref.inl
sleeepyjack: Addressed in [15f2752](https://github.com/NVIDIA/cuCollections/pull/246/commits/15f27523d42857c6900b5c00700a2fd668049ae9)

include/cuco/static_set_ref.cuh
sleeepyjack: I hope [03adc20](https://github.com/NVIDIA/cuCollections/pull/246/commits/03adc20c51f47bcf31a96ed91182daabc1313f52) sounds dramatic enough :D

examples/static_set/device_reference_example.cu
sleeepyjack: Also fixed typo in [63ed50f](https://github.com/NVIDIA/cuCollections/pull/246/commits/63ed50fbe1e4ad1d99fa42de86e087b4f3ebbaf0)

include/cuco/static_set_ref.cuh
```diff
@@ -0,0 +1,127 @@
...
+  using value_type  = typename storage_ref_type::value_type;   ///< Probing scheme element type
+  using size_type   = typename storage_ref_type::size_type;    ///< Probing scheme size type
+  using key_equal   = KeyEqual;  ///< Type of key equality binary callable
+
+  static constexpr int cg_size = probing_scheme_type::cg_size;  ///< Cooperative group size
+  static constexpr int window_size =
+    storage_ref_type::window_size;                    ///< Number of elements handled per window
+  static constexpr cuda::thread_scope scope = Scope;  ///< Thread scope
+
+  // TODO default ctor?
```
sleeepyjack, line 55: Do we want default ctors for the `_ref` classes for convenience, e.g., storing refs in shmem or thrust containers?

include/cuco/detail/static_set/static_set_ref.inl
sleeepyjack: I can't make the atomic load thing work with the current `storage_ref` implementation. We need to refactor it first. I would suggest putting this issue in the backlog for now so we can get this PR merged since it's a blocker for other work on this branch.

include/cuco/static_set_ref.cuh
sleeepyjack: (we can also add this later; I'll leave the TODO for now)

tests/utility/prime.cu
```diff
@@ -0,0 +1,34 @@
...
+#pragma nv_diag_suppress 177  // WAR for weird catch2 warning
+TEST_CASE("Prime number computation", "")
+{
+  REQUIRE(cuco::detail::next_prime(0) == 0);
+  REQUIRE(cuco::detail::next_prime(6) == 7);
+  REQUIRE(cuco::detail::next_prime(13) == 13);
+  REQUIRE(cuco::detail::next_prime(17177758132) == 17177758133);
+
+  // make sure constexpr evaluation is possible
+  static_assert(cuco::detail::next_prime(6) == 7, "Constexpr evaluation fails");
```
sleeepyjack, line 32: While doing the sync with the dev branch, I needed to also refactor the prime computation part. While at it, I decided to enable constexpr prime computation as well.
Those tests are currently failing for no apparent reason. The following godbolt holds the same code and works fine: https://godbolt.org/z/WW9zvEnrj

PointKernel: ... have a good rest :fire:

include/cuco/detail/prime.hpp
```diff
@@ -22,11 +22,12 @@
 #include <algorithm>
 #include <array>
 #include <cstdint>
+#include <limits>
 
 namespace cuco {
 namespace detail {
 
-inline constexpr std::array<int64_t, 140746> primes = {
+inline constexpr std::array<uint64_t, 140739> primes = {
```
sleeepyjack, line 30: oh lol

tests/utility/prime.cu
sleeepyjack: Addressed in [f2cd41b](https://github.com/NVIDIA/cuCollections/pull/246/commits/f2cd41bc781ad1c98b219685b34b053f0ac039a4)

PR #248, opened 2022-12-03
Include missing header

PointKernel:
The PR includes a missing header in `bitwise_compare.cuh`. 

The bug was discovered during @sleeepyjack 's work on mixins for `static_set` (https://github.com/NVIDIA/cuCollections/pull/246/files#r1038822258).


sleeepyjack (approved):
Good point merging this directly into dev.

PR #249, opened 2022-12-05
Remove the use of sentinel namespace

PointKernel:
This is a followup PR of #247. It gets rid of the use of the sentinel namespace in tests, benchmarks, and examples and updates godbolt links.

### DIFF
--- a/examples/static_map/custom_type_example.cu
+++ b/examples/static_map/custom_type_example.cu
@@ -93,9 +93,7 @@ int main(void)
   // Construct a map with 100,000 slots using the given empty key/value sentinels. Note the
   // capacity is chosen knowing we will insert 80,000 keys, for an load factor of 80%.
   cuco::static_map<custom_key_type, custom_value_type> map{
-    100'000,
-    cuco::sentinel::empty_key{empty_key_sentinel},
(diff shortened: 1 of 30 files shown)


examples/static_map/custom_type_example.cu
```diff
@@ -93,9 +93,7 @@ int main(void)
   // Construct a map with 100,000 slots using the given empty key/value sentinels. Note the
   // capacity is chosen knowing we will insert 80,000 keys, for an load factor of 80%.
   cuco::static_map<custom_key_type, custom_value_type> map{
-    100'000,
-    cuco::sentinel::empty_key{empty_key_sentinel},
-    cuco::sentinel::empty_value{empty_value_sentinel}};
+    100'000, cuco::empty_key{empty_key_sentinel}, cuco::empty_value{empty_value_sentinel}};
```
sleeepyjack, line 96: So in our internal code we explicitly specify the tparam for sanity reasons, but in public code we encourage letting the tparam be deduced which makes for better readability. I'm happy with this approach.

sleeepyjack (approved):
Thanks for catching this! LGTM

PR #252, opened 2022-12-08
Declare primes array inline constexpr

sleeepyjack:
Until now, `cuco::detail::primes` was declared as `constexpr` instead of `inline constexpr`.
This prevents global deduplication by the compiler and accounts for ~1MB overhead in binary size per TU.


PointKernel (approved):
LGTM

Thanks!

PR #264, opened 2023-01-25
Customize libcudacxx version

PointKernel:
This PR allows cuco to use a custom version of libcudacxx which doesn't depend on `rapids-cmake`.

### DIFF
--- a/cmake/thirdparty/get_libcudacxx.cmake
+++ b/cmake/thirdparty/get_libcudacxx.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2021, NVIDIA CORPORATION.
+# Copyright (c) 2021-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


cmake/thirdparty/get_libcudacxx.cmake
```diff
@@ -13,16 +13,22 @@
...
+  add_library(libcudacxx INTERFACE)
+  target_include_directories(libcudacxx
+    INTERFACE   $<BUILD_INTERFACE:${libcudacxx_SOURCE_DIR}/include>
+                $<INSTALL_INTERFACE:include/cuco/libcudacxx>)
+  install(DIRECTORY ${libcudacxx_SOURCE_DIR}/include/ DESTINATION include/cuco/libcudacxx)
+  install(DIRECTORY ${libcudacxx_SOURCE_DIR}/libcxx/include/ DESTINATION include/cuco/libcxx/include)
 endfunction()
 
-find_and_configure_libcudacxx()
+find_and_configure_libcudacxx(1.9.1)
```
robertmaynard, line 34: The 1.9.1 tag is broken, you have to use branch-1.9.1.

Which is what rapids-cmake 23.02 is using.

robertmaynard (changes requested):
I think you just want to update to using rapids-cmake 23.02 instead of 22.12

cmake/thirdparty/get_libcudacxx.cmake
robertmaynard: Specifically the atomic headers are broken since they reference types that don't exist, due to a incorrect back porting

robertmaynard: Hmmm. Looks like libcudacxx deleted and recreated the 1.9.1 tag so new clones of the repo now have the corrected atomic version but older clones won't.

PointKernel: > Which is what rapids-cmake 23.02 is using.

Understood. The main goal of this PR is to make libcudacxx version in cuco not depend on the version defined by `rapids-cmake`.

robertmaynard: That will break downstream projects which also bring in libcudacxx since we will either get the wrong version or have ODR violations due to differing implementations of the same functions

PointKernel: Oh no :exploding_head: 

Can we link downstream projects to a cuco-favorable libcudacxx version to solve the conflict?
cc @jrhemstad

PointKernel:
Closing this issue since there is no way to deal with version conflicts or ODR violations for now.

PR #266, opened 2023-01-30
Guard CUDA runtime APIs with error checking

PointKernel:
Similar to https://github.com/rapidsai/cudf/pull/12531

This PR adds error checking for CUDA runtime APIs.

### DIFF
--- a/tests/static_map/insert_and_find_test.cu
+++ b/tests/static_map/insert_and_find_test.cu
@@ -66,7 +66,7 @@ TEMPLATE_TEST_CASE_SIG("Parallel insert-or-update",
   static constexpr int Blocks  = 1024;
   static constexpr int Threads = 128;
   parallel_sum<<<Blocks, Threads>>>(m.get_device_mutable_view());
-  cudaDeviceSynchronize();
+  CUCO_CUDA_TRY(cudaDeviceSynchronize());
 
   thrust::device_vector<Key> d_keys(Iters);
   thrust::device_vector<Value> d_values(Iters);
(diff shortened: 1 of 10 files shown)


tests/static_map/insert_and_find_test.cu
```diff
@@ -66,7 +66,7 @@ TEMPLATE_TEST_CASE_SIG("Parallel insert-or-update",
   static constexpr int Blocks  = 1024;
   static constexpr int Threads = 128;
   parallel_sum<<<Blocks, Threads>>>(m.get_device_mutable_view());
-  cudaDeviceSynchronize();
+  CUCO_CUDA_TRY(cudaDeviceSynchronize());
```
sleeepyjack, line 69: How does catch2 handle exceptions or failing assertions? Will the test case just fail or will this nuke the whole test suite? As a workaround, we could wrap CUDA calls with a catch2 clause, e.g., `REQUIRE(cudaSuccess == cudaDeviceSynchronize())`

sleeepyjack (approved):

tests/static_map/insert_and_find_test.cu
PointKernel: `REQUIRE` would return if a test case fails while `CHECK` won't stop at the first failure and will complete the whole test suite.

PointKernel: We can also wrap `CUCO_CUDA_TRY(...)` with `REQUIRE_NOTHROW(CUCO_CUDA_TRY(...))`

PointKernel: > How does catch2 handle exceptions or failing assertions? Will the test case just fail or will this nuke the whole test suite? 

Just did a quick test, throwing an error will fail the current test case but other tests in the test suite will continue. So the behavior is similar to `CHECK_NOTHROW`. In that sense, wrapping those calls with catch2 macros is not necessary.

PR #270, opened 2023-02-07
Rewrite `static_map` benchmarks with nvbench

PointKernel:
This PR rewrites `static_map` benchmarks with nvbench. 

The goal is to create a baseline to evaluate the performance of the new `static_set` and compare the new and old implementations of `static_map`.


benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,101 @@
...
+
+#include <cuco/static_map.cuh>
+
+#include <nvbench/nvbench.cuh>
+
+#include <thrust/device_vector.h>
+#include <thrust/execution_policy.h>
+#include <thrust/iterator/discard_iterator.h>
+
+namespace cuco::benchmark {
```
sleeepyjack: :O I didn't know this syntax was possible

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * @brief A benchmark evaluating multi-value `contains` performance:
+ * - Total number of insertions: 100'000'000
```
sleeepyjack: I did remove this line in my PR since we need to keep in mind to change it every time we change the default value. Not worth the effort in my book.

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,101 @@
...
+ */
+template <typename Key, typename Value, typename Dist>
+std::enable_if_t<(sizeof(Key) == sizeof(Value)), void> nvbench_static_map_contains(
```
sleeepyjack:
```suggestion
std::enable_if_t<(sizeof(Key) == sizeof(Value)), void> static_map_contains(
```
I removed this prefix. Since we're going to switch to nvbench entirely, there's no need for mentioning it explicitly.

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,101 @@
...
+  nvbench::state& state, nvbench::type_list<Key, Value, Dist>)
+{
+  auto const num_keys      = state.get_int64("NumInputs");
+  auto const occupancy     = state.get_float64("Occupancy");
+  auto const matching_rate = state.get_float64("MatchingRate");
```
sleeepyjack:
```suggestion
  auto const num_keys      = state.get_int64_or_default("NumInputs", defaults::N);
  auto const occupancy     = state.get_float64_or_default("Occupancy", defaults::OCCUPANCY);
  auto const matching_rate = state.get_float64_or_default("MatchingRate", defaults::MATCHING_RATE);
```
Alternatively you can move the `using namespace defaults` to the top so you don't have to be explicit about the namespace here.

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,101 @@
...
+  CUCO_CUDA_TRY(cudaStreamSynchronize(nullptr));
+
+  gen.dropout(thrust::device, keys.begin(), keys.end(), matching_rate);
+
+  state.add_element_count(num_keys, "NumInputs");
+  state.set_global_memory_rw_bytes(num_keys * sizeof(cuco::pair_type<Key, Value>));
+  state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
+    map.contains(keys.begin(),
+                 keys.end(),
+                 thrust::make_discard_iterator(),
```
sleeepyjack, line 63: So this eliminates the need for allocating memory for the result but won't affect timing?

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,101 @@
...
+
+using namespace defaults;
+
+NVBENCH_BENCH_TYPES(nvbench_static_map_contains,
+                    NVBENCH_TYPE_AXES(KEY_TYPE_RANGE,
+                                      VALUE_TYPE_RANGE,
+                                      nvbench::type_list<dist_type::unique>))
+  .set_name("static_map_contains_occupancy")
+  .set_type_axes_names({"Key", "Value", "Distribution"})
+  .set_timeout(100)                  // Custom timeout: 100 s. Default is 15 s.
```
sleeepyjack: Shall we adapt this timeout for all benchmarks?

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,101 @@
...
+  .set_max_noise(MAX_NOISE)          // Custom noise: 3%. By default: 0.5%.
+  .add_int64_axis("NumInputs", {N})  // Total number of key/value pairs: 100'000'000
+  .add_float64_axis("Occupancy", OCCUPANCY_RANGE)
+  .add_float64_axis("MatchingRate", {MATCHING_RATE});
```
sleeepyjack:
```suggestion
  .add_float64_axis("Occupancy", OCCUPANCY_RANGE);
```
You can remove those axis here for better readability

benchmarks/hash_table/static_map/find_bench.cu
```diff
@@ -0,0 +1,101 @@
...
+
+  std::size_t const size = num_keys / occupancy;
+
+  thrust::device_vector<Key> keys(num_keys);
+
+  key_generator gen;
+  gen.generate<Dist>(state, thrust::device, keys.begin(), keys.end());
+
+  auto pairs_begin = thrust::make_transform_iterator(
+    keys.begin(), [] __device__(auto i) { return cuco::pair_type<Key, Value>(i, i); });
```
sleeepyjack: I'm still scratching my head over if using a `transform_iterator` over `transform` is correct here. With the iterator approach, we only load the keys from gmem and construct the pairs ad-hoc in the kernel. This means we actually load `sizeof(key_type)*N` from memory instead of loading `sizeof(pair_type)*N` bytes. My experiments show that there's not much difference in timing. So maybe this discussion is just bikeshedding.

benchmarks/hash_table/static_map/contains_bench.cu
PointKernel: Removed

benchmarks/hash_table/static_map/contains_bench.cu
PointKernel: Introduced in c++17

benchmarks/hash_table/static_map/contains_bench.cu
PointKernel: Cannot really recall why I used it in the first place. This would be only useful when the kernel runtime is long and cannot finish the minimal 10 iterations within 15 seconds. For small kernels, there should be no effect.

benchmarks/hash_table/static_map/find_bench.cu
PointKernel: For query actions like `find`, `count`, this is not an issue since `pairs` are used to insert thus using iterators or not won't affect the benchmark results. For `insert`, how important is this memory load compared to the overall runtime?

sleeepyjack: > For insert, how important is this memory load compared to the overall runtime?

It's not affecting the overall runtime that much (<2%). 

The question is if we're bending the  rules of the benchmark setup with this.

PointKernel: Technically the memory load we are talking about happens [here](https://github.com/PointKernel/cuCollections/blob/ecd39452cd134e42c0d46ce76a1fefea60d254e5/include/cuco/detail/static_map_kernels.cuh#L100). Something like:
```cpp
auto pair = *(pairs_begin + i);
```
Thus it's loading `sizeof(pair_type)` whatsoever, right?

sleeepyjack: If I understand correctly how `thrust::transform_iterator` works, then the actual magic happens inside the `operator*` here. Inside the operator, we load a key from gmem and construct a new pair from it, which we then return. So we're just _loading_ a key instead of a key-value pair from gmem.

.gitignore
```diff
@@ -140,3 +140,6 @@ ENV/
 
 # clang
 compile_commands.json
+
+# figures
+*.eps
```
PointKernel, line 145: this is the figure format generated by our performance analysis notebooks

benchmarks/hash_table/static_map/contains_bench.cu
sleeepyjack: I think I left this in for the multimap benchmarks. We can always remove it in the future.

benchmarks/CMakeLists.txt
```diff
@@ -84,8 +83,11 @@ ConfigureBench(DYNAMIC_MAP_BENCH "${DYNAMIC_MAP_BENCH_SRC}")
 
 ###################################################################################################
 # - static_map benchmarks -------------------------------------------------------------------------
-set(STATIC_MAP_BENCH_SRC "${CMAKE_CURRENT_SOURCE_DIR}/hash_table/static_map_bench.cu")
-ConfigureBench(STATIC_MAP_BENCH "${STATIC_MAP_BENCH_SRC}")
+ConfigureNVBench(STATIC_MAP_BENCH
+  hash_table/static_map/contains_bench.cu
+  hash_table/static_map/erase_bench.cu
+  hash_table/static_map/find_bench.cu
+  hash_table/static_map/insert_bench.cu)
```
sleeepyjack:
```suggestion
  hash_table/static_map/insert_bench.cu
  hash_table/static_map/find_bench.cu
  hash_table/static_map/contains_bench.cu
  hash_table/static_map/erase_bench.cu)
```

Most times, we want to take a look at the `insert` and `find` perf, so we should move those to the top.

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * @brief A benchmark evaluating `contains` performance:
+ */
+template <typename Key, typename Value, typename Dist>
+std::enable_if_t<(sizeof(Key) == sizeof(Value)), void> static_map_contains(
+  nvbench::state& state, nvbench::type_list<Key, Value, Dist>)
+{
+  auto const num_keys      = state.get_int64_or_default("NumInputs", defaults::N);
+  auto const occupancy     = state.get_float64_or_default("Occupancy", defaults::OCCUPANCY);
+  auto const matching_rate = state.get_float64_or_default("MatchingRate", defaults::MATCHING_RATE);
```
sleeepyjack:
```suggestion
  auto const num_keys      = state.get_int64_or_default("NumInputs", N);
  auto const occupancy     = state.get_float64_or_default("Occupancy", OCCUPANCY);
  auto const matching_rate = state.get_float64_or_default("MatchingRate", MATCHING_RATE);
```

Just a suggestion (also how I did this for the multimap benchmarks). I do find your approach more expressive.

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,95 @@
...
+
+  std::size_t const size = num_keys / occupancy;
+
+  thrust::device_vector<Key> keys(num_keys);
+
+  key_generator gen;
+  gen.generate(dist_from_state<Dist>(state), keys.begin(), keys.end());
+
+  auto pairs_begin = thrust::make_transform_iterator(
+    keys.begin(), [] __device__(auto i) { return cuco::pair_type<Key, Value>(i, i); });
```
sleeepyjack:
```suggestion
    keys.begin(), [] __device__(auto i) { return cuco::pair_type<Key, Value>(i, {}); });
```

This way we don't expect `Key` and `Value` to be mutually convertible.

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -0,0 +1,95 @@
...
+  nvbench::state& state, nvbench::type_list<Key, Value, Dist>)
+{
+  state.skip("Key should be the same type as Value.");
+}
+
+NVBENCH_BENCH_TYPES(static_map_contains,
+                    NVBENCH_TYPE_AXES(KEY_TYPE_RANGE,
+                                      VALUE_TYPE_RANGE,
+                                      nvbench::type_list<dist_type::unique>))
+  .set_name("static_map_contains_occupancy")
```
sleeepyjack:
```suggestion
  .set_name("static_map_contains_unique_occupancy")
```

Naming convention: `<map>_<op>_<dist>_<axis>`

sleeepyjack: This is just one occurrence. If we decide to use this naming convention, we need to update all other occurrences too.

benchmarks/CMakeLists.txt
PointKernel: Done. Eventually, this will be formatted by CMake format following alphabetical order.

benchmarks/hash_table/static_map/erase_bench.cu
```diff
@@ -0,0 +1,115 @@
...
+  std::size_t const size = num_keys / occupancy;
+
+  thrust::device_vector<Key> keys(num_keys);
+
+  key_generator gen;
+  gen.generate(dist_from_state<Dist>(state), keys.begin(), keys.end());
+
+  thrust::device_vector<pair_type> pairs(num_keys);
+  thrust::transform(thrust::device, keys.begin(), keys.end(), pairs.begin(), [] __device__(auto i) {
+    return pair_type(i, i);
```
PointKernel:
```suggestion
    return pair_type(i, {});
```

benchmarks/hash_table/static_map/find_bench.cu
```diff
@@ -0,0 +1,94 @@
...
+
+  std::size_t const size = num_keys / occupancy;
+
+  thrust::device_vector<Key> keys(num_keys);
+
+  key_generator gen;
+  gen.generate(dist_from_state<Dist>(state), keys.begin(), keys.end());
+
+  auto pairs_begin = thrust::make_transform_iterator(
+    keys.begin(), [] __device__(auto i) { return cuco::pair_type<Key, Value>(i, i); });
```
PointKernel:
```suggestion
    keys.begin(), [] __device__(auto i) { return cuco::pair_type<Key, Value>(i, {}); });
```

benchmarks/hash_table/static_map/contains_bench.cu
PointKernel: My intention is to use `unique` as the default distribution for all tests except varying "multiplicity". `insert` is the only function we test all three distributions thus I've added distribution in the benchmark name for clarity.

By knowing the default distribution is `unique`, explicitly highlighting it in the name is not necessary IMO. What do you think?

sleeepyjack: I'm ok with that. We need to change the naming in the multimap benchmarks then, because I used my proposed scheme. Not a big issue though - just for consistency.

sleeepyjack (approved):

benchmarks/hash_table/static_map/contains_bench.cu
PointKernel: Got it. Distribution info added for consistency.

I didn't realize you are already following this convention to name multimap benchmarks.

PR #271, opened 2023-02-08
New key generator to be used in benchmarks

sleeepyjack:
Provides a new random key distribution generator based on `thrust` random distributions which is between 8-12x faster than the former host-sided key generator.

nvbench benchmarks have been ported to use the new generator. gbench benchmarks are ported afterward in #270.


benchmarks/hash_table/static_multimap/insert_bench.cu
```diff
@@ -21,37 +21,41 @@
...
 
-  thrust::device_vector<cuco::pair_type<Key, Value>> d_pairs(h_pairs);
+  thrust::device_vector<pair_type> pairs(num_keys);
+  thrust::transform(thrust::device,
+                    keys.begin(),
+                    keys.end(),
+                    pairs.begin(),
+                    [] __host__ __device__(Key const& key) {
+                      return thrust::raw_reference_cast(pair_type(key, 42));
+                    });
```
PointKernel: ~~This changed the original behavior of matching rate control. Should we use `dropout` here?~~
Just realized we don't vary the matching rate in insert bench. Sorry for the confusion.

sleeepyjack: Can you elaborate some more? How did the behavior change, and where do we need `dropout` in ìnsert_bench`?

sleeepyjack:
The `dist_type::gaussian` is subject to change. @gravitino and I are exploring different possibilities for getting an explainable distribution with a single parameter that reflects the heterogenicity of key multiplicities.

benchmarks/hash_table/static_multimap/insert_bench.cu
```diff
@@ -60,41 +64,45 @@ std::enable_if_t<(sizeof(Key) == sizeof(Value)), void> nvbench_static_multimap_i
...
-  nvbench::enum_type_list<dist_type::GAUSSIAN, dist_type::GEOMETRIC, dist_type::UNIFORM>;
 
-using multiplicity = nvbench::enum_type_list<1, 2, 4, 8, 16, 32, 64, 128, 256>;
+NVBENCH_BENCH_TYPES(nvbench_static_multimap_insert,
+                    NVBENCH_TYPE_AXES(key_type, value_type, nvbench::type_list<dist_type::unique>))
+  .set_name("staic_multimap_insert_unique_occupancy")
+  .set_type_axes_names({"Key", "Value", "Distribution"})
+  .set_max_noise(3)                            // Custom noise: 3%. By default: 0.5%.
+  .add_int64_axis("NumInputs", {100'000'000})  // Total number of key/value pairs: 100'000'000
+  .add_float64_axis("Occupancy", nvbench::range(0.1, 0.9, 0.1));
```
PointKernel:
```suggestion
  .set_type_axes_names({"key", "value", "distribution"})
  .set_max_noise(3)                            // Custom noise: 3%. By default: 0.5%.
  .add_int64_axis("num_keys", {100'000'000})  // Total number of key/value pairs: 100'000'000
  .add_float64_axis("occupancy", nvbench::range(0.1, 0.9, 0.1));
```
Let's use snake case consistently for the axis naming

sleeepyjack: Good idea!

benchmarks/defaults.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace cuco {
+namespace benchmark {
+
+namespace defaults {
+using KEY_TYPE_RANGE   = nvbench::type_list<nvbench::int32_t, nvbench::int64_t>;
+using VALUE_TYPE_RANGE = nvbench::type_list<nvbench::int32_t, nvbench::int64_t>;
+
+auto constexpr N             = 100'000'000;
+auto constexpr OCCUPANCY     = 0.8;
```
PointKernel: I'm inclined to use `0.5` as the default occupancy since it's also the default occupancy in libcudf. It happened several times that we improved performance for a certain occupancy but introduced performance regression in libcudf.

benchmarks/hash_table/static_multimap/insert_bench.cu
sleeepyjack: nvbench uses camel-case for the other columns so the table header looks like this:
```
| key | value | distribution | num_keys  | occupancy | multiplicity | Samples | CPU Time  | Noise | GPU Time  | Noise | Elem/s | GlobalMem BW | BWUtil |
```

PointKernel: Ah let's stay with camel-case then.

benchmarks/defaults.hpp
sleeepyjack: addressed in [ccf10d9](https://github.com/NVIDIA/cuCollections/pull/271/commits/ccf10d9da10462cbb51b320c6f77fad15332d04d)

benchmarks/key_generator.hpp
```diff
@@ -18,97 +18,201 @@
...
+   * @param output_begin Start of the output sequence
+   * @param output_end End of the output sequence
+   * @param axis Name of the parameter axis that holds the multiplicity/skew
+   */
+  template <typename Dist, typename ExecPolicy, typename OutputIt>
+  void generate(nvbench::state const& state,
+                ExecPolicy exec_policy,
+                OutputIt out_begin,
+                OutputIt out_end,
+                std::string axis = "")
```
PointKernel: Can we remove `axis`?

benchmarks/hash_table/static_multimap/insert_bench.cu
```diff
@@ -14,44 +14,50 @@
...
 
-  thrust::device_vector<cuco::pair_type<Key, Value>> d_pairs(h_pairs);
+  thrust::device_vector<pair_type> pairs(num_keys);
+  thrust::transform(thrust::device,
+                    keys.begin(),
+                    keys.end(),
+                    pairs.begin(),
+                    [] __host__ __device__(Key const& key) {
+                      return thrust::raw_reference_cast(pair_type(key, 42));
+                    });
```
PointKernel:
```suggestion
  auto pairs_begin = thrust::make_transform_iterator(
      keys.begin(),
      [] __device__(auto i) { return pair_type(i, i); });
```
IIRC, `transform` takes a long time to compile and transform iterator can avoid materializing the data.

sleeepyjack: > transform takes a long time to compile

I was already wondering why this takes longer to compile than before. Let's see if the transform iterator impacts the timing since we move the pair ctor into the critical section.

sleeepyjack: Turns out there's on average <1% diff in timings but compile times are only 3% slower with `transform` compared to using `transform_iterator`

benchmarks/hash_table/static_multimap/count_bench.cu
```diff
@@ -14,102 +14,103 @@
...
+      return pair_type(key, {});
+    });
 
-  thrust::device_vector<Key> d_keys(h_keys);
-  thrust::device_vector<cuco::pair_type<Key, Value>> d_pairs(h_pairs);
+  gen.dropout(keys.begin(), keys.end(), matching_rate);
 
-  state.add_element_count(num_keys, "NumKeys");
+  state.add_element_count(num_keys);
+  state.set_global_memory_rw_bytes(num_keys * sizeof(pair_type));
```
sleeepyjack: Should this be `num_keys * sizeof(Key)` instead, since `count` just takes keys as input but doesn't produce any output besides a single integer? @PointKernel any thoughts on this?

PointKernel: If we want to track the number of bytes "counted" per second. Then yes, it would be `num_keys * sizeof(key_type)`.

benchmarks/key_generator.hpp
```diff
@@ -16,99 +16,209 @@
...
+                          while (val < 0 or val >= num_keys) {
+                            // Re-sample if the value is outside the range [0, N)
+                            // This is necessary because the normal distribution is not bounded
+                            // might be a better way to do this, e.g., discard(n)
+                            val = normal_dist(rng);
+                          }
+                          return val;
+                        });
+    } else {
+      // TODO static assert fail
```
PointKernel:
```suggestion
      CUCO_FAIL("Unexpected distribution type");
```

benchmarks/hash_table/static_multimap/count_bench.cu
```diff
@@ -14,102 +14,102 @@
...
 #include <key_generator.hpp>
 
 #include <cuco/static_multimap.cuh>
 
 #include <nvbench/nvbench.cuh>
 
 #include <thrust/device_vector.h>
+#include <thrust/transform.h>
+
+namespace cuco::benchmark {
```
PointKernel: Would like to get your opinions on this.

Per https://github.com/rapidsai/cudf/issues/11734, `cuco::benchmark::key_generator` can be considered as a `cuco` implementation while benchmark functions are considered as a consumer of `cuco` thus shouldn't be scoped within the `cuco::benchmark` namespace. libcudf is removing `cuco::test` for all test bodies thus we should probably do the same for cuco benchmarks, examples and tests. @sleeepyjack what do you think?

benchmarks/hash_table/static_multimap/insert_bench.cu
```diff
@@ -14,44 +14,48 @@
...
 
 #include <cuco/static_multimap.cuh>
 
 #include <nvbench/nvbench.cuh>
 
 #include <thrust/device_vector.h>
+#include <thrust/transform.h>
+
+namespace cuco::benchmark {
+
```
PointKernel:
```suggestion
using namespace cuco::benchmark;
```
marker for potential namespace removal

benchmarks/key_generator.hpp
```diff
@@ -16,99 +16,209 @@
...
     }
-    case dist_type::UNIFORM: {
-      std::uniform_int_distribution<Key> distribution{1, static_cast<Key>(num_keys / Multiplicity)};
+  }
+
+  /**
+   * @brief Overload of 'generate' which automatically selects a suitable execution policy
+   */
+  template <typename Dist, typename OutputIt>
+  void generate(Dist dist, OutputIt out_begin, OutputIt out_end)
```
PointKernel, line 132: Does this sync internally or do users have to sync on their own to make sure the data generation is complete? The same question for `dropout`.

benchmarks/distribution.hpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION.
```
PointKernel:
```suggestion
 * Copyright (c) 2023, NVIDIA CORPORATION.
```

PointKernel (approved):
LGTM

PR #273, opened 2023-02-21
Benchmark cleanups

sleeepyjack:
This PR fixes some minor issues with the new key generator.
The main goal is to remove the `nvbench` include from the key generator, so it can be used in other projects without parsing any `nvbench` source.


benchmarks/key_generator.hpp
```diff
@@ -169,7 +184,9 @@ class key_generator {
   {
     using value_type = typename std::iterator_traits<InOutIt>::value_type;
 
-    if (keep_prob >= 1.0) {
```
PointKernel, line 172: shame on we two

PointKernel (approved):
LGTM

PR #274, opened 2023-02-21
Benchmark parity

sleeepyjack:
This PR aims to achieve parity/consistency between each benchmark for better readability.

Depends on #273


benchmarks/hash_table/static_multimap/insert_bench.cu
```diff
@@ -29,16 +29,16 @@ using namespace cuco::benchmark;
...
 template <typename Key, typename Value, typename Dist>
 std::enable_if_t<(sizeof(Key) == sizeof(Value)), void> static_multimap_insert(
   nvbench::state& state, nvbench::type_list<Key, Value, Dist>)
 {
   using pair_type = cuco::pair_type<Key, Value>;
 
-  auto const num_keys  = state.get_int64_or_default("NumInputs", N);
-  auto const occupancy = state.get_float64_or_default("Occupancy", OCCUPANCY);
+  auto const num_keys  = state.get_int64_or_default("NumInputs", defaults::N);
+  auto const occupancy = state.get_float64_or_default("Occupancy", defaults::OCCUPANCY);
```
PointKernel, line 40: are `defaults::` still needed with `using namespace cuco::benchmark::defaults;`?

sleeepyjack: Addressed in [a87f398](https://github.com/NVIDIA/cuCollections/pull/274/commits/a87f3982ef86e25c82ccbd896e67cba6b5eb9f2a). tl;dr I removed the `using` statement and added the `defaults::` prefix to all variables, which is a bit more expressive.

benchmarks/hash_table/static_map/contains_bench.cu
```diff
@@ -46,24 +47,22 @@ std::enable_if_t<(sizeof(Key) == sizeof(Value)), void> static_map_contains(
...
-    keys.begin(), [] __device__(auto i) { return cuco::pair_type<Key, Value>(i, {}); });
+  thrust::device_vector<pair_type> pairs(num_keys);
+  thrust::transform(keys.begin(), keys.end(), pairs.begin(), [] __device__(Key const& key) {
+    return pair_type(key, {});
+  });
 
   cuco::static_map<Key, Value> map{size, cuco::empty_key<Key>{-1}, cuco::empty_value<Value>{-1}};
-  map.insert(pairs_begin, pairs_begin + num_keys);
-  CUCO_CUDA_TRY(cudaStreamSynchronize(nullptr));
+  map.insert(pairs.begin(), pairs.end());
```
PointKernel, line 56: So the `insert` internally thus we don't have to manually sync right after?

PointKernel (approved):
LGTM. Thanks!

benchmarks/hash_table/static_map/contains_bench.cu
sleeepyjack: Yep, it will run in the default stream so the kernel syncs wrt the host

PR #275, opened 2023-02-23
Move random key generator to public API

sleeepyjack:
The new key generator is pretty handy when writing standalone PoCs/benchmarks for cuco.
So, I would like to move it to the public API, which can then be used directly from the cuco cmake target (Cc @robertmaynard, who proposed this solution in an offline discussion).

I'm not 100% settled on where exactly to put this file. I'm open to bikeshedding.

This PR sits on top of #274


include/cuco/util/key_generator.hpp
```diff
@@ -34,21 +34,36 @@
 #include <time.h>
 #include <type_traits>
 
-namespace cuco::benchmark {
+namespace cuco::util {
 
 namespace dist_type {
```
PointKernel: Actually, is `distribution` better than `dist_type`?

PointKernel (commented):
I'm inclined to use `utility` instead of the short term`util` to name the folder and namespace.

include/cuco/util/key_generator.hpp
sleeepyjack: Yep, that's a remnant from the prototype. I'll change it to `distribution`

sleeepyjack: addressed in [5bfa854](https://github.com/NVIDIA/cuCollections/pull/275/commits/5bfa854e95eb87c07e1dbde0f77c249830678685)

sleeepyjack:
> I'm inclined to use `utility` instead of the short term`util` to name the folder and namespace.

addressed in [3306a95](https://github.com/NVIDIA/cuCollections/pull/275/commits/3306a95f8159732f5f38ce82d4fce5296ef3fce2)

PointKernel (approved):

PR #276, opened 2023-02-23
Port dynamic_map benchmarks to nvbench and remove gbench dependency

sleeepyjack:
This PR ports dynamic_map benchmarks to nvbench and removes deprecated gbench dependency.


benchmarks/CMakeLists.txt
```diff
@@ -96,3 +61,11 @@ ConfigureNVBench(STATIC_MULTIMAP_BENCH
   hash_table/static_multimap/retrieve_bench.cu
   hash_table/static_multimap/query_bench.cu
   hash_table/static_multimap/count_bench.cu)
+
+###################################################################################################
+# - dynamic_map benchmarks ------------------------------------------------------------------------
+ConfigureNVBench(DYNAMIC_MAP_BENCH
```
PointKernel: https://github.com/sleeepyjack/cuCollections/blob/61d498cbca7b53a5403d1ed9bb0667f58e7949a5/benchmarks/CMakeLists.txt#L31-L35

This reminds me `ConfigureNVBench` could be renamed as `ConfigureBench`. And at line 35 in the link above. `/nvbenchmarks` could be just `/benchmarks`

sleeepyjack: addressed in [53f2c66](https://github.com/NVIDIA/cuCollections/pull/276/commits/53f2c6684f7ad64442c12ec4cf3ef83c33c9bf29)

benchmarks/defaults.hpp
```diff
@@ -32,7 +32,10 @@ auto constexpr MULTIPLICITY  = 8;
 auto constexpr MATCHING_RATE = 0.5;
 auto constexpr MAX_NOISE     = 3;
 auto constexpr SKEW          = 0.5;
+auto constexpr BATCH_SIZE    = 1'000'000;
```
PointKernel, line 35: Could we use a larger batch size to reduce the benchmark time?

benchmarks/hash_table/dynamic_map/erase_bench.cu
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION.
```
PointKernel:
```suggestion
 * Copyright (c) 2023, NVIDIA CORPORATION.
```

PointKernel (commented):
LGTM.

@Nicolas-Iskos, can you please help to review the PR?

sleeepyjack:
rerun tests

PointKernel (approved):
LGTM

benchmarks/defaults.hpp
sleeepyjack: I don't know what reasonable values we should set. Let's check with @Nicolas-Iskos so we can address this in a follow-up PR.

PR #277, opened 2023-02-23
Fix static_map::find benchmark output type

sleeepyjack:
This PR fixes a minor bug by using the proper output type in `static_map::find` benchmark.


PointKernel (approved):
Thanks for the fix. I've updated the PR description since it will go into the final commit info.

PR #280, opened 2023-03-13
Add static map count_by_key_example.cu

sleeepyjack:
This PR adds an example of how to use `cuco::static_map::insert_and_find` to compute a simple histogram over keys.

~~Note that this example requires sm_70+ to compile due to some implementation details of `insert_and_find`.~~ fixed in [3007c25](https://github.com/NVIDIA/cuCollections/pull/280/commits/3007c25ae1c04a94fbe552f37e4a2249fc462f47).


examples/static_map/count_by_key_example.cu
```diff
@@ -0,0 +1,162 @@
...
+    }
+    idx += loop_stride;
+  }
+
+  // compute number of successfully inserted new keys for each block
+  // and atomically add to the grand total
+  uint64_t block_unique_keys = BlockReduce(temp_storage).Sum(thread_unique_keys);
+  if (threadIdx.x == 0) {
+    cuda::atomic_ref<uint64_t, cuda::thread_scope_device> grid_unique_keys(
+      *thrust::raw_pointer_cast(num_unique_keys));
```
PointKernel, line 90: nit: this is a thrust-specific operation while the input is a generic iterator. Would static cast work?

examples/static_map/count_by_key_example.cu
```diff
@@ -0,0 +1,162 @@
...
+ * @param[in] num_keys The total number of keys and values
+ * @param[out] num_unique_keys The total number of distinct keys inserted
+ */
+template <int64_t BlockSize, typename Map, typename KeyIter, typename UniqueIter>
+__global__ void count_by_key(Map map_view,
+                             KeyIter keys,
+                             uint64_t num_keys,
+                             UniqueIter num_unique_keys)
+{
+  typedef cub::BlockReduce<uint64_t, BlockSize> BlockReduce;
```
PointKernel:
```suggestion
  using BlockReduce = cub::BlockReduce<uint64_t, BlockSize> ;
```
nit: prefer `using` over `typedef`

PointKernel (commented):
LGTM. Thanks!

Worth adding a link in https://github.com/NVIDIA/cuCollections#examples

examples/static_map/count_by_key_example.cu
sleeepyjack: I think it's safe to use `thrust::raw_pointer_cast` here as the [docs](https://thrust.github.io/doc/group__memory__management_ga8855ccf5c15c0529206c3cc9cf348af1.html#ga8855ccf5c15c0529206c3cc9cf348af1) state:

> creates a "raw" pointer from a pointer-like type, simply returning the wrapped pointer, should it exist.

So it should work fine on _any_ pointer-like type, not just with Thrust facilities.

examples/static_map/count_by_key_example.cu
sleeepyjack: Addressed in [0a8f84a](https://github.com/NVIDIA/cuCollections/pull/280/commits/0a8f84a5fd5ac670727f55f269fb3ff95ba3198b)

PointKernel (approved):

PR #287, opened 2023-04-03
Remove outdated doc

PointKernel:
The statement of map doesn't support erasing keys is no longer valid and this PR removes the out-of-date documentation.

### DIFF
--- a/include/cuco/dynamic_map.cuh
+++ b/include/cuco/dynamic_map.cuh
@@ -44,7 +44,6 @@ namespace cuco {
  *
  * Current limitations:
  * - Requires keys that are Arithmetic
- * - Does not support erasing keys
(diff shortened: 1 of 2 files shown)


include/cuco/dynamic_map.cuh
```diff
@@ -44,7 +44,6 @@ namespace cuco {
  *
  * Current limitations:
  * - Requires keys that are Arithmetic
```
sleeepyjack: Is this a requirement specific to dynamic map or an error in the docs?

sleeepyjack (approved):
LGTM Thanks! 🚀

include/cuco/dynamic_map.cuh
PointKernel: Updated

PR #299, opened 2023-05-02
Clean up test code

PointKernel:
Some minor cleanups of the test code:

- Use `conditional_t` instead of `if constexpr` to minimize redundancy
- Use `std::distance` instead of raw `-` operator to determine the iterator scope
- `const` when possible


sleeepyjack (approved):
LGTM

PR #301, opened 2023-05-03
Update `static_map` size computation

### LINKED ISSUE #300: [ENHANCEMENT]: static_map insert_and_find() should increment size_ on new insertions.

It would be useful if static_map<>::device_mutable_view::insert_and_find() incremented static_map::size_ on successful insert.

Otherwise if an application is exclusively inserting values via `insert_and_find()` then `get_size()` always returns 0.

How do we know what size to make the output buffer when calling `retrieve_all()`?

Increment `static_map::size_` on successful insert by `insert_and_find()`.

Or, please describe how to get the accurate size of a `static_map` which inserts data via `insert_and_find()`.

_No response_

### Additional context

_No response_

PointKernel:
Closes #300 

Several changes involved in this PR:

- Using a custom kernel to compute the number of filled slots in map and multimap
- Fixing a bug where erased slots are not counted as empty slots
- Allowing size computation for more than 2^31 elements and to be replaced by cub reduce sum once cub-v2.1.0 is available
- Changing the `static_map` public APIs like `insert`, `insert_if` and `erase` to return the number of successful insertions/erasures
- Adding `erased_key_sentinel` into multimap

Note: `get_size` is **expensive**, use the return value of bulk insert/erase when possible.

### DIFF
--- a/include/cuco/detail/static_multimap/static_multimap.inl
+++ b/include/cuco/detail/static_multimap/static_multimap.inl
@@ -909,23 +909,20 @@ template <typename Key,
 std::size_t static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::get_size(
   cudaStream_t stream) const noexcept
 {
-  static_assert(sizeof(pair_atomic_type) == sizeof(value_type));
-  auto view = get_device_view();
+  using namespace experimental::detail;
 
-  auto counter =
-    experimental::detail::counter_storage<std::size_t, Scope, Allocator>{slot_allocator_};
+  auto view    = get_device_view();
+  auto counter = counter_storage<std::size_t, Scope, Allocator>{slot_allocator_};
   counter.reset(stream);
 
   auto const grid_size =
-    (this->get_capacity() +
-     experimental::detail::CUCO_DEFAULT_STRIDE * experimental::detail::CUCO_DEFAULT_BLOCK_SIZE -
-     1) /
-    (experimental::detail::CUCO_DEFAULT_STRIDE * experimental::detail::CUCO_DEFAULT_BLOCK_SIZE);
+    (this->get_capacity() + CUCO_DEFAULT_STRIDE * CUCO_DEFAULT_BLOCK_SIZE - 1) /
+    (CUCO_DEFAULT_STRIDE * CUCO_DEFAULT_BLOCK_SIZE);
 
   // TODO: custom kernel to be replaced by cub::DeviceReduce::Sum when cub version is bumped to
   // v2.1.0
-  detail::size<experimental::detail::CUCO_DEFAULT_BLOCK_SIZE>
-    <<<grid_size, experimental::detail::CUCO_DEFAULT_BLOCK_SIZE, 0, stream>>>(view, counter.data());
+  detail::size<CUCO_DEFAULT_BLOCK_SIZE>
+    <<<grid_size, CUCO_DEFAULT_BLOCK_SIZE, 0, stream>>>(view, counter.data());
 
   return counter.load_to_host(stream);
 }
--- a/include/cuco/detail/static_multimap/static_multimap.inl
+++ b/include/cuco/detail/static_multimap/static_multimap.inl
@@ -909,20 +909,21 @@ template <typename Key,
 std::size_t static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::get_size(
   cudaStream_t stream) const noexcept
 {
-  using namespace experimental::detail;
-
-  auto view    = get_device_view();
-  auto counter = counter_storage<std::size_t, Scope, Allocator>{slot_allocator_};
+  auto view = get_device_view();
+  auto counter =
+    experimental::detail::counter_storage<std::size_t, Scope, Allocator>{slot_allocator_};
   counter.reset(stream);
 
   auto const grid_size =
-    (this->get_capacity() + CUCO_DEFAULT_STRIDE * CUCO_DEFAULT_BLOCK_SIZE - 1) /
-    (CUCO_DEFAULT_STRIDE * CUCO_DEFAULT_BLOCK_SIZE);
+    (this->get_capacity() +
+     experimental::detail::CUCO_DEFAULT_STRIDE * experimental::detail::CUCO_DEFAULT_BLOCK_SIZE -
+     1) /
+    (experimental::detail::CUCO_DEFAULT_STRIDE * experimental::detail::CUCO_DEFAULT_BLOCK_SIZE);
 
   // TODO: custom kernel to be replaced by cub::DeviceReduce::Sum when cub version is bumped to
   // v2.1.0
-  detail::size<CUCO_DEFAULT_BLOCK_SIZE>
-    <<<grid_size, CUCO_DEFAULT_BLOCK_SIZE, 0, stream>>>(view, counter.data());
+  detail::size<experimental::detail::CUCO_DEFAULT_BLOCK_SIZE>
+    <<<grid_size, experimental::detail::CUCO_DEFAULT_BLOCK_SIZE, 0, stream>>>(view, counter.data());
 
   return counter.load_to_host(stream);
 }
--- a/include/cuco/detail/static_multimap/static_multimap.inl
+++ b/include/cuco/detail/static_multimap/static_multimap.inl
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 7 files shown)


mattpulver:
In our particular application, we are looking to only use `insert_and_find()` to insert values (I know bulk insert is more efficient but this is required by our current architecture) without ever erasing any elements.

It would be nice if there was a version of `get_size()` that doesn't bother to check the `erase` sentinel since it is never used in many applications. This would naturally correspond with the version of `static_map`'s constructor that doesn't bother to set the `erased_key_sentinel`.

PointKernel:
@mattpulver Correct me if I am wrong, but the performance regression with an extra erase sentinel check would be negligible.

PointKernel:
I did a test between checking both sentinels and checking empty sentinel only and the performance difference is noticeable only when the map occupancy is really low (about 10%). Dispatching code path at runtime for 0.1% speedup doesn't seem worth the effort.

|  Key  |  Value  |  Distribution  |  Occupancy  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |        Diff |   %Diff |  Status  |
|-------|---------|----------------|-------------|------------|-------------|------------|-------------|-------------|---------|----------|
|  I32  |   I32   |     UNIQUE     |     0.1     |  14.009 ms |       8.23% |  13.460 ms |       0.41% | -549.268 us |  -3.92% |   FAIL   |
|  I32  |   I32   |     UNIQUE     |     0.2     |   6.798 ms |       4.30% |   6.803 ms |       4.46% |    4.697 us |   0.07% |   PASS   |
|  I32  |   I32   |     UNIQUE     |     0.3     |   4.513 ms |       0.37% |   4.509 ms |       0.33% |   -4.458 us |  -0.10% |   PASS   |
|  I32  |   I32   |     UNIQUE     |     0.4     |   3.398 ms |       0.39% |   3.389 ms |       0.37% |   -9.034 us |  -0.27% |   PASS   |
|  I32  |   I32   |     UNIQUE     |     0.5     |   2.727 ms |       0.46% |   2.723 ms |       0.45% |   -3.621 us |  -0.13% |   PASS   |
|  I32  |   I32   |     UNIQUE     |     0.6     |   2.280 ms |       0.52% |   2.274 ms |       0.48% |   -6.090 us |  -0.27% |   PASS   |
|  I32  |   I32   |     UNIQUE     |     0.7     |   1.957 ms |       0.43% |   1.955 ms |       0.50% |   -2.171 us |  -0.11% |   PASS   |
|  I32  |   I32   |     UNIQUE     |     0.8     |   1.716 ms |       0.47% |   1.713 ms |       0.44% |   -2.730 us |  -0.16% |   PASS   |
|  I32  |   I32   |     UNIQUE     |     0.9     |   1.524 ms |       0.43% |   1.526 ms |       0.43% |    1.666 us |   0.11% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.1     |  26.163 ms |       0.01% |  26.165 ms |       0.01% |    2.014 us |   0.01% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.2     |  14.032 ms |      10.56% |  14.074 ms |      10.77% |   42.244 us |   0.30% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.3     |   8.845 ms |       5.22% |   8.860 ms |       5.69% |   15.303 us |   0.17% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.4     |   6.559 ms |       0.05% |   6.559 ms |       0.05% |   -0.097 us |  -0.00% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.5     |   5.252 ms |       0.06% |   5.252 ms |       0.06% |    0.222 us |   0.00% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.6     |   4.383 ms |       0.48% |   4.381 ms |       0.05% |   -2.101 us |  -0.05% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.7     |   3.758 ms |       0.09% |   3.759 ms |       0.08% |    0.885 us |   0.02% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.8     |   3.291 ms |       0.09% |   3.293 ms |       0.58% |    1.528 us |   0.05% |   PASS   |
|  I64  |   I64   |     UNIQUE     |     0.9     |   2.929 ms |       0.11% |   2.928 ms |       0.08% |   -0.657 us |  -0.02% |   PASS   |

include/cuco/detail/static_multimap/static_multimap.inl
```diff
@@ -906,10 +909,22 @@ template <typename Key,
 std::size_t static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::get_size(
   cudaStream_t stream) const noexcept
 {
-  auto begin  = thrust::make_transform_iterator(raw_slots(), detail::slot_to_tuple<Key, Value>{});
-  auto filled = cuco::detail::slot_is_filled<Key>{get_empty_key_sentinel()};
+  using namespace experimental::detail;
```
sleeepyjack: I'm always a bit cautious about pulling the `detail` namespace into the current scope since it can result in naming collisions which are hard to figure out during debugging.

include/cuco/static_map.cuh
```diff
@@ -1350,16 +1359,21 @@ class static_map {
   /**
    * @brief Gets the number of elements in the hash map.
    *
+   * @param stream Stream used for size computation
+   *
    * @return The number of elements in the map
    */
-  std::size_t get_size() const noexcept { return size_; }
+  std::size_t get_size(cudaStream_t stream = 0) const noexcept;
```
sleeepyjack:
```suggestion
  [[nodiscard]] std::size_t get_size(cudaStream_t stream = 0) const noexcept;
```

include/cuco/static_map.cuh
```diff
@@ -1350,16 +1359,21 @@ class static_map {
...
 
   /**
    * @brief Gets the load factor of the hash map.
    *
    * @return The load factor of the hash map
    */
-  float get_load_factor() const noexcept { return static_cast<float>(size_) / capacity_; }
+  float get_load_factor() const noexcept
```
sleeepyjack:
```suggestion
  float get_load_factor(cudaStream_t stream = 0) const noexcept
```
Since this function implicitly calls `get_size(stream)`, it should also take a stream argument and pass it on to `get_size`.

include/cuco/static_multimap.cuh
```diff
@@ -1342,9 +1353,11 @@ class static_multimap {
   }
 
  private:
-  std::size_t capacity_{};                      ///< Total number of slots
-  Key empty_key_sentinel_{};                    ///< Key value that represents an empty slot
-  Value empty_value_sentinel_{};                ///< Initial value of empty slot
+  std::size_t capacity_{};        ///< Total number of slots
+  Key empty_key_sentinel_{};      ///< Key value that represents an empty slot
+  Value empty_value_sentinel_{};  ///< Initial value of empty slot
+  // TODO multimap erase
```
sleeepyjack, line 1359: So for now the sentinel has been introduced to multimap because the size computation uses the common kernel?

sleeepyjack (approved):
This PR adds several long-missing features. Thanks!
LGTM apart from the few suggestions I made.

include/cuco/static_map.cuh
PointKernel: good point

include/cuco/static_multimap.cuh
PointKernel: Right, we would eventually add this feature thus putting a placeholder as a reminder (also for the convenience of common kernel)

include/cuco/static_map.cuh
sleeepyjack: Resolved, since the missing [[nodiscard]] attribute is on-purpose to not break with existing downstream projects.

PointKernel: We can apply it to `get_load_factor` and `get_size` though

include/cuco/static_map.cuh
sleeepyjack: Resolved, since the missing [[nodiscard]] attribute is on-purpose to not break with existing downstream projects.

include/cuco/detail/static_multimap/static_multimap.inl
PointKernel: shortcut removed

PR #307, opened 2023-05-11
Fix narrow conversion issues in probing iterator

PointKernel:
Similar to #306 

This PR fixes narrow conversions inside the probing iterator. Corresponding tests are added to exercise this issue.


sleeepyjack (approved):
This will be implicitly fixed by #304. Let's merge this PR as a WAR until #304 is ready.

PR #310, opened 2023-05-19
New hash function alternatives

### LINKED ISSUE #290: [ENHANCEMENT]: Add XXHash

The current default hasher `murmur3_32` is computational expensive and might lead to suboptimal performance in libcudf's hash join for when the build table is small enough so that it fits in L1 or L2$ (see rapidsai/cudf#10587).

Libcudf has an open feature request (see rapidsai/cudf#12829), which proposes to use a faster hash function, namely [XXHash](https://github.com/Cyan4973/xxHash). I propose adding XXHash as a second hasher to cuco and subsequently compare performance against murmur3_32.

_No response_

### Additional context

_No response_

sleeepyjack:
This PR introduces a set of new hash function alternatives, namely

- `xxhash_32` and `xxhash_64`: 32- and 64-bit versions of the famous [xxHash](https://github.com/Cyan4973/xxHash) hash function.
- `fmix_32` and `fmix_64`: 32- and 64-bit versions of the Murmur3 integer/avalanche finalizer.

`xxhash_32` is now the default hash function for all cuco data structures, due to its better performance in scenarios where the data structure fits into L1$ or L2$. For larger data structures, the performance advantage over Murmurhash3 is negligible.

Benchmark scenario:
Run `cuco::static_set::contains()` multiple times in a row to ensure the set is resident in the cache. Target occupancy is set to 50%. We vary the table size to see the effects of different cache levels. `Ref Time` shows the runtime for `murmurhash3_32` and `Cmp Time` shows the time for `xxhash_32`. All experiments have been conducted on a NVIDIA A100-SXM4-80GB.

|  Key  |  Distribution  |  NumInputs  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |         Diff |   %Diff |
|-------|----------------|-------------|------------|-------------|------------|-------------|--------------|---------|
|  I32  |     UNIQUE     |    8000     | 654.880 us |       0.44% | 658.101 us |       0.40% |     3.220 us |   0.49% | 
|  I32  |     UNIQUE     |    80000    | 850.838 us |       0.36% | 813.983 us |       0.40% |   -36.854 us |  -4.33% |
|  I32  |     UNIQUE     |   800000    |   2.372 ms |       2.98% |   2.319 ms |       0.10% |   -52.546 us |  -2.22% |  
|  I32  |     UNIQUE     |   8000000   |  28.026 ms |       0.16% |  27.976 ms |       0.02% |   -50.574 us |  -0.18% | 
|  I32  |     UNIQUE     |  80000000   | 319.231 ms |       0.03% | 318.746 ms |       0.01% |  -484.905 us |  -0.15% | 
|  I64  |     UNIQUE     |    8000     | 680.446 us |       0.40% | 646.500 us |       0.45% |   -33.946 us |  -4.99% | 
|  I64  |     UNIQUE     |    80000    | 854.186 us |       0.28% | 827.175 us |       0.32% |   -27.010 us |  -3.16% | 
|  I64  |     UNIQUE     |   800000    |   2.628 ms |       0.10% |   2.523 ms |       0.93% |  -104.811 us |  -3.99% | 
|  I64  |     UNIQUE     |   8000000   |  32.144 ms |       0.03% |  31.949 ms |       0.01% |  -195.392 us |  -0.61% | 
|  I64  |     UNIQUE     |  80000000   | 346.017 ms |       0.00% | 344.892 ms |       0.01% | -1124.911 us |  -0.33% | 

In addition, here are some isolated benchmarks calling each hash function in a tight loop on the device:

|                     Hash                     | Samples | CPU Time  | Noise | GPU Time  | Noise |  Elem/s  |
|----------------------------------------------|---------|-----------|-------|-----------|-------|----------|
|            `cuco::detail::MurmurHash3_32<int>` |    332x |  1.520 ms | 0.68% |  1.510 ms | 0.26% | 662.091G |
|           `cuco::detail::MurmurHash3_32<long>` |    174x |  2.891 ms | 0.33% |  2.882 ms | 0.02% | 347.000G |
| `cuco::detail::MurmurHash3_32<key_128_bytes>` |     39x | 12.910 ms | 0.08% | 12.900 ms | 0.01% |  77.521G |
|                 `cuco::detail::XXHash_32<int>` |    405x |  1.245 ms | 0.78% |  1.236 ms | 0.04% | 809.233G |
|                `cuco::detail::XXHash_32<long>` |    206x |  2.443 ms | 0.40% |  2.434 ms | 0.02% | 410.887G |
|      `cuco::detail::XXHash_32<key_128_bytes>` |     48x | 10.444 ms | 0.10% | 10.434 ms | 0.01% |  95.839G |
|                 `cuco::detail::XXHash_64<int>` |    193x |  2.614 ms | 0.37% |  2.604 ms | 0.02% | 384.051G |
|                `cuco::detail::XXHash_64<long>` |    156x |  3.231 ms | 0.30% |  3.221 ms | 0.02% | 310.452G |
|      `cuco::detail::XXHash_64<key_128_bytes>` |     21x | 24.833 ms | 0.04% | 24.822 ms | 0.01% |  40.286G |
|        `cuco::detail::MurmurHash3_fmix32<int>` |    464x |  1.089 ms | 0.88% |  1.080 ms | 0.05% | 926.083G |
|       `cuco::detail::MurmurHash3_fmix64<long>` |    277x |  1.819 ms | 0.54% |  1.809 ms | 0.02% | 552.668G |

Closes #290


benchmarks/defaults.hpp
```diff
@@ -35,7 +35,7 @@ auto constexpr SKEW          = 0.5;
 auto constexpr BATCH_SIZE    = 1'000'000;
 auto constexpr INITIAL_SIZE  = 50'000'000;
 
-auto const N_RANGE             = nvbench::range(10'000'000, 100'000'000, 20'000'000);
+auto const N_RANGE = std::vector<nvbench::int64_t>{8'000, 80'000, 800'000, 8'000'000, 80'000'000};
```
PointKernel: Just to make sure, do we want to apply this to all benchmark cases?

benchmarks/hash_bench.cu
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * @brief A benchmark evaluating performance of various hash functions
+ */
+template <typename Hash>
+void hash_eval(nvbench::state& state, nvbench::type_list<Hash>)
+{
+  bool const materialize_result = false;
+  constexpr auto block_size     = 128;
+  auto const num_keys           = state.get_int64_or_default("NumInputs", defaults::N * 10);
+  auto const grid_size = state.get_int64_or_default("GridSize", SDIV(num_keys, block_size * 16));
```
PointKernel: Probably wroth removing the default `GridSize` if we have one.

include/cuco/detail/hash_functions/murmurhash3.cuh
```diff
@@ -16,8 +16,94 @@
...
+
+  using argument_type = Key;       ///< The type of the values taken as argument
+  using result_type   = uint32_t;  ///< The type of the hash values produced
+
+  /**
+   * @brief Constructs a MurmurHash3_fmix32 hash function with the given `seed`.
+   *
+   * @param seed A custom number to randomize the resulting hash value
+   */
+  __host__ __device__ constexpr MurmurHash3_fmix32(uint32_t seed = 0) : seed_(seed) {}
```
PointKernel:
```suggestion
  __host__ __device__ constexpr MurmurHash3_fmix32(uint32_t seed = 0) : seed_{seed} {}
```
nit: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es23-prefer-the--initializer-syntax

include/cuco/detail/hash_functions/murmurhash3.cuh
```diff
@@ -16,8 +16,94 @@
...
+ *
+ * @throw Key type must be 4 bytes in size
+ *
+ * @tparam Key The type of the values to hash
+ */
+template <typename Key>
+struct MurmurHash3_fmix32 {
+  static_assert(sizeof(Key) == 4, "Key type must be 4 bytes in size.");
+
+  using argument_type = Key;       ///< The type of the values taken as argument
```
PointKernel, line 34: seems never used

include/cuco/detail/hash_functions/murmurhash3.cuh
```diff
@@ -16,8 +16,94 @@
...
+  using result_type   = uint64_t;  ///< The type of the hash values produced
+
+  /**
+   * @brief Constructs a MurmurHash3_fmix64 hash function with the given `seed`.
+   *
+   * @param seed A custom number to randomize the resulting hash value
+   */
+  __host__ __device__ constexpr MurmurHash3_fmix64(uint64_t seed = 0) : seed_(seed) {}
```
PointKernel:
```suggestion
  __host__ __device__ constexpr MurmurHash3_fmix64(uint64_t seed = 0) : seed_{seed} {}
```

include/cuco/detail/hash_functions/murmurhash3.cuh
```diff
@@ -38,15 +124,12 @@ struct MurmurHash3_32 {
...
-  /// Default constructor
-  __host__ __device__ constexpr MurmurHash3_32() : MurmurHash3_32{0} {}
-
   /**
    * @brief Constructs a MurmurHash3_32 hash function with the given `seed`.
    *
    * @param seed A custom number to randomize the resulting hash value
    */
-  __host__ __device__ constexpr MurmurHash3_32(uint32_t seed) : m_seed(seed) {}
+  __host__ __device__ constexpr MurmurHash3_32(uint32_t seed = 0) : fmix32_(0), seed_(seed) {}
```
PointKernel:
```suggestion
  __host__ __device__ constexpr MurmurHash3_32(uint32_t seed = 0) : fmix32_{0}, seed_{seed} {}
```

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+ public:
+  using argument_type = Key;       ///< The type of the values taken as argument
+  using result_type   = uint32_t;  ///< The type of the hash values produced
+
+  /**
+   * @brief Constructs a XXH32 hash function with the given `seed`.
+   *
+   * @param seed A custom number to randomize the resulting hash value
+   */
+  __host__ __device__ constexpr XXHash_32(uint32_t seed = 0) : seed_(seed) {}
```
PointKernel:
```suggestion
  __host__ __device__ constexpr XXHash_32(uint32_t seed = 0) : seed_{seed} {}
```

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+        h64 += (bytes[offset] & 0xFF) * prime5;
+        h64 = rotl(h64, 11) * prime1;
+        ++offset;
+      }
+    }
+    return finalize(h64);
+  }
+
+ private:
+  constexpr __host__ __device__ uint64_t rotl(uint64_t h, int8_t r) const noexcept
```
PointKernel, line 335: worth trying `__force_inline__` since the operation is constantly repeated.

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+ public:
+  using argument_type = Key;       ///< The type of the values taken as argument
+  using result_type   = uint64_t;  ///< The type of the hash values produced
+
+  /**
+   * @brief Constructs a XXH64 hash function with the given `seed`.
+   *
+   * @param seed A custom number to randomize the resulting hash value
+   */
+  __host__ __device__ constexpr XXHash_64(uint64_t seed = 0) : seed_(seed) {}
```
PointKernel:
```suggestion
  __host__ __device__ constexpr XXHash_64(uint64_t seed = 0) : seed_{seed} {}
```

include/cuco/detail/hash_functions/xxhash.cuh
PointKernel: I was thinking of creating a base class for xxhash for common utilities like `rotl` and `finalize` then realized it doesn't provide much convenience.

include/cuco/dynamic_map.cuh
```diff
@@ -208,7 +208,7 @@ class dynamic_map {
    * @param stream Stream used for executing the kernels
    */
   template <typename InputIt,
-            typename Hash     = cuco::murmurhash3_32<key_type>,
+            typename Hash     = cuco::xxhash_32<key_type>,
```
PointKernel: Probably leave this to a separate PR where an extensive performance evaluation is needed to support this change.

include/cuco/hash_functions.cuh
```diff
@@ -16,10 +16,35 @@
 
 #pragma once
 
-#include <cuco/detail/probe_sequence_impl.cuh>
```
PointKernel, line 19: this is embarrassing

include/cuco/hash_functions.cuh
```diff
@@ -16,10 +16,35 @@
...
+/**
+ * @brief The 32-bit integer finalizer function of `MurmurHash3` to hash the given argument on host
+ * and device.
+ *
+ * @throw Key type must be 4 bytes in size
+ *
+ * @tparam Key The type of the values to hash
+ */
+template <typename Key>
+using fmix_32 = detail::MurmurHash3_fmix32<Key>;
```
PointKernel: Is `fmix_32` the proper short term? I'm inclined to keep its full name for clarity `murmurhash3_fmix32`. Another question: what does `f` stand for here? If it's "finalizer", `fmix` (or finalizer mixer) sounds wordy. Educate me if I'm talking nonsense.

include/cuco/hash_functions.cuh
```diff
@@ -16,10 +16,35 @@
 
 #pragma once
 
-#include <cuco/detail/probe_sequence_impl.cuh>
+#include <cuco/detail/hash_functions/murmurhash3.cuh>
+#include <cuco/detail/hash_functions/xxhash.cuh>
+
+#include <cuda/std/type_traits>
```
PointKernel:
```suggestion
```
seems irrelevant

benchmarks/defaults.hpp
sleeepyjack: Now that you mention it, I should probably declare this as a second range `N_RANGE_CACHE` or so.

benchmarks/hash_bench.cu
sleeepyjack: The grid size here is specific to this kernel, where I'm calling the hash function in a loop.

include/cuco/detail/hash_functions/murmurhash3.cuh
sleeepyjack: Right, this is just for completeness, since the other hasher define this type too.

include/cuco/dynamic_map.cuh
sleeepyjack: Good point. I think we also should define an alias `cuco::default_hash_function` and use that instead.

include/cuco/hash_functions.cuh
sleeepyjack: It's either "mix function" or "finalizer/mixer". No clue which one is actually correct. That's just how the function was named in the smhasher implementation and everyone I know of are going with this term.

include/cuco/hash_functions.cuh
sleeepyjack: Good catch! That's a leftover from when I experimented with the `default_hash_function` alias.

include/cuco/dynamic_map.cuh
sleeepyjack: Shall we add the default hasher to the probing scheme, i.e., `cuco::linear_probing<int32_t CGSize, typename Hash1 = cuco::default_hash_function>`?

PointKernel: probing method doesn't know the data structure key type

sleeepyjack: Yeah, I figured that when I was typing the previous comment -.- Maybe as a template template parameter?

include/cuco/hash_functions.cuh
PointKernel: There were two[ test failures ](https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/nvidia%2Fgpuci%2FcuCollections%2Fprb%2FcuCollections-gpu-build/detail/cuCollections-gpu-build/1284/pipeline)with xxhash being the default hasher. Anything we should concern about there?

PointKernel: https://github.com/rurban/smhasher/blob/7d6813f00e0a7be3bde9aae4739b1c765d9f5827/MurmurHash3.cpp#L66

Ah `finalization mix`.

I would go with the full term `murmurhash3_fmix32` and `murmurhash3_fmix64` since they are murmur hash3 specific

sleeepyjack: Nope, that was just me not adjusting the unit tests when switching to the new default hasher. I'll rewind that for now and then the tests should pass again (TM)

benchmarks/defaults.hpp
sleeepyjack: Fixed in [339f907](https://github.com/NVIDIA/cuCollections/pull/310/commits/339f9070e3df77984abd93dfd384918d5ff8d171)

include/cuco/hash_functions.cuh
sleeepyjack: >  would go with the full term murmurhash3_fmix32 and murmurhash3_fmix64 since they are murmur hash3 specific

Maybe `murmurhash3_fmix_32` to follow the same scheme as `xxhash_32` and `murmurhash3_32`? Or is this too snake-case-y?

PointKernel: > Maybe murmurhash3_fmix_32 to follow the same scheme as xxhash_32 and murmurhash3_32?

sure, works for me

sleeepyjack: Addressed in [8b1cf87](https://github.com/NVIDIA/cuCollections/pull/310/commits/8b1cf874ca788d121e8a96fb67f3ea4b5e4aad05)

include/cuco/detail/hash_functions/murmurhash3.cuh
sleeepyjack: That said, I should use `argument_type` instead of `Key` in the implementation where I can. Will do.

include/cuco/detail/hash_functions/xxhash.cuh
sleeepyjack: Just tested this: No noticeable change in the benchmark results with and without `__forceinline__`.

sleeepyjack:
@bdice would you be so kind to leave a review if you have bandwidth?

include/cuco/dynamic_map.cuh
PointKernel: > Maybe as a template template parameter?

Probably not, the data structure key type shouldn't have an effect on how its probing scheme works.

benchmarks/hash_bench.cu
PointKernel: Probably use `get_int64` directly instead of `get_int64_or_default` since there is no default `GridSize` anyway.

include/cuco/static_set.cuh
```diff
@@ -86,8 +86,8 @@ template <class Key,
           cuda::thread_scope Scope = cuda::thread_scope_device,
           class KeyEqual           = thrust::equal_to<Key>,
           class ProbingScheme      = experimental::double_hashing<4,  // CG size
-                                                             cuco::murmurhash3_32<Key>,
-                                                             cuco::murmurhash3_32<Key>>,
+                                                             cuco::xxhash_32<Key>,
+                                                             cuco::xxhash_32<Key>>,
```
PointKernel: A separate PR for this change per the previous discussions.

benchmarks/CMakeLists.txt
```diff
@@ -78,3 +78,6 @@ ConfigureBench(DYNAMIC_MAP_BENCH
   hash_table/dynamic_map/find_bench.cu
   hash_table/dynamic_map/contains_bench.cu
   hash_table/dynamic_map/erase_bench.cu)
+
+ConfigureBench(HASH_BENCH
```
bdice, line 82: Match existing style so this doesn't look like it belongs under the dynamic map benchmarks:
```suggestion
###################################################################################################
# - hash benchmarks -------------------------------------------------------------------------------
ConfigureBench(HASH_BENCH
```

benchmarks/hash_bench.cu
```diff
@@ -0,0 +1,105 @@
...
+#include <cuco/hash_functions.cuh>
+
+#include <nvbench/nvbench.cuh>
+
+#include <thrust/device_vector.h>
+
+#include <cstdint>
+
+using namespace cuco::benchmark;
+using namespace cuco::utility;
```
bdice, line 30: I'd prefer to avoid `using` declarations. If you write the full types, that will help readers who are unfamiliar with cuco internals (like me) know what namespaces contain the identifiers being used.

include/cuco/static_set.cuh
sleeepyjack: I see, I thought this was just targeting the pre-refactor structs. Will do.

include/cuco/detail/hash_functions/murmurhash3.cuh
sleeepyjack: On second thought, I'm not sure if I should use these type aliases in the implementation. Stating the types explicitly makes it more readable. Anyways, these aliases are just for convenience and they are even deprecated in C++17: https://en.cppreference.com/w/cpp/utility/hash

include/cuco/static_set.cuh
sleeepyjack: Addressed in [3d3f730](https://github.com/NVIDIA/cuCollections/pull/310/commits/3d3f73097211bab81267cbc583b96ca2e2c2156b)

benchmarks/hash_bench.cu
sleeepyjack: Addressed on [3d05d30](https://github.com/NVIDIA/cuCollections/pull/310/commits/3d05d30236305055cedaac053721ae9ad21eafac)

include/cuco/detail/hash_functions/murmurhash3.cuh
PointKernel: how about we make this deprecation explicitly?
```suggestion
using argument_type [[deprecated]] = Key; ///< The type of the values taken as argument
```
The same to `result_type` and similar type aliases in other hashers.

sleeepyjack: Based on offline discussion: We keep the type aliases since they are pretty convenient and we make use of them in the benchmark scripts.

PointKernel (approved):
Looks great!

PointKernel (approved):
Looks great!

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+   *
+   * @param key The input argument to hash
+   * @return A resulting hash value for `key`
+   */
+  constexpr result_type __host__ __device__ operator()(Key const& key) const noexcept
+  {
+    // TODO do we need to add checks/hints for alignment?
+    constexpr auto nbytes        = sizeof(Key);
+    char const* const bytes      = (char const*)&key;      ///< per-byte access
+    uint32_t const* const blocks = (uint32_t const*)&key;  ///< 4-byte word access
```
bdice, line 91: Avoid C-style casts.

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+
+    // remaining data can be processed in 4-byte chunks
+    if constexpr ((nbytes % 16) >= 4) {
+      for (; offset <= nbytes - 4; offset += 4) {
+        h32 += blocks[offset / 4] * prime3;
+        h32 = rotl(h32, 17) * prime4;
+      }
+    }
+
+    // the following loop is only needed if the size of the key is no multiple of the block size
```
bdice, line 137:
```suggestion
    // the following loop is only needed if the size of the key is not a multiple of the block size
```

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @tparam Key The type of the values to hash
+ */
+template <typename Key>
+struct XXHash_32 {
+ private:
+  static constexpr uint32_t prime1 = 0x9E3779B1U;
```
bdice, line 63: I have a weak preference for lowercase hex literals / literal suffixes like `0x9e3779b1u` because I find them more readable, but that's totally a matter of taste.

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+   * @param key The input argument to hash
+   * @return A resulting hash value for `key`
+   */
+  constexpr result_type __host__ __device__ operator()(Key const& key) const noexcept
+  {
+    // TODO do we need to add checks/hints for alignment?
+    constexpr auto nbytes         = sizeof(Key);
+    char const* const bytes       = (char const*)&key;      ///< per-byte access
+    uint32_t const* const blocks4 = (uint32_t const*)&key;  ///< 4-byte word access
+    uint64_t const* const blocks8 = (uint64_t const*)&key;  ///< 8-byte word access
```
bdice, line 242: Avoid C-style casts.

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+   * @return A resulting hash value for `key`
+   */
+  constexpr result_type __host__ __device__ operator()(Key const& key) const noexcept
+  {
+    // TODO do we need to add checks/hints for alignment?
+    constexpr auto nbytes        = sizeof(Key);
+    char const* const bytes      = (char const*)&key;      ///< per-byte access
+    uint32_t const* const blocks = (uint32_t const*)&key;  ///< 4-byte word access
+
+    uint32_t offset = 0;
```
bdice, line 93: Probably should be a `std::size_t` to match `nbytes`.

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -0,0 +1,354 @@
...
+   */
+  constexpr result_type __host__ __device__ operator()(Key const& key) const noexcept
+  {
+    // TODO do we need to add checks/hints for alignment?
+    constexpr auto nbytes         = sizeof(Key);
+    char const* const bytes       = (char const*)&key;      ///< per-byte access
+    uint32_t const* const blocks4 = (uint32_t const*)&key;  ///< 4-byte word access
+    uint64_t const* const blocks8 = (uint64_t const*)&key;  ///< 8-byte word access
+
+    uint64_t offset = 0;
```
bdice, line 244: Probably should be a `std::size_t` to match `nbytes`.

tests/utility/hash_test.cu
```diff
@@ -0,0 +1,157 @@
...
+#endif
+
+    // 32*4=128-byte key to test the pipelined outermost hashing loop
+    large_key<32> k10(123456789);             // key
+    uint64_t s10 = 0;                         // seed
+    cuco::xxhash_64<large_key<32>> h10(s10);  // hasher
+    CHECK(h10(k10) == 2031761887105658523ULL);
+  }
+
+  // TODO SECTION("Check if device-generated hash values match the reference implementation.")
```
bdice, line 96: Is this still a TODO?

bdice (commented):
Awesome work here @sleeepyjack! I see this PR was already merged but I wanted to look it over just so I am familiar with the implementation. I have a few comments here that you can apply in a follow-up PR if you would like. 👍

sleeepyjack:
@bdice thank you so much! Reading through your comments, they are all valid and I will address them in a follow-up PR.

tests/utility/hash_test.cu
sleeepyjack: Ha, good catch! Yes, I'll add them in the next PR.

benchmarks/hash_bench.cu
sleeepyjack: Right, we consider benchmarks as user-facing code so we should not make these kinds of shortcuts. Since these namespaces are a bit wordy, how about two-character namespace aliases?

bdice: I don't mind the verbosity at all, honestly. 😄 I'd vote `cuco::utility` over `ut` or `cu` or `util` any day.

bdice:
Also, I didn't leave a comment about this but wanted to share my work: in cudf I chose to use an intrinsic to rotate left:

https://github.com/rapidsai/cudf/blob/37f76c820ddf833a80b4ce706b9c3e84908e51ff/cpp/include/cudf/detail/utilities/hash_functions.cuh#L62-L66

However, it looks like this compiles out to `shf.l.wrap.b32` either way so it doesn't matter which you choose.
https://godbolt.org/z/8b199szzK

There is no 64-bit rotate left intrinsic (afaik) so I would not recommend a change here (so that 32-bit and 64-bit rotation functions look the same).

PR #313, opened 2023-06-01
xxhash cleanups

sleeepyjack:
This PR is a follow-up on #310 and addresses review comments from @bdice.

### DIFF
--- a/include/cuco/detail/hash_functions/xxhash.cuh
+++ b/include/cuco/detail/hash_functions/xxhash.cuh
@@ -86,9 +86,10 @@ struct XXHash_32 {
   constexpr result_type __host__ __device__ operator()(Key const& key) const noexcept
   {
     // TODO do we need to add checks/hints for alignment?
-    constexpr auto nbytes             = sizeof(Key);
-    char const* const bytes           = (char const*)&key;           ///< per-byte access
-    std::uint32_t const* const blocks = (std::uint32_t const*)&key;  ///< 4-byte word access
+    constexpr auto nbytes   = sizeof(Key);
+    char const* const bytes = reinterptret_cast<char const*>(&key);  ///< per-byte access
+    std::uint32_t const* const blocks =
+      reinterpret_cast<std::uint32_t const*>(&key);  ///< 4-byte word access
 
     std::uint32_t offset = 0;
     std::uint32_t h32;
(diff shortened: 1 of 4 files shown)


include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -86,20 +86,20 @@ struct XXHash_32 {
   constexpr result_type __host__ __device__ operator()(Key const& key) const noexcept
   {
     // TODO do we need to add checks/hints for alignment?
```
sleeepyjack, line 88: Can someone comment on this? I'm not sure if the excessive `reinterpret_cast`ing leads to pitfalls.

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -60,22 +60,22 @@ namespace cuco::detail {
...
-  static constexpr uint32_t prime1 = 0x9E3779B1U;
-  static constexpr uint32_t prime2 = 0x85EBCA77U;
-  static constexpr uint32_t prime3 = 0xC2B2AE3DU;
-  static constexpr uint32_t prime4 = 0x27D4EB2FU;
-  static constexpr uint32_t prime5 = 0x165667B1U;
+  static constexpr std::uint32_t prime1 = 0x9E3779B1u;
+  static constexpr std::uint32_t prime2 = 0x85EBCA77u;
+  static constexpr std::uint32_t prime3 = 0xC2B2AE3Du;
+  static constexpr std::uint32_t prime4 = 0x27D4EB2Fu;
+  static constexpr std::uint32_t prime5 = 0x165667B1u;
```
PointKernel: I think @bdice's idea was to use lowercase consistently, e.g. 
```suggestion
  static constexpr std::uint32_t prime1 = 0x9e3779b1u;
  static constexpr std::uint32_t prime2 = 0x85ebca77u;
  static constexpr std::uint32_t prime3 = 0xc2b2ae3du;
  static constexpr std::uint32_t prime4 = 0x27d4eb2fu;
  static constexpr std::uint32_t prime5 = 0x165667b1u;
```
Not a hard requirement though

include/cuco/detail/hash_functions/xxhash.cuh
PointKernel: cc @jrhemstad

include/cuco/detail/hash_functions/xxhash.cuh
```diff
@@ -314,13 +314,13 @@ struct XXHash_64 {
     // remaining data can be processed in 4-byte chunks
     if constexpr (((nbytes % 32) % 8) >= 4) {
       for (; offset <= nbytes - 4; offset += 4) {
-        h64 ^= (blocks4[offset / 4] & 0xFFFFFFFFULL) * prime1;
+        h64 ^= (blocks4[offset / 4] & 0xFFFFFFFFull) * prime1;
```
PointKernel:
```suggestion
        h64 ^= (blocks4[offset / 4] & 0xffffffffull) * prime1;
```

include/cuco/detail/hash_functions/xxhash.cuh
PointKernel: Based on the offline discussion, downcasting to the char pointer shouldn't cause any problems and `blocks` (or the `uint32_t` pointer) won't be used if the key size is smaller than 4 bytes so the current implementation should be safe.

include/cuco/detail/hash_functions/xxhash.cuh
sleeepyjack: fixed in [00e5b54](https://github.com/NVIDIA/cuCollections/pull/313/commits/00e5b54eb31aba90d396552c2552c3492c98fe0f)

include/cuco/detail/hash_functions/xxhash.cuh
sleeepyjack: fixed in [00e5b54](https://github.com/NVIDIA/cuCollections/pull/313/commits/00e5b54eb31aba90d396552c2552c3492c98fe0f)

include/cuco/detail/hash_functions/xxhash.cuh
sleeepyjack: [8c4f167](https://github.com/NVIDIA/cuCollections/pull/313/commits/8c4f1673f09a77ebbb12cc43ea4275410b37b0cb) sprinkles in some `[[maybe_unused]]` attributes for good measure.

PointKernel (approved):
LGTM

PR #314, opened 2023-06-04
Add experimental `static_map`

PointKernel:
Contributes to #110 

This PR adds `experimental::static_map` and involves several changes to the existing code:

- Extracts common `open_addressing_impl` and `open_addressing_ref_impl` classes to minimize duplicates between map and set implementations
- Updates the existing code and fixes bugs: invalid type conversion in `attemp_insert`, narrow conversions inside probing scheme, doc improvement, etc.


include/cuco/static_set.cuh
```diff
@@ -16,8 +16,9 @@
 
 #pragma once
 
+#include <cuco/detail/open_addressing_impl.cuh>
+
```
sleeepyjack:
```suggestion
#include <cuco/detail/open_addressing_impl.cuh>
```

include/cuco/static_set.cuh
```diff
@@ -16,8 +16,9 @@
...
 #include <cuco/cuda_stream_ref.hpp>
-#include <cuco/detail/__config>
 #include <cuco/detail/prime.hpp>
```
sleeepyjack:
```suggestion
```

include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -0,0 +1,421 @@
...
+   * @return An iterator to the position at which the equivalent key is stored
+   */
+  template <typename ProbeKey>
+  [[nodiscard]] __device__ const_iterator find(
+    cooperative_groups::thread_block_tile<cg_size> const& group, ProbeKey const& key) const noexcept
+  {
+    auto const& ref_ = static_cast<ref_type const&>(*this);
+    return ref_.static_map_ref_impl_.find(group, key, ref_.predicate_);
+  }
+};
```
sleeepyjack: These need to be conditionally enabled using `enable_if` based on the `cg_size`. Probably the same for `static_set_ref`.

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,536 @@
...
+ */
+template <class Key,
+          class Value,
+          class Extent,
+          cuda::thread_scope Scope,
+          class KeyEqual,
+          class ProbingScheme,
+          class Allocator,
+          class Storage>
+class open_addressing_impl {
```
sleeepyjack, line 67:
```suggestion
class open_address_container {
```
Just a naming suggestion.

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,536 @@
...
+   * @return The sentinel value used to represent an empty key slot
+   */
+  [[nodiscard]] constexpr key_equal predicate() const noexcept { return predicate_; }
+
+  /**
+   * @brief Gets the sentinel value used to represent an empty key slot.
+   *
+   * @return The sentinel value used to represent an empty key slot
+   */
+  [[nodiscard]] constexpr probing_scheme_type probing_scheme() const noexcept
```
sleeepyjack:
```suggestion
  [[nodiscard]] constexpr probing_scheme_type const& probing_scheme() const noexcept
```

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,536 @@
...
+  {
+    return probing_scheme_;
+  }
+
+  /**
+   * @brief Gets the sentinel value used to represent an empty key slot.
+   *
+   * @return The sentinel value used to represent an empty key slot
+   */
+  [[nodiscard]] constexpr allocator_type allocator() const noexcept { return allocator_; }
```
sleeepyjack:
```suggestion
  [[nodiscard]] constexpr allocator_type const& allocator() const noexcept { return allocator_; }
```

include/cuco/detail/static_map/static_map.inl
```diff
@@ -0,0 +1,344 @@
...
+          class ProbingScheme,
+          class Allocator,
+          class Storage>
+template <typename OutputIt>
+OutputIt
+static_map<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::retrieve_all(
+  OutputIt output_begin, cuda_stream_ref stream) const
+{
+  auto const is_filled = detail::slot_is_filled<key_type>(this->empty_key_sentinel());
+  return static_map_impl_->retrieve_all(output_begin, is_filled, stream);
```
```suggestion
  return impl_->retrieve_all(output_begin, is_filled, stream);
```
I think the shorter version is still self-explanatory

include/cuco/static_map.cuh
```diff
@@ -37,6 +40,443 @@
...
+ * @tparam Allocator Type of allocator used for device storage
+ * @tparam Storage Slot window storage type
+ */
+
+template <class Key,
+          class T,
+          class Extent             = cuco::experimental::extent<std::size_t>,
+          cuda::thread_scope Scope = cuda::thread_scope_device,
+          class KeyEqual           = thrust::equal_to<Key>,
+          class ProbingScheme      = experimental::double_hashing<4,  // CG size
```
sleeepyjack:
```suggestion
          class ProbingScheme      = cuco::experimental::double_hashing<4,  // CG size
```

include/cuco/static_map.cuh
```diff
@@ -37,6 +40,443 @@
...
+                                                             cuco::murmurhash3_32<Key>,
+                                                             cuco::murmurhash3_32<Key>>,
+          class Allocator          = cuco::cuda_allocator<std::byte>,
+          class Storage            = cuco::experimental::aow_storage<1>>
```
sleeepyjack, line 96: Is `window_size=1` the optimal value or is this just a dummy value?

sleeepyjack (commented):
First round of review. Lots of changes so I give it another pass. Thanks for this incredible work!

include/cuco/static_map.cuh
PointKernel: It is the optimal value (for our benchmarks)

include/cuco/detail/open_addressing_impl.cuh
PointKernel: discard this change for now to align with STL behavior

include/cuco/detail/open_addressing_impl.cuh
PointKernel: stay with `open_addressing_impl` for now based on the offline discussion

examples/static_set/device_ref_example.cu
```diff
@@ -51,8 +51,11 @@ __global__ void custom_contains(SetRef set, InputIterator keys, std::size_t n, O
   int64_t const loop_stride = gridDim.x * blockDim.x;
   int64_t idx               = blockDim.x * blockIdx.x + threadIdx.x;
 
+  auto const tile =
+    cooperative_groups::tiled_partition<SetRef::cg_size>(cooperative_groups::this_thread_block());
+
   while (idx < n) {
-    found[idx] = set.contains(*(keys + idx));
+    found[idx] = set.contains(tile, *(keys + idx));
```
sleeepyjack, line 58: 👍

include/cuco/detail/common_functors.cuh
```diff
@@ -0,0 +1,55 @@
...
+template <typename StorageRef>
+struct get_slot {
+  StorageRef storage_;  ///< Storage ref
+
+  /**
+   * @brief Constructs `get_slot` functor with the given storage ref.
+   *
+   * @param s Input storage ref
+   */
+  constexpr get_slot(StorageRef s) noexcept : storage_{s} {}
```
sleeepyjack: `explicit` or not?

include/cuco/detail/common_functors.cuh
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * @brief Accesses the slot content with the given index.
+   *
+   * @param idx The slot index
+   * @return The slot content
+   */
+  __device__ constexpr typename StorageRef::value_type operator()(
```
sleeepyjack: You can probably use `auto` as the return type.

include/cuco/detail/equal_wrapper.cuh
```diff
@@ -38,55 +38,55 @@ enum class equal_result : int32_t { UNEQUAL = 0, EMPTY = 1, EQUAL = 2 };
...
    * @param lhs Left-hand side element to check equality
    * @param rhs Right-hand side element to check equality
-   * @return Three way equality comparison result
+   *
+   * @return `EQUAL` if `lhs` and `rhs` are equivalent. `UNEQUAL` otherwise.
    */
-  template <typename LHS, typename RHS>
-  __device__ constexpr equal_result equal_to(LHS const& lhs, RHS const& rhs) const noexcept
+  template <typename U>
+  __device__ constexpr equal_result equal_to(T const& lhs, U const& rhs) const noexcept
```
sleeepyjack, line 66: Just to make sure I understand this correctly: `T` is the key type and we expect it to be always on the lhs of the operator. `U` can be any type that is _transparent_ with respect to `T`.

include/cuco/detail/equal_wrapper.cuh
```diff
@@ -38,55 +38,55 @@ enum class equal_result : int32_t { UNEQUAL = 0, EMPTY = 1, EQUAL = 2 };
...
 
   /**
    * @brief Order-sensitive equality operator.
    *
-   * This function always compares the left-hand side element against `sentinel_` value first
-   * then perform a equality check with the given `equal_` callable, i.e., `equal_(lhs, rhs)`.
-   *
+   * @note This function always compares the left-hand side element against `empty_sentinel_` value
+   * first then perform a equality check with the given `equal_` callable, i.e., `equal_(lhs, rhs)`.
    * @note Container (like set or map) keys MUST be always on the left-hand side.
```
sleeepyjack, line 76: That's what I meant in my previous comment xD

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,534 @@
...
+template <class Key,
+          class Value,
+          class Extent,
+          cuda::thread_scope Scope,
+          class KeyEqual,
+          class ProbingScheme,
+          class Allocator,
+          class Storage>
+class open_addressing_impl {
+  static_assert(sizeof(Value) <= 8, "Container does not support slot types larger than 8 bytes.");
```
sleeepyjack, line 68: This needs to be relaxed (as discussed in offline discussion regarding b2b CAS)

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,534 @@
...
+                      ProbingScheme>,
+    "ProbingScheme must inherit from cuco::detail::probing_scheme_base");
+
+ public:
+  static constexpr auto cg_size      = ProbingScheme::cg_size;  ///< CG size used for probing
+  static constexpr auto window_size  = Storage::window_size;    ///< Window size used for probing
+  static constexpr auto thread_scope = Scope;                   ///< CUDA thread scope
+
+  using key_type   = Key;    ///< Key type
+  using value_type = Value;  ///< The storage value type, NOT payload type
```
sleeepyjack, line 86: That is a good addition to the comment. I was about to ask that exact question xD

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,534 @@
...
+   * @param empty_key_sentinel The reserved key value for empty slots
+   * @param empty_slot_sentinel The reserved slot value for empty slots
+   * @param pred Key equality binary predicate
+   * @param probing_scheme Probing scheme
+   * @param alloc Allocator used for allocating device storage
+   * @param stream CUDA stream used to initialize the data structure
+   */
+  constexpr open_addressing_impl(Extent capacity,
+                                 key_type empty_key_sentinel,
+                                 value_type empty_slot_sentinel,
```
sleeepyjack, line 119: I do like this concept of having sentinels for the key and the slot, instead of the key and the payload.

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,534 @@
...
+   *
+   * @tparam Predicate Type of predicate indicating if the given slot is filled
+   *
+   * @param is_filled Predicate indicating if the given slot is filled
+   * @param stream CUDA stream used to get the number of inserted elements
+   *
+   * @return The number of elements in the container
+   */
+  template <typename Predicate>
+  [[nodiscard]] size_type size(Predicate const& is_filled, cuda_stream_ref stream) const noexcept
```
sleeepyjack, line 458: Why is the predicate needed here? I suspect it has something to do with this data structure being a swiss army knife, i.e., flexible enough to handle both set and map.

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,534 @@
...
+  {
+    return empty_key_sentinel_;
+  }
+
+  /**
+   * @brief Gets the key comparator.
+   *
+   * @return The comparator used to compare keys
+   */
+  [[nodiscard]] constexpr key_equal key_eq() const noexcept { return predicate_; }
```
sleeepyjack, line 499: Not a fan of the abbreviation but I see why you did it in order to prevent the name collision.

include/cuco/detail/open_addressing_ref_impl.cuh
```diff
@@ -0,0 +1,619 @@
...
+                         value_type const& value,
+                         Predicate const& predicate) noexcept
+  {
+    static_assert(cg_size == 1, "Non-CG operation is incompatible with the current probing scheme");
+    auto probing_iter = probing_scheme_(key, storage_ref_.num_windows());
+
+    while (true) {
+      auto const window_slots = storage_ref_[*probing_iter];
+
+      // TODO: perf gain with #pragma unroll since num_windows is build time constant
```
sleeepyjack: Since the window size is statically known, the compiler should be able to do the unrolling automatically.

include/cuco/detail/open_addressing_ref_impl.cuh
```diff
@@ -0,0 +1,619 @@
...
+   * @param key The key to search for
+   * @param predicate Predicate used to compare slot content against `key`
+   *
+   * @return A boolean indicating whether the probe key is present
+   */
+  template <typename ProbeKey, typename Predicate>
+  [[nodiscard]] __device__ bool contains(ProbeKey const& key,
+                                         Predicate const& predicate) const noexcept
+  {
+    static_assert(cg_size == 1, "Non-CG operation is incompatible with the current probing scheme");
```
sleeepyjack, line 363: 👍 Good WAR

include/cuco/detail/open_addressing_ref_impl.cuh
```diff
@@ -0,0 +1,619 @@
...
+   * @tparam Predicate Predicate type
+   *
+   * @param slot Pointer to the slot in memory
+   * @param value Element to insert
+   * @param predicate Predicate used to compare slot content against `key`
+   *
+   * @return Result of this operation, i.e., success/continue/duplicate
+   */
+  template <typename Predicate>
+  [[nodiscard]] __device__ insert_result attempt_insert(value_type* slot,
```
sleeepyjack, line 558: Should this function also implement the b2b CAS idea or where shall we put it?

include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -0,0 +1,455 @@
...
+   * @param key The key to search for
+   *
+   * @return An iterator to the position at which the equivalent key is stored
+   */
+  template <typename ProbeKey>
+  [[nodiscard]] __device__ const_iterator find(ProbeKey const& key) const noexcept
+  {
+    // CRTP: cast `this` to the actual ref type
+    auto const& ref_ = static_cast<ref_type const&>(*this);
+    return ref_.impl_.find(key, ref_.predicate_);
```
sleeepyjack, line 428: I don't quite understand how this "dispatch to impl" design works with the mixins stuff. We want to dispatch specialized code paths based on the given set of mixins.

```cpp
auto find(Key) {
  //...
  if constexpr (has_operator_v<Mixins..., cuco::insert>) {
    // atomic load
  } else {
    // relaxed/window load
  }
  //...
}
```

From my understanding, the impl type doesn't know about the given mixin list.

include/cuco/static_map.cuh
```diff
@@ -1424,3 +1863,4 @@ class static_map {
 }  // namespace cuco
 
 #include <cuco/detail/static_map.inl>
```
sleeepyjack, line 1868: Is GH UI weird or are we including both `<cuco/detail/static_map.inl>` and `<cuco/detail/static_map/static_map.inl>`?

include/cuco/detail/equal_wrapper.cuh
PointKernel: Correct

include/cuco/detail/open_addressing_impl.cuh
PointKernel: > flexible enough to handle both set and map.

Exactly

include/cuco/detail/open_addressing_impl.cuh
PointKernel: Me neither. This is only to align with STL convention: https://en.cppreference.com/w/cpp/container/unordered_map/key_eq

include/cuco/detail/open_addressing_ref_impl.cuh
PointKernel: would be in a followup PR

include/cuco/static_map.cuh
PointKernel: `<cuco/detail/static_map.inl>` is the legacy code and `<cuco/detail/static_map/static_map.inl>` is where the new experimental map located. `<cuco/detail/static_map.inl>` to be removed once we get rid of the legacy map.

include/cuco/detail/static_map/static_map_ref.inl
PointKernel: Right, that's something I didn't think about during this refactoring.

PointKernel: We could pass a non-type template parameter to impl based on the given mixins

include/cuco/detail/open_addressing_impl.cuh
```diff
@@ -0,0 +1,534 @@
...
+template <class Key,
+          class Value,
+          class Extent,
+          cuda::thread_scope Scope,
+          class KeyEqual,
+          class ProbingScheme,
+          class Allocator,
+          class Storage>
+class open_addressing_impl {
+  static_assert(sizeof(Value) <= 16, "Container does not support slot types larger than 16 bytes.");
```
sleeepyjack: Shall we restrict this to 8 bytes until we implement b2b CAS?

sleeepyjack: Also, should we move this assert to the public API of map/set so that this restriction is visible to the user? Edit: The assert is only missing in the public `static_map` header

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: Let's move this to another PR since the change might be significant

PR #315, opened 2023-06-05
Enable fast_int functionality for dynamic extent

sleeepyjack:
Adds `fast_int` functionality to `cuco::extent`.

Related to #284


sleeepyjack:
Hmm, I'm getting a segfault in the unit tests:
```
-------------------------------------------------------------------------------
Unique sequence - int32_t, cuco::test::probe_sequence::double_hashing, 1
  Non-inserted keys should not be contained.
-------------------------------------------------------------------------------
/workspaces/cuCollections/tests/static_set/unique_sequence_test.cu:49
...............................................................................

/workspaces/cuCollections/tests/static_set/unique_sequence_test.cu:49: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
```
And compute-sanitizer points me to an out-of-bounds access [here](https://github.com/NVIDIA/cuCollections/blob/87ad966648685a8856c913a8effea5b12faf9dd7/include/cuco/detail/static_set/static_set_ref.inl#L421). Not sure why that's the case.

Cc @PointKernel

PointKernel:
> And compute-sanitizer points me to an out-of-bounds access [here](https://github.com/NVIDIA/cuCollections/blob/87ad966648685a8856c913a8effea5b12faf9dd7/include/cuco/detail/static_set/static_set_ref.inl#L421). Not sure why that's the case.

Printing out the value of `*probing_iter` would give more information.

sleeepyjack:
Found the root of the cause and fixed it (see #315 and [5afcc8c](https://github.com/NVIDIA/cuCollections/pull/315/commits/5afcc8c4f15cd72f1388273baffe0e099393976d)).

include/cuco/detail/probing_scheme_impl.inl
```diff
@@ -97,9 +99,10 @@ __host__ __device__ constexpr auto linear_probing<CGSize, Hash>::operator()(
   ProbeKey const& probe_key, Extent upper_bound) const noexcept
 {
   using size_type = typename Extent::value_type;
-  return detail::probing_iterator<Extent>{static_cast<size_type>(hash_(probe_key) % upper_bound),
-                                          1,  // step size is 1
-                                          upper_bound};
+  return detail::probing_iterator<Extent>{
+    static_cast<size_type>(cuco::detail::sanitize_hash<size_type>(hash_(probe_key)) % upper_bound),
```
PointKernel:
```suggestion
    cuco::detail::sanitize_hash<size_type>(hash_(probe_key)) % upper_bound,
```
`static_cast` should be no longer needed since `sanitize_hash` does the casting already. Similar to other castings in the file.

include/cuco/detail/storage/storage_base.cuh
```diff
@@ -71,20 +71,30 @@ class storage_base {
...
+   * @return The extent.
+   */
+  [[nodiscard]] __host__ __device__ constexpr extent_type extent() const noexcept
+  {
+    return extent_;
   }
 
  protected:
-  extent_type capacity_;  ///< Total number of elements
+  extent_type extent_;  ///< Total number of elements
```
PointKernel, line 97: I was using `capacity` mainly to align with STL, where `size` indicates the number of filled elements in the container and `capacity` indicates the maximum number of elements a container can hold.

include/cuco/static_set_ref.cuh
```diff
@@ -71,6 +71,8 @@ class static_set_ref
                       ProbingScheme>,
     "ProbingScheme must inherit from cuco::detail::probing_scheme_base");
 
+  // TODO static_assert that StorageRef::extent_type is valid_extent
```
PointKernel: Needs to be done in this PR

PointKernel (changes requested):
Similar to `static_set_ref`, `aow_storage` or `aow_storage_ref` should be exclusively constructed from a `valid_extent` as well.

include/cuco/detail/probing_scheme_impl.inl
sleeepyjack: Addressed in [3d93880](https://github.com/NVIDIA/cuCollections/pull/315/commits/3d938805464a7b3fb930c24019e960afc36eb173)

include/cuco/detail/storage/storage_base.cuh
sleeepyjack: This has already been resolved by a commit that's on the dev branch

include/cuco/static_set_ref.cuh
sleeepyjack: Addressed in [9f01917](https://github.com/NVIDIA/cuCollections/pull/315/commits/9f0191755b686b3a6a5015a990aacc3ba2a2b006)

sleeepyjack:
> Similar to static_set_ref, aow_storage or aow_storage_ref should be exclusively constructed from a valid_extent as well.

@PointKernel I added the check to the `open_addressing_ref_impl`. I don't think we should make the storage type only constructible from `valid_extent`, since the imposed constrains introduced by this type are only needed if we use a probing scheme over the storage. The storage class is more generic, i.e., we could use it for other things that don't use probing.

sleeepyjack:
CI build is failing because we need to make some changes to `tests/static_set/capacity_test.cu`.

include/cuco/detail/storage/storage_base.cuh
PointKernel: So the convention is to have `capacity()`/`num_windows()` returning `size_type` and `extent()`/`window_extent` returning extent type.

PointKernel (approved):

include/cuco/extent.cuh
```diff
@@ -120,30 +105,9 @@ struct extent<SizeType, dynamic_extent> {
...
-  }
-  if constexpr (N != dynamic_extent) {
-    return extent<SizeType,
-                  static_cast<std::size_t>(*cuco::detail::lower_bound(cuco::detail::primes.begin(),
-                                                                      cuco::detail::primes.end(),
-                                                                      static_cast<uint64_t>(size)) *
-                                           CGSize)>{};
-  }
-}
+[[nodiscard]] auto constexpr make_valid_extent(extent<SizeType, N> ext);
```
PointKernel: Just to double check, is this function invokable on the device side? If so, we need to add `__host__ __device__` qualifiers. Otherwise, we need to find a way to make it work since this feature is required by several places in libcudf.

sleeepyjack: That's an excellent topic! The problem I'm currenrly facing wirh making this function callable from the device, too, is that it needs access to the prime array which is declared `inline constexpr`, which means it's only accessible from the host. We could declare it e.g. `__constant__` but it would be a runtime constant then.

include/cuco/detail/open_addressing_ref_impl.cuh
```diff
@@ -59,6 +60,13 @@ class open_addressing_ref_impl {
                       ProbingScheme>,
     "ProbingScheme must inherit from cuco::detail::probing_scheme_base");
 
+  static_assert(is_window_extent_v<typename StorageRef::extent_type>,
+                "Extent is not a valid cuco::window_extent");
+  static_assert(ProbingScheme::cg_size == StorageRef::extent_type::cg_size,
+                "Extent has incompatible CG size");
+  static_assert(StorageRef::window_size == StorageRef::extent_type::window_size,
+                "Extent has incompatible window size");
```
sleeepyjack, line 68: This ensures the extent is valid

PointKernel:
Merge as it is for now to unblock the following work.

Certain details need to be revisited:

- [ ] `CGSize` in `window_extent`
- [ ] `is_window_extent` assert in ref base class
- [ ] benchmark results

PR #318, opened 2023-06-14
Make XXHash the default hash function

sleeepyjack:
This PR introduces `xxhash_32` as the new default hash function for cuco's data structures.

Additional changes:

- Add `cuco::default_hash_function` alias for convenience.
- Default Double Hashing's second hash function: `typename Hash2 = Hash1` for convenience.

### DIFF
--- a/include/cuco/probe_sequences.cuh
+++ b/include/cuco/probe_sequences.cuh
@@ -60,7 +60,7 @@ class linear_probing : public detail::probe_sequence_base<CGSize> {
  * @tparam Hash1 Unary callable type
  * @tparam Hash2 Unary callable type
  */
-template <uint32_t CGSize, typename Hash1, typename Hash2>
+template <uint32_t CGSize, typename Hash1, typename Hash2 = Hash1>
 class double_hashing : public detail::probe_sequence_base<CGSize> {
  public:
   using probe_sequence_base_type =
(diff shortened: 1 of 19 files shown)


include/cuco/probe_sequences.cuh
```diff
@@ -60,7 +60,7 @@ class linear_probing : public detail::probe_sequence_base<CGSize> {
  * @tparam Hash1 Unary callable type
  * @tparam Hash2 Unary callable type
  */
-template <uint32_t CGSize, typename Hash1, typename Hash2>
+template <uint32_t CGSize, typename Hash1, typename Hash2 = Hash1>
```
PointKernel, line 63: I hesitated at first glance then find it's actually not a bad idea.

PointKernel (changes requested):
@sleeepyjack can you please hold until #314 is merged?

Also, it would be great to include benchmark comparisons for at least `static_set` to support this change.

sleeepyjack:
It was a bit difficult to get robust benchmark results (see NVIDIA/nvbench#135)
but here are the benchmark results for `cuco::experimental::static_map<int, int>::contains_async`, measuring speedup of xxHash over MurmurHash3 using the minimum runtime as the underlying metric.
| Table size | 6.4MB  | 64MB  | 640MB | 6400MB |
|------------|--------|-------|-------|--------|
| Speedup    | 18.75% | 8.57% | 0.99% | 0.08%  |

PointKernel (approved):
LGTM!

@sleeepyjack The PR is ready to go once the merge conflict is resolved.

PR #319, opened 2023-06-20
Clean up `cuco::pair`

PointKernel:
Contributes to #110 
Depends on #314 

This PR:
- deprecates `cuco::pair_type` alias
- fixes issues with `cuco::make_pair`
- separates `pair` declarations and implementation details


include/cuco/detail/traits.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ */
+
+#pragma once
+
+#include <thrust/device_reference.h>
+#include <thrust/tuple.h>
+
+#include <tuple>
+#include <type_traits>
```
sleeepyjack: Since this is a new file, we should switch to the `cuda/std` namespace when possible to get rid of the `--relaxed-constexpr` flag at some point.

include/cuco/detail/utils.cuh
```diff
@@ -102,5 +102,112 @@ struct strong_type {
...
+constexpr std::size_t next_pow2(std::size_t v) noexcept
+{
+  --v;
+  v |= v >> 1;
+  v |= v >> 2;
+  v |= v >> 4;
+  v |= v >> 8;
+  v |= v >> 16;
+  return ++v;
+}
```
sleeepyjack: This can be replaced with `cuda::std::bit_ceil` from the `<cuda/std/bit>` header.

include/cuco/detail/traits.hpp
PointKernel: good call. moved to libcuda++ when possible.

sleeepyjack (approved):

PR #325, opened 2023-06-28
Modify `experimental::static_map::retrieve_all` API

PointKernel:
This PR modifies the public `experimental::static_map::retrieve_all` API to take a key iterator and a value iterator separately.

### DIFF
--- a/include/cuco/detail/static_map/functors.cuh
+++ b/include/cuco/detail/static_map/functors.cuh
@@ -17,16 +17,51 @@
 
 #include <cuco/detail/bitwise_compare.cuh>
 
+#include <thrust/tuple.h>
+
 namespace cuco {
 namespace experimental {
+namespace static_map_ns {
 namespace detail {
+
+/**
+ * @brief Device functor returning the content of the slot indexed by `idx`.
+ *
+ * @tparam StorageRef Storage ref type
+ */
+template <typename StorageRef>
+struct get_slot {
+  StorageRef storage_;  ///< Storage ref
+
+  /**
+   * @brief Constructs `get_slot` functor with the given storage ref.
+   *
+   * @param s Input storage ref
+   */
+  explicit constexpr get_slot(StorageRef s) noexcept : storage_{s} {}
+
+  /**
+   * @brief Accesses the slot content with the given index.
+   *
+   * @param idx The slot index
+   * @return The slot content
+   */
+  __device__ constexpr auto operator()(typename StorageRef::size_type idx) const noexcept
+  {
+    auto const window_idx      = idx / StorageRef::window_size;
+    auto const intra_idx       = idx % StorageRef::window_size;
+    auto const [first, second] = storage_[window_idx][intra_idx];
+    return thrust::make_tuple(first, second);
+  }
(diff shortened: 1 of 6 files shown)


include/cuco/detail/static_map/functors.cuh
```diff
@@ -17,17 +17,51 @@
 
 #include <cuco/detail/bitwise_compare.cuh>
 
+#include <thrust/tuple.h>
```
sleeepyjack, line 20: Can we use `cuda::std::tuple` instead?

include/cuco/detail/static_map/static_map.inl
```diff
@@ -237,13 +237,21 @@ template <class Key,
           class ProbingScheme,
           class Allocator,
           class Storage>
-template <typename OutputIt>
-OutputIt
+template <typename KeyOut, typename ValueOut>
+std::pair<KeyOut, ValueOut>
 static_map<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::retrieve_all(
-  OutputIt output_begin, cuda_stream_ref stream) const
+  KeyOut keys_out, ValueOut values_out, cuda_stream_ref stream) const
```
sleeepyjack, line 243: Why do we need to use separate key and value iterators instead of using a key-value pair iterator?

include/cuco/detail/static_map/static_map.inl
```diff
@@ -237,13 +237,21 @@ template <class Key,
...
 {
-  auto const is_filled = static_map_ns::detail::slot_is_filled(this->empty_key_sentinel());
-  return impl_->retrieve_all(output_begin, is_filled, stream);
+  // XXX: thrust counting iterator doesn't work if size_type is larger than 32-bit
```
sleeepyjack: WTF, really??

include/cuco/detail/static_map/static_map.inl
```diff
@@ -237,13 +237,21 @@ template <class Key,
...
+  auto const begin = thrust::make_transform_iterator(
+    thrust::counting_iterator<size_type>(0),
```
```suggestion
    thrust::counting_iterator<size_type>{0},
```

include/cuco/detail/static_map/static_map.inl
PointKernel: Seems like a false alarm. I'm removing the comment for now and see if anyone complains.

include/cuco/detail/static_map/static_map.inl
PointKernel: This would not introduce the construction of `cuco::pair` to users and relax the requirement of the iterator `value_type` must be constructible from the container's `value_type`.

include/cuco/detail/static_map/functors.cuh
PointKernel: It doesn't work since the map `retrieve_all` passes a zip iterator to the `open_addressing_impl` and `thrust::tuple` (from zip iterator) is not assignable from a `cuda::std::tuple` -_-||

sleeepyjack (approved):

PR #326, opened 2023-06-29
Fix xxhash64 implementation

sleeepyjack:
There is a typo in the current `xxhash_64` implementation which results in incorrect hash values if the key type is not a multiple of 4 bytes.

This PR fixes this typo and adds tests for <4byte key types.

### DIFF
--- a/tests/utility/hash_test.cu
+++ b/tests/utility/hash_test.cu
@@ -48,21 +48,28 @@ __host__ __device__ bool check_hash_result(typename Hash::argument_type const& k
 template <typename OutputIter>
 __global__ void check_hash_result_kernel_64(OutputIter result)
 {
-  result[0] = check_hash_result<cuco::xxhash_64<int32_t>>(0, 0, 4246796580750024372);
-  result[1] = check_hash_result<cuco::xxhash_64<int32_t>>(0, 42, 3614696996920510707);
-  result[2] = check_hash_result<cuco::xxhash_64<int32_t>>(42, 0, 15516826743637085169);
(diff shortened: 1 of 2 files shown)


sleeepyjack:
cc @bdice

tests/utility/hash_test.cu
```diff
@@ -101,28 +112,38 @@ TEST_CASE("Test cuco::xxhash_64", "")
...
+  result[i++] = check_hash_result<cuco::xxhash_32<int64_t>>(42, 0, 2332451213);
+  result[i++] = check_hash_result<cuco::xxhash_32<int64_t>>(123456789, 0, 1561711919);
 
 #if defined(CUCO_HAS_INT128)
-  result[8] = check_hash_result<cuco::xxhash_32<__int128>>(123456789, 0, 1846633701);
+  result[i++] = check_hash_result<cuco::xxhash_32<__int128>>(123456789, 0, 1846633701);
 #endif
 
-  result[9] = check_hash_result<cuco::xxhash_32<large_key<32>>>(123456789, 0, 3715432378);
+  result[i++] = check_hash_result<cuco::xxhash_32<large_key<32>>>(123456789, 0, 3715432378);
```
PointKernel, line 135: @sleeepyjack can you please remind me why we are not using a global array to avoid duplications?

sleeepyjack: Agreed, this is a bad solution. What does your proposal look like?

PointKernel: https://godbolt.org/z/9T9e764K7 would this work?

sleeepyjack: Result from offline discussion: move to backlog for now

PointKernel (approved):

PR #334, opened 2023-07-17
Remove extra mod

bdice:
This PR removes an extra modulus operation. 8 is a factor of 32, so `(x % 32) % 8 == x % 8`.

Ref: https://github.com/rapidsai/cudf/pull/13612/files#r1264529130

cc: @sleeepyjack


sleeepyjack (approved):
Ha, nice catch! That was a leftover from when `size` was a static constant. Here's a small repro that the compiler actually emits fewer instructions after removing the extra `%`: https://godbolt.org/z/vcGsqaPd1

PR #337, opened 2023-07-26
Add block synchronisation after reset of CG-level counters

### LINKED ISSUE #336: [BUG]: data race in static_multimap pair_retrieve
### Is this a duplicate?

### Type of Bug

Something else

### Describe the bug

Hash-based joins in cudf that use static_multimap's pair_retrieve are not racecheck clean: eg.

```
========= Warning: Race reported between Write access at 0x6db0 in /home/wence/Documents/src/rapids/cudf/cpp/build/cuda-11.8.0/branch-23.08/release/_deps/cuco-src/include/cuco/detail/static_multimap/kernels.cuh:500:void cuco::detail::pair_retrieve<(unsigned int)128, (unsigned int)32, (unsigned int)2, (unsigned int)96, (bool)1, thrust::transform_iterator<cudf::detail::make_pair_function<cudf::experimental::row::hash::device_row_hasher<cudf::hashing::detail::default_hash, cudf::nullate::DYNAMIC>, int>, thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, thrust::use_default, thrust::use_default>, thrust::zip_iterator<thrust::tuple<thrust::discard_iterator<thrust::use_default>, int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, thrust::zip_iterator<thrust::tuple<thrust::discard_iterator<thrust::use_default>, int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, cuda::__4::atomic<unsigned long, (cuda::std::__4::__detail::thread_scope)1>, cuco::static_multimap<unsigned int, int, (cuda::std::__4::__detail::thread_scope)1, rmm::mr::stream_allocator_adaptor<default_allocator<char>>, cuco::double_hashing<(unsigned int)2, cudf::hashing::detail::MurmurHash3_x86_32<unsigned int>, cudf::hashing::detail::MurmurHash3_x86_32<unsigned int>>>::device_view, cudf::detail::pair_equality<cudf::experimental::row::equality::strong_index_comparator_adapter<cudf::experimental::row::equality::device_row_comparator<(bool)0, cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>>>>(T6, long, T7, T8, T9 *, T10, T11)
=========     and Read access at 0x25410 in /home/wence/Documents/src/rapids/cudf/cpp/build/cuda-11.8.0/branch-23.08/release/_deps/cuco-src/include/cuco/detail/static_multimap/kernels.cuh:536:void cuco::detail::pair_retrieve<(unsigned int)128, (unsigned int)32, (unsigned int)2, (unsigned int)96, (bool)1, thrust::transform_iterator<cudf::detail::make_pair_function<cudf::experimental::row::hash::device_row_hasher<cudf::hashing::detail::default_hash, cudf::nullate::DYNAMIC>, int>, thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, thrust::use_default, thrust::use_default>, thrust::zip_iterator<thrust::tuple<thrust::discard_iterator<thrust::use_default>, int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, thrust::zip_iterator<thrust::tuple<thrust::discard_iterator<thrust::use_default>, int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, cuda::__4::atomic<unsigned long, (cuda::std::__4::__detail::thread_scope)1>, cuco::static_multimap<unsigned int, int, (cuda::std::__4::__detail::thread_scope)1, rmm::mr::stream_allocator_adaptor<default_allocator<char>>, cuco::double_hashing<(unsigned int)2, cudf::hashing::detail::MurmurHash3_x86_32<unsigned int>, cudf::hashing::detail::MurmurHash3_x86_32<unsigned int>>>::device_view, cudf::detail::pair_equality<cudf::experimental::row::equality::strong_index_comparator_adapter<cudf::experimental::row::equality::device_row_comparator<(bool)0, cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>>>>(T6, long, T7, T8, T9 *, T10, T11) [4 hazards]
========= 
```

### How to Reproduce

```python
import cudf
df = cudf.DataFrame({"a": [1, 2, 3], "b": [1, 2, 3]})
inter = df.merge(df, on=["a", "b"], how="left")
```

```
compute-sanitizer --tool=racecheck python bug.py
```

I can try and build a cuco only reproducer but I am not very familiar with the API...

### Expected behavior

No races.
Here's the bit of code that's being complained about:

https://github.com/NVIDIA/cuCollections/blob/a2833dbfb1b7d4915d530bd6adb45092a87a3b07/include/cuco/detail/static_multimap/kernels.cuh#L498-L538

With a race conflict between lines 500 and 536, `view.pair_retrieve` syncs the CG [edit: only the smaller `active_cg` so I don't think that is enough anyway], but then writes to the `__shared__` `flushing_cg_counter` without syncing afterwards. So we can exit the loop on line 534 potentially with an unsynced write. Equally, if the loop never executes, we will also not sync the CG. So plausibly 
```diff
diff --git a/include/cuco/detail/static_multimap/kernels.cuh b/include/cuco/detail/static_multimap/kernels.cuh
index ca5f898..1ffa170 100644
--- a/include/cuco/detail/static_multimap/kernels.cuh
+++ b/include/cuco/detail/static_multimap/kernels.cuh
@@ -532,6 +532,8 @@ __global__ void pair_retrieve(InputIt first,
     idx += loop_stride;
   }
 
+  flushing_cg.sync();
+
   // Final flush of output buffer
   if (flushing_cg_counter[flushing_cg_id] > 0) {
     view.flush_output_buffer(flushing_cg,
```
is enough.

Aside: I think there is also a potential race in the `pair_retrieve` function since if the trail of `while(running)` flushes the buffer, `cg_counter` is written to, but the CG is not synced before the next iteration's read (that's `device_view_impl.inl` lines 1068 vs 1096.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

wence-:
This is needed because if we go round loops again, we might read before things have been reset.

- Tentatively closes #336.

### DIFF
--- a/include/cuco/detail/static_multimap/device_view_impl.inl
+++ b/include/cuco/detail/static_multimap/device_view_impl.inl
@@ -1002,6 +1002,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
           flushing_cg, *flushing_cg_counter, output_buffer, num_matches, output_begin);
         // First lane reset warp-level counter
         if (flushing_cg.thread_rank() == 0) { *flushing_cg_counter = 0; }
+        flushing_cg.sync();
       }
 
       current_slot = next_slot(current_slot);
@@ -1094,6 +1095,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
         flush_output_buffer(g, *cg_counter, output_buffer, num_matches, output_begin);
         // First lane reset CG-level counter
         if (lane_id == 0) { *cg_counter = 0; }
+        g.sync();
       }
       current_slot = next_slot(current_slot);
     }  // while running
@@ -1430,6 +1432,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
                             contained_output_begin);
         // First lane reset warp-level counter
         if (flushing_cg.thread_rank() == 0) { *flushing_cg_counter = 0; }
+        flushing_cg.sync();
       }
 
       current_slot = next_slot(current_slot);
@@ -1541,6 +1544,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
                             contained_output_begin);
         // First lane reset CG-level counter
         if (lane_id == 0) { *cg_counter = 0; }
+        g.sync();
       }
       current_slot = next_slot(current_slot);
     }  // while running
--- a/include/cuco/detail/static_multimap/device_view_impl.inl
+++ b/include/cuco/detail/static_multimap/device_view_impl.inl
@@ -1000,6 +1000,9 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
       if (*flushing_cg_counter + flushing_cg.size() * vector_width() > buffer_size) {
         flush_output_buffer(
           flushing_cg, *flushing_cg_counter, output_buffer, num_matches, output_begin);
+        // Everyone in the group reads the counter when flushing, so
+        // sync before writing.
+        flushing_cg.sync();
         // First lane reset warp-level counter
         if (flushing_cg.thread_rank() == 0) { *flushing_cg_counter = 0; }
         flushing_cg.sync();
@@ -1093,6 +1096,9 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
       // Flush if the next iteration won't fit into buffer
       if ((*cg_counter + g.size()) > buffer_size) {
         flush_output_buffer(g, *cg_counter, output_buffer, num_matches, output_begin);
+        // Everyone in the group reads the counter when flushing, so
+        // sync before writing.
+        g.sync();
         // First lane reset CG-level counter
         if (lane_id == 0) { *cg_counter = 0; }
         g.sync();
@@ -1430,6 +1436,9 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
                             num_matches,
                             probe_output_begin,
                             contained_output_begin);
+        // Everyone in the group reads the counter when flushing, so
+        // sync before writing.
+        flushing_cg.sync();
         // First lane reset warp-level counter
         if (flushing_cg.thread_rank() == 0) { *flushing_cg_counter = 0; }
         flushing_cg.sync();
@@ -1542,6 +1551,9 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
                             num_matches,
                             probe_output_begin,
                             contained_output_begin);
+        // Everyone in the group reads the counter when flushing, so
+        // sync before writing.
+        g.sync();
         // First lane reset CG-level counter
         if (lane_id == 0) { *cg_counter = 0; }
         g.sync();
--- a/include/cuco/detail/static_multimap/kernels.cuh
+++ b/include/cuco/detail/static_multimap/kernels.cuh
@@ -387,6 +387,8 @@ __global__ void retrieve(InputIt first,
 
   if (flushing_cg.thread_rank() == 0) { flushing_cg_counter[flushing_cg_id] = 0; }
 
+  flushing_cg.sync();
+
   while (flushing_cg.any(idx < n)) {
     bool active_flag        = idx < n;
     auto active_flushing_cg = cg::binary_partition<flushing_cg_size>(flushing_cg, active_flag);
@@ -499,6 +501,8 @@ __global__ void pair_retrieve(InputIt first,
   __shared__ uint32_t flushing_cg_counter[num_flushing_cgs];
 
   if (flushing_cg.thread_rank() == 0) { flushing_cg_counter[flushing_cg_id] = 0; }
+
+  flushing_cg.sync();
 
   while (flushing_cg.any(idx < n)) {
     bool active_flag        = idx < n;
--- a/include/cuco/detail/static_multimap/kernels.cuh
+++ b/include/cuco/detail/static_multimap/kernels.cuh
@@ -416,6 +416,7 @@ __global__ void retrieve(InputIt first,
     idx += loop_stride;
   }
 
+  flushing_cg.sync();
   // Final flush of output buffer
   if (flushing_cg_counter[flushing_cg_id] > 0) {
     view.flush_output_buffer(flushing_cg,
@@ -532,6 +533,7 @@ __global__ void pair_retrieve(InputIt first,
     idx += loop_stride;
   }
 
+  flushing_cg.sync();
   // Final flush of output buffer
   if (flushing_cg_counter[flushing_cg_id] > 0) {
     view.flush_output_buffer(flushing_cg,


include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -1094,6 +1095,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
         flush_output_buffer(g, *cg_counter, output_buffer, num_matches, output_begin);
         // First lane reset CG-level counter
         if (lane_id == 0) { *cg_counter = 0; }
+        g.sync();
```
wence-, line 1104: `*cg_counter` is read without a sync on line 1068.

include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -1002,6 +1002,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
           flushing_cg, *flushing_cg_counter, output_buffer, num_matches, output_begin);
         // First lane reset warp-level counter
         if (flushing_cg.thread_rank() == 0) { *flushing_cg_counter = 0; }
+        flushing_cg.sync();
```
wence-, line 1008: This one might not be needed because the counter is read again for the first time by lane 0, but with an atomicAdd (does that provide an ordering guarantee wrt the write?).

include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -1430,6 +1432,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
                             contained_output_begin);
         // First lane reset warp-level counter
         if (flushing_cg.thread_rank() == 0) { *flushing_cg_counter = 0; }
+        flushing_cg.sync();
```
wence-, line 1444: counter is read with an atomicAdd on the next iteration, so same comment as above.

include/cuco/detail/static_multimap/device_view_impl.inl
```diff
@@ -1541,6 +1544,7 @@ class static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::device_view_
                             contained_output_begin);
         // First lane reset CG-level counter
         if (lane_id == 0) { *cg_counter = 0; }
+        g.sync();
```
wence-, line 1559: counter is read without a sync on the next loop iteration.

include/cuco/detail/static_multimap/kernels.cuh
```diff
@@ -416,6 +416,7 @@ __global__ void retrieve(InputIt first,
     idx += loop_stride;
   }
 
+  flushing_cg.sync();
```
wence-, line 421: If the loop doesn't iterate at all, then we never sync the CG before the read on the following line. Additionally, `retrieve` and `retrieve_outer` only sync the `active_flushing_cg` (which might not sync all of `flushing_cg` AIUI).

include/cuco/detail/static_multimap/kernels.cuh
```diff
@@ -532,6 +533,7 @@ __global__ void pair_retrieve(InputIt first,
...
```
wence-, line 540: Same as above.

wence- (commented):
I am not sure _all_ of these syncs are necessary, but have commented on my rationale for them.

PointKernel:
@wence- Thanks for putting up the PR. Is this ready for review?

wence-:
> @wence- Thanks for putting up the PR. Is this ready for review?

I think so. I think I don't understand fully the semantics of some of the early returns in some of the device functions, but I think these bits are right

include/cuco/detail/static_multimap/device_view_impl.inl
PointKernel, line 1008:
```suggestion
```
This is redundant since `any` would sync internally
https://github.com/NVIDIA/cuCollections/blob/88ff1e4eb7bbdd5a0d1c9025548e1451d8128a93/include/cuco/detail/static_multimap/device_view_impl.inl#L951

include/cuco/detail/static_multimap/device_view_impl.inl
PointKernel, line 1104:
```suggestion
```
Seems redundant also since the read on line 1068 is right after a CG `ballot`. Similar to warp vote functions, CG vote functions like `ballot`, `any` and `all` imply an internal sync per [CUDA programming guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#thread-block-tile).

PointKernel (commented):
I take a closer look at the proposed changes and find they might not be necessary since CG vote functions imply synchronizations. @wence- Does the race check pass after adding those syncs?

include/cuco/detail/static_multimap/device_view_impl.inl
wence-: So `__ballot_sync` which is what the documentation for `g.ballot()` refers to in [warp vote functions](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-vote-functions) explicitly states

> These intrinsics do not imply a memory barrier. They do not guarantee any memory ordering.

So I _think_ that means without explicit synchronisation there is no happens-before ordering between the lane-zero write to `*cg_counter` and the read in the next loop iteration.

There's no problem I think between the write on line 1079 and the subsequent write on line 1097 because those are both done by the same thread.

include/cuco/detail/static_multimap/device_view_impl.inl
wence-: Note also that the `atomicAdd` is on lane-zero of `probing_cg` whereas this write is on lane-zero of `flushing_cg` (so those _might_ be different threads).

wence-: As below, at least at the warp level (`__any_sync`), the intrinsic doesn't provide any memory ordering guarantees per https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-vote-functions

wence-: > but with an atomicAdd (does that provide an ordering guarantee wrt the write?).

`atomicAdd` has [relaxed memory order](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions) so non-atomic reads/writes to the same location can migrate past the atomic read/write I believe. I guess this is why it is better to use `atomic_ref` because that protects a variable, not a location.

wence-:
> I take a closer look at the proposed changes and find they might not be necessary since CG vote functions imply synchronizations. @wence- Does the race check pass after adding those syncs?

The modifications in kernels.cuh (to sync after the loop) are the ones necessary to pass race check. The modifications in device_view_impl.inl are not necessary for that test, but _I think_ still necessary for correctness (I've responded to your comments to explain why I think they are necessary: basically, the warp-level vote functions don't provide memory barriers).

PointKernel (commented):
@wence- You are right. I was missing the point that warp vote functions don't communicate through memory and they don't guarantee any memory ordering.

To uncover more similar issues like this, I build your branch locally and run `compute-sanitizer` on multimap unit tests:
```bash
 compute-sanitizer --tool racecheck ./STATIC_MULTIMAP_TEST
```
The race check shows that there are 13 warnings with the current code. We can fix them all in this PR if you have the bandwidth. Otherwise, I will merge the PR as it is and put up follow-up PRs to solve the issue. I'm happy with either option.

wence-:
> @wence- You are right. I was missing the point that warp vote functions don't communicate through memory and they don't guarantee any memory ordering.
> 
> To uncover more similar issues like this, I build your branch locally and run `compute-sanitizer` on multimap unit tests:
> 
> ```shell
>  compute-sanitizer --tool racecheck ./STATIC_MULTIMAP_TEST
> ```
> 
> The race check shows that there are 13 warnings with the current code. We can fix them all in this PR if you have the bandwidth. Otherwise, I will merge the PR as it is and put up follow-up PRs to solve the issue. I'm happy with either option.

Let me have a go next week...

wence-:
I have pushed fixes. `STATIC_MULTIMAP_TEST` is now racecheck clean with cuda 11.8 and 12.2, it is _not_ clean with 12.0 which I think means either there are compute-sanitizer false positives or else bugs in 12.0.

PointKernel (approved):
LGTM. Thanks!

PR #344, opened 2023-08-02
Fix doxygen check with unsupported version.

bdice:
While running `pre-commit`, I got this error:

```bash
./ci/checks/doxygen.sh: line 19: ￼: command not found
./ci/checks/doxygen.sh: line 20: ￼: command not found
./ci/checks/doxygen.sh: line 21: ￼: command not found
```

I have doxygen 1.9.7 installed, which isn't a supported version in this script. The Doxygen check is supposed to pass silently when run locally with an unsupported version (it runs properly in CI, instead).

It seems like there's an unrecognized symbol in the bash script from PR #177: https://github.com/NVIDIA/cuCollections/commit/05fb1dbe9fed4f446cecb0b024232ede1dea6926

This PR removes that symbol, fixing the error.


PointKernel (approved):

bdice:
@PointKernel Does CI need to pass before a PR is merged? This is still running GPU tests. I don't expect a failure, just wasn't sure if CI was supposed to be required or not.

PointKernel:
> Does CI need to pass before a PR is merged? 

For actual cpp/CUDA changes, I would wait for CI to complete before merging the PR. This work only touches the doxygen file and our style checks passed already thus the remaining cpp tests are not mandatory.

PR #345, opened 2023-08-02
Improve implememtation details in experimental data structures

PointKernel:
This PR fixes issues and adds new features requested by https://github.com/rapidsai/cudf/pull/13807.

It:

- removes the requirement of the second hasher from double hashing must be constructible from an integer
- fixes an issue in map iterator `!=` operator
- overloads map iterator access operator
- allows zero capacity container
- adds `capacity_test` back since several corner cases need to be exercised

### DIFF
--- a/include/cuco/detail/extent/extent.inl
+++ b/include/cuco/detail/extent/extent.inl
@@ -80,8 +80,8 @@ template <int32_t CGSize, int32_t WindowSize, typename SizeType, std::size_t N>
     (static_cast<uint64_t>(std::numeric_limits<SizeType>::max()) < max_prime)
       ? std::numeric_limits<SizeType>::max()
       : static_cast<SizeType>(max_prime);
-  auto const size = SDIV(ext, CGSize * WindowSize);
-  if (size < 0 or size > max_value) { CUCO_FAIL("Invalid input extent"); }
+  auto const size = SDIV(std::max(ext, 1), CGSize * WindowSize);
+  if (size > max_value) { CUCO_FAIL("Invalid input extent"); }
 
   if constexpr (N == dynamic_extent) {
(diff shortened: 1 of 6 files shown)


sleeepyjack:
`h1==h2` isn't ideal. If two distinct keys suffer from a hash collision, they will follow the same exact probing pattern. Using two different hash functions reduces the likelihood of running into these so-called secondary clustering effects.

PointKernel:
> `h1==h2` isn't ideal. If two distinct keys suffer from a hash collision, they will follow the same exact probing pattern. Using two different hash functions reduces the likelihood of running into these so-called secondary clustering effects.

Agreed. The following code will fail if the `custom_hasher` doesn't have an int ctor:
```cpp
  auto my_probe = cuco::experimental::double_hashing<custom_hasher>(custom_hasher{});
```
This makes me realize that our current code is based on the assumption that only one hasher type is used in double hashing and the second hasher _must_ have an integer ctor:
https://github.com/NVIDIA/cuCollections/blob/6bc62c6fc9d2edbd1dacbe1b242b173d4cffad99/include/cuco/probing_scheme.cuh#L112
This is a too-restrictive requirement and using two identical hashers is the first step to lower the bar. I haven't figured out a way to avoid the secondary collision yet by following this idea.

PointKernel:
> `h1==h2` isn't ideal. If two distinct keys suffer from a hash collision, they will follow the same exact probing pattern. Using two different hash functions reduces the likelihood of running into these so-called secondary clustering effects.

Changes reverted at https://github.com/NVIDIA/cuCollections/pull/345/commits/8ac0ee87d7e5a23edf583c5d8357dd77f437443d since it requires a larger scope of adjustment/cleanup and will be in a separate PR.

PR #350, opened 2023-08-11
Trie

amukkara:
Trie and associated bit-vector code with tests

PR #352, opened 2023-08-17
Add `dynamic_bitset`

amukkara:
This PR adds `dynamic_bitset` code that will be used in Trie data structure. Trie will be integrated in a separate PR #350.

Since `dynamic_bitset` is not intended to be part of public-facing API, all files (.cuh and .inl) are located in include/cuco/detail/trie/dynamic_bitset

Tests are added in tests/dynamic_bitset


GPUtester:

Admins can comment `ok to test` to allow this one PR to run or `add to allowlist` to allow all future PRs from the same author to run.

PointKernel:
add to allowlist

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+namespace cuco {
+namespace experimental {
+
+/**
+ * @brief Struct to store ranks of bits at 256-bit intervals
+ */
+struct rank {
+  uint32_t abs_hi_;              ///< Upper 32 bits of base
+  uint8_t abs_lo_;               ///< Lower 8 bits of base
+  std::array<uint8_t, 3> rels_;  ///< Offsets for 64-bit sub-intervals
```
PointKernel:
```suggestion
  cuda::std::array<uint8_t, 3> rels_;  ///< Offsets for 64-bit sub-intervals
```
`cuda::std::array` is preferred since the data would be accessed from device

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  /**
+   * @brief Sets base rank of current 256-bit interval
+   *
+   * @param abs Base rank
+   */
+  void set_abs(uint64_t abs)
+  {
+    abs_hi_ = static_cast<uint32_t>(abs >> 8);
```
PointKernel: Note to myself. Should find a way to get rid of magic numbers like `8` and `3`

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+  uint32_t abs_hi_;              ///< Upper 32 bits of base
+  uint8_t abs_lo_;               ///< Lower 8 bits of base
+  std::array<uint8_t, 3> rels_;  ///< Offsets for 64-bit sub-intervals
+
+  /**
+   * @brief Gets base rank of current 256-bit interval
+   *
+   * @return The base rank
+   */
+  __host__ __device__ uint64_t abs() const
```
PointKernel:
```suggestion
  __host__ __device__ constexpr uint64_t abs() const noexcept
```
`constexpr` and `noexcept` when possible

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+  {
+    return (static_cast<uint64_t>(abs_hi_) << 8) | abs_lo_;
+  }
+
+  /**
+   * @brief Sets base rank of current 256-bit interval
+   *
+   * @param abs Base rank
+   */
+  void set_abs(uint64_t abs)
```
PointKernel:
```suggestion
  constexpr void set_abs(uint64_t abs) noexcept
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+ * @tparam Extent Data structure size type
+ * @tparam Scope The scope in which operations will be performed by individual threads.
+ * @tparam Allocator Type of allocator used for device storage
+ * @tparam Storage Slot window storage type
+ */
+
+template <class Extent             = cuco::experimental::extent<std::size_t>,
+          cuda::thread_scope Scope = cuda::thread_scope_device,
+          class Allocator          = cuco::cuda_allocator<std::byte>,
+          class Storage            = cuco::experimental::storage<1>>
```
PointKernel:
```suggestion
template <class Allocator          = cuco::cuda_allocator<std::byte>>
```

- `Scope` seems not related to this work
- Not sure bit vector would benefit from the cuco extent type, we could just use `using size_type = size_t;` internally
- Storage doesn't need to be exposed since tuning `window_size` doesn't apply in this case.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+  void build();
+
+  using size_type = typename Extent::value_type;  ///< size type to specify bit index
+  /**
+   * @brief Modifies a single bit
+   *
+   * @param index position of bit to be modified
+   * @param bit new value of bit
+   */
+  void set(size_type index, bool bit);
```
PointKernel:
```suggestion
  void set(size_type index, bool bit) noexcept;
```
Similar to other cuco host operations: would a "bulk" host API work here? Something like:
```cpp
void set(std::span<size_type> indices, std::span<bool> bit) noexcept;
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+  ~bit_vector();
+
+  /**
+   * @brief adds a new bit at the end
+   *
+   * Grows internal storage if needed
+   *
+   * @param bit Boolean value of new bit to be added
+   */
+  void append(bool bit);
```
PointKernel:
```suggestion
  void append(bool bit) noexcept;
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * @brief Builds indexes for rank and select
+   *
+   * Also creates device-side snapshot
+   */
+  void build();
```
PointKernel:
```suggestion
  void build() noexcept;
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+   * @param bit new value of bit
+   */
+  void set(size_type index, bool bit);
+
+  /**
+   * @brief Sets last bit to specified value
+   *
+   * @param bit new value of last bit
+   */
+  void set_last(bool bit);
```
PointKernel:
```suggestion
  void set_last(bool bit) noexcept;
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+  void append(bool bit);
+
+  /**
+   * @brief Builds indexes for rank and select
+   *
+   * Also creates device-side snapshot
+   */
+  void build();
+
+  using size_type = typename Extent::value_type;  ///< size type to specify bit index
```
PointKernel:
```suggestion
  using size_type = std::size_t;  ///< size type to specify bit index
```
And use `size_type` throughout the whole class.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+   * @param bit new value of last bit
+   */
+  void set_last(bool bit);
+
+  static constexpr auto cg_size      = 1;      ///< CG size used for probing
+  static constexpr auto window_size  = 1;      ///< Window size used for probing
+  static constexpr auto thread_scope = Scope;  ///< CUDA thread scope
+
+  using extent_type =
+    decltype(make_window_extent<cg_size, window_size>(std::declval<Extent>()));  ///< Extent type
```
PointKernel:
```suggestion
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+  using allocator_type = Allocator;                                              ///< Allocator type
+  using storage_type =
+    detail::storage<Storage, size_type, extent_type, allocator_type>;  ///< Storage type
```
PointKernel:
```suggestion
  using slot_type = uint64_t;
  using storage_type =
    aow_storage<slot_type, 1, size_type, allocator_type>;  ///< Storage type
```
The slot type should be independent of `size_type`. Imagine `using size_type = int32_t`;
Also, can use `aow_storage` directly to simplify the logic.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+ * @tparam Storage Slot window storage type
+ */
+
+template <class Extent             = cuco::experimental::extent<std::size_t>,
+          cuda::thread_scope Scope = cuda::thread_scope_device,
+          class Allocator          = cuco::cuda_allocator<std::byte>,
+          class Storage            = cuco::experimental::storage<1>>
+class bit_vector {
+ public:
+  bit_vector();
```
PointKernel:
```suggestion
  bit_vector(Allocator const& allocator = Allocator{});
```
We need to at least pass a default allocator

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+
+#include <thrust/device_vector.h>
+
+#include <cuda/atomic>
+
+namespace cuco {
+namespace experimental {
+
+/**
+ * @brief Struct to store ranks of bits at 256-bit intervals
```
PointKernel: Is 256 chosen for performance purposes?

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * @brief Get the number of bits bit_vector holds
+   *
+   * @return Number of bits bit_vector holds
+   */
+  size_t size() const { return n_bits_; }
+
+ private:
+  uint64_t n_bits_;  ///< Number of bits added to bit_vector
```
PointKernel:
```suggestion
  size_type n_bits_;  ///< Number of bits added to bit_vector
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,200 @@
...
+
+  // Host-side structures
+  std::vector<uint64_t> words_;     ///< Words vector that represents all bits
+  std::vector<rank> ranks_;         ///< Holds the rank values for every 256-th bit (4-th word)
+  std::vector<rank> ranks0_;        ///< Same as ranks_ but for `0` bits
+  std::vector<uint64_t> selects_;   ///< Holds pointers to (0, 256, 512...)th `1` bit in ranks_
+  std::vector<uint64_t> selects0_;  ///< Same as selects_, but for `0` bits
```
PointKernel:
```suggestion
  std::vector<slot_type> words_;     ///< Words vector that represents all bits
  std::vector<rank> ranks_;         ///< Holds the rank values for every 256-th bit (4-th word)
  std::vector<rank> ranks0_;        ///< Same as ranks_ but for `0` bits
  std::vector<size_type> selects_;   ///< Holds indices to (0, 256, 512...)th `1` bit in ranks_
  std::vector<size_type> selects0_;  ///< Same as selects_, but for `0` bits
```
I could be wrong, but the idea here is to use `size_type` and `slot_type` consistently to avoid plain `uint64_t`.

PointKernel (commented):
Thanks for putting up the PR! :fire: 

Just a quick pass of the top-level bitvector header. Some are the issues that need to be solved and some are just notes to myself. Feel free to address reviews on any topics you are comfortable dealing with. @sleeepyjack and I would take care of the code-style or software engineering stuff.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: 3894d5a984023a9e417c7f9e409f742d9b76fc76

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: 10c09006eea55817ad931214a4b40685611a6fef

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: d063cc50bc0b5ad0b02b8f032ba75c4a55b2f968

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: d063cc50bc0b5ad0b02b8f032ba75c4a55b2f968

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: a371e27b6c612e7a61bee2fe6a6e70654f006051

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: 91b628a5925bec0b4a4c7f21dc6ebad40c8ba0d9

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: It is a tradeoff between space efficiency and performance. Larger values will create a smaller index, but searches will be slower.

This is hardcoded to 256 (four 64-bit words) now, but can be made configurable. Main change would be in `rank` structure which stores a single base and deltas for each 64-bit word.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,223 @@
...
+   * @return Number of bits bit_vector holds
+   */
+  size_type constexpr size() const noexcept { return n_bits_; }
+
+ private:
+  size_type n_bits_;  ///< Number of bits bit_vector currently holds
+
+  const size_type bits_per_word   = sizeof(slot_type) * 8;  ///< Bits in a word
+  const size_type words_per_block = 4;  ///< Provides tradeoff between space efficiency and perf.
+  const size_type bits_per_block  = words_per_block * bits_per_word;
```
PointKernel:
```suggestion
  constexpr size_type bits_per_word   = sizeof(slot_type) * 8;  ///< Bits in a word
  constexpr size_type words_per_block = 4;  ///< Provides tradeoff between space efficiency and perf.
  constexpr size_type bits_per_block  = words_per_block * bits_per_word;
```
Also, I feel those should be public thus downstream objects can use to eliminate magic numbers

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,223 @@
...
+                              slot_type word,
+                              size_type count_in,
+                              std::vector<size_type>& selects) noexcept;
+};
+
+}  // namespace experimental
+}  // namespace cuco
+
+#include <cuco/detail/trie/bit_vector/bit_vector.inl>
+#include <cuco/detail/trie/bit_vector/bit_vector_ref.inl>
```
PointKernel:
```suggestion
```

include/cuco/detail/trie/bit_vector/bit_vector_ref.cuh
```diff
@@ -0,0 +1,47 @@
...
+ private:
+  storage_ref_type words_ref_, ranks_ref_, selects_ref_, ranks0_ref_, selects0_ref_;
+
+  // Mixins need to be friends with this class in order to access private members
+  template <typename Op, typename Ref>
+  friend class detail::operator_impl;
+};
+
+}  // namespace experimental
+}  // namespace cuco
```
PointKernel:
```suggestion
}  // namespace cuco

#include <cuco/detail/trie/bit_vector/bit_vector_ref.inl>
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,223 @@
...
+#pragma once
+
+#include <cuco/detail/trie/bit_vector/bit_vector_ref.cuh>
+#include <cuco/extent.cuh>
+#include <cuco/storage.cuh>
+#include <cuco/utility/allocator.hpp>
+
+#include <thrust/device_vector.h>
+
+#include <cuda/atomic>
```
PointKernel:
```suggestion
```
seems not used

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,223 @@
...
+  {
+    return (static_cast<uint64_t>(abs_hi_) << 8) | abs_lo_;
+  }
+
+  /**
+   * @brief Sets base rank of current 256-bit interval
+   *
+   * @param abs Base rank
+   */
+  void constexpr set_abs(uint64_t abs) noexcept
```
PointKernel: Does this need to be `__host__ __device__`? as `abs()`?

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,223 @@
...
+ *
+ * In addition to standard bitvector get/set operations, this class provides
+ * rank and select operation API. It maintains index structures to make both these
+ * new operations close to constant time.
+ * Bitvector construction happens on host, after which the structures are moved to device.
+ * All subsequent read-only operations access device structures only.
+ *
+ * @tparam Allocator Type of allocator used for device storage
+ */
+
```
PointKernel:
```suggestion
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,223 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cuco/detail/trie/bit_vector/bit_vector_ref.cuh>
+#include <cuco/extent.cuh>
```
PointKernel:
```suggestion
```

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,221 @@
...
+  }
+};
+
+/**
+ * @brief Union of 64-bit word with rank
+ *
+ * Need this so that all aow_storage structures in bitvector have 64-bit element type
+ */
+union rank_union {
+  uint64_t word_;  ///< word view
```
PointKernel: Note: this should be `slot_type` instead of plain integer.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * In addition to standard bitvector get/set operations, this class provides
+ * rank and select operation API. It maintains index structures to make both these
+ * new operations close to constant time.
+ * Bitvector construction happens on host, after which the structures are moved to device.
+ * All subsequent read-only operations access device structures only.
+ *
+ * @tparam Allocator Type of allocator used for device storage
+ */
+
```
sleeepyjack:
```suggestion
```
Remove empty line so the Doxygen style check won't complain.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,238 @@
...
+template <class Allocator = cuco::cuda_allocator<std::byte>>
+class bit_vector {
+ public:
+  bit_vector(Allocator const& allocator = Allocator{});
```
sleeepyjack: Missing docs (Doxygen style check will complain)

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,238 @@
...
+  ~bit_vector();
+
+  /**
+   * @brief adds a new bit at the end
+   *
+   * Grows internal storage if needed
```
sleeepyjack: I agree with @PointKernel that since this data structure can grow dynamically, we should add a `dynamic_` prefix to the name. We can take care of this.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,238 @@
...
+   * @return Number of bits bit_vector holds
+   */
+  size_type constexpr size() const noexcept { return n_bits_; }
+
+ private:
+  size_type n_bits_;  ///< Number of bits bit_vector currently holds
+
+  const size_type bits_per_word   = sizeof(slot_type) * 8;  ///< Bits in a word
+  const size_type words_per_block = 4;  ///< Provides tradeoff between space efficiency and perf.
+  const size_type bits_per_block  = words_per_block * bits_per_word;
```
sleeepyjack:
```suggestion
  static constexpr size_type bits_per_word   = sizeof(slot_type) * 8;  ///< Bits in a word
  static constexpr size_type words_per_block = 4;  ///< Provides tradeoff between space efficiency and perf.
  static constexpr size_type bits_per_block  = words_per_block * bits_per_word;
```

include/cuco/detail/trie/bit_vector/bit_vector.inl
```diff
@@ -0,0 +1,241 @@
...
+}
+
+template <class Allocator>
+bit_vector<Allocator>::~bit_vector()
+{
+  delete aow_words_;
+  delete aow_ranks_;
+  delete aow_selects_;
+  delete aow_ranks0_;
+  delete aow_selects0_;
```
sleeepyjack: It is easy to forget a deletion here, and thus run into a memory leak. One way to ship around that is to define the members as `std::unique_ptr`, which takes care of managing the object lifetime.

@PointKernel would it make sense to apply a PIMPL idiom to this class?

include/cuco/detail/trie/bit_vector/bit_vector.inl
sleeepyjack, line 1: Since this class is no longer in the public-facing API, it would be ok to move the implementation to the `.cuh` file and remove the `.inl` file. Not a requirement though.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
sleeepyjack: Also, this data structure comes with some additional bells and whistles compared to a plain bit vector and is also specific to the trie implementation. Can we come up with a more descriptive name?

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,238 @@
...
+  template <typename KeyIt, typename OutputIt>
+  void get(KeyIt keys_begin,
+           KeyIt keys_end,
+           OutputIt outputs_begin,
+           cuda_stream_ref stream = {}) const noexcept;
+
+  using allocator_type = Allocator;  ///< Allocator type
+  using slot_type      = uint64_t;   ///< Slot type
+  using storage_type =
+    aow_storage<slot_type, 1, extent<size_type>, allocator_type>;  ///< Storage type
```
sleeepyjack: The advantage of using `aow_storage` is that it enables vectorized loads of windows. Since you're not relying on that feature (window size is 1), I tend to say that we don't need this specialized storage. Instead, you can use `thrust::host_vector` and `thrust::device_vector` for your storage classes, which should also simplify some of the data migration parts.

include/cuco/detail/trie/bit_vector/bit_vector.inl
```diff
@@ -0,0 +1,241 @@
...
+namespace cuco {
+namespace experimental {
+
+template <class Allocator>
+bit_vector<Allocator>::bit_vector(Allocator const& allocator)
+  : words_(),
+    ranks_(),
+    selects_(),
+    n_bits_(0),
+    allocator_(allocator),
```
sleeepyjack: Do we actually use this allocator?

include/cuco/detail/trie/bit_vector/bit_vector.inl
```diff
@@ -0,0 +1,241 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace cuco {
+namespace experimental {
+
+template <class Allocator>
+bit_vector<Allocator>::bit_vector(Allocator const& allocator)
+  : words_(),
```
sleeepyjack: Minor style thingy: 
```suggestion
  : words_{},
```

include/cuco/detail/trie/bit_vector/bit_vector.inl
```diff
@@ -0,0 +1,241 @@
...
+
+template <class Allocator>
+void bit_vector<Allocator>::set_last(bool bit) noexcept
+{
+  set(n_bits_ - 1, bit);
+}
+
+template <class Allocator>
+template <typename KeyIt, typename OutputIt>
+void bit_vector<Allocator>::get(KeyIt keys_begin,
```
sleeepyjack: Iterator-based algorithms typically pass the end of the output (i.e., `outputs_end`) as the return value. Not sure if that makes sense here, since there is a 1:1 relationship between inputs and outputs.

include/cuco/detail/trie/bit_vector/bit_vector.inl
```diff
@@ -0,0 +1,241 @@
...
+    outputs[key_id] = ref.get(keys[key_id]);
+    key_id += loop_stride;
+  }
+}
+
+template <class Allocator>
+void bit_vector<Allocator>::build() noexcept
+{
+  build_ranks_and_selects(words_, ranks_, selects_, false);   // 1-bits
+  build_ranks_and_selects(words_, ranks0_, selects0_, true);  // 0-bits
```
sleeepyjack: This sure would be a candidate for rudimentary parallelization, i.e., one thread per function call.

include/cuco/detail/trie/bit_vector/bit_vector.inl
```diff
@@ -0,0 +1,241 @@
...
+}
+
+template <class Allocator>
+void bit_vector<Allocator>::add_selects_entry(size_type word_id,
+                                              slot_type word,
+                                              size_type count,
+                                              std::vector<size_type>& selects) noexcept
+{
+  while (word != 0) {
+    size_type pos = __builtin_ctzll(word);
```
sleeepyjack: This is a GCC builtin. There's probably a portable way of doing this using libcudacxx backports from the stdlib `<bit>` header.

include/cuco/detail/trie/bit_vector/bit_vector.inl
```diff
@@ -0,0 +1,241 @@
...
+  }
+}
+
+template <class Allocator>
+template <class T>
+void bit_vector<Allocator>::copy_host_array_to_aow(storage_type** aow,
+                                                   std::vector<T>& host_array) noexcept
+{
+  uint64_t num_elements = host_array.size();
+  *aow                  = new storage_type(extent<size_type>{num_elements + 1}, allocator_);
```
sleeepyjack: Plain `new` is considered harmful ;)

include/cuco/detail/trie/bit_vector/bit_vector_ref.cuh
```diff
@@ -0,0 +1,47 @@
...
+   * @param selects_ref Non-owning ref of selects slot storage
+   * @param ranks0_ref Non-owning ref of ranks0 slot storage
+   * @param selects0_ref Non-owning ref of selects0 slot storage
+   */
+  __host__ __device__ explicit constexpr bit_vector_ref(storage_ref_type words_ref,
+                                                        storage_ref_type ranks_ref,
+                                                        storage_ref_type selects_ref,
+                                                        storage_ref_type ranks0_ref,
+                                                        storage_ref_type selects0_ref) noexcept;
+
```
sleeepyjack: Passing each storage one-by-one is suboptimal, since they might be incompatible (e.g. different lengths). Since this is not user-facing code, it's not a huge problem, but consider wrapping these in a single struct which internally checks for the length requirement.

include/cuco/detail/trie/bit_vector/bit_vector_ref.cuh
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include <cuco/operator.hpp>
+
+namespace cuco {
+namespace experimental {
```
sleeepyjack: This is declared in public namespace, but we want it under detail namespace.

include/cuco/detail/trie/bit_vector/bit_vector_ref.cuh
```diff
@@ -0,0 +1,47 @@
...
+                                                        storage_ref_type selects_ref,
+                                                        storage_ref_type ranks0_ref,
+                                                        storage_ref_type selects0_ref) noexcept;
+
+ private:
+  storage_ref_type words_ref_, ranks_ref_, selects_ref_, ranks0_ref_, selects0_ref_;
+
+  // Mixins need to be friends with this class in order to access private members
+  template <typename Op, typename Ref>
+  friend class detail::operator_impl;
```
sleeepyjack: I'm reconsidering if the operator mixin design makes sense here. The only device-sided operations are all `const`, so there's no benefit in dispatching optimized code paths based on the operations used. Plain `__device__` member functions should be equivalent and are less complex to implement.

include/cuco/detail/trie/bit_vector/bit_vector_ref.inl
```diff
@@ -0,0 +1,238 @@
...
+namespace detail {
+
+template <typename StorageRef, typename... Operators>
+class operator_impl<op::bv_read_tag, bit_vector_ref<StorageRef, Operators...>> {
+  using ref_type  = bit_vector_ref<StorageRef, Operators...>;  ///< Bitvector ref type
+  using size_type = typename StorageRef::size_type;            ///< Size type
+  using slot_type = typename StorageRef::value_type;           ///< Slot type
+
+  const size_type bits_per_word   = sizeof(slot_type) * 8;
+  const size_type words_per_block = 4;  //< This should match the defintion in bit_vector
```
sleeepyjack: These can be `constexpr`.

include/cuco/detail/trie/bit_vector/bit_vector_ref.inl
```diff
@@ -0,0 +1,238 @@
...
+   *
+   * @param key Position of starting bit
+   *
+   * @return Index of next set bit
+   */
+  [[nodiscard]] __device__ size_type find_next_set(size_type key) const noexcept
+  {
+    auto const& ref_  = static_cast<ref_type const&>(*this);
+    size_type word_id = key / bits_per_word;
+    size_type bit_id  = key % bits_per_word;
```
sleeepyjack: Declaring `bits_per_word`as `constexpr` should speed up these otherwise expensive operations `/` and `%` considerably.

include/cuco/detail/trie/bit_vector/bit_vector_ref.inl
```diff
@@ -0,0 +1,238 @@
...
+    size_type rank_id = word_id / words_per_block;
+    size_type rel_id  = word_id % words_per_block;
+
+    auto rank   = rank_union{ref_.ranks_ref_[rank_id][0]}.rank_;
+    size_type n = rank.abs();
+
+    if (rel_id != 0) { n += rank.rels_[rel_id - 1]; }
+
+    n += __popcll(ref_.words_ref_[word_id][0] & ((1UL << bit_id) - 1));
```
sleeepyjack: I think we now have `cuda::std::popcount`. Although it's just a style thing, since there's no point in adding portability here.

tests/CMakeLists.txt
```diff
@@ -95,3 +95,13 @@ ConfigureTest(STATIC_MULTIMAP_TEST
...
+
+###################################################################################################
+# - bit_vector tests ------------------------------------------------------------------------------
+ConfigureTest(BIT_VECTOR_TEST
+    bit_vector/find_next_set_test.cu
+    bit_vector/get_test.cu
+    bit_vector/set_test.cu
+    bit_vector/rank_test.cu
+    bit_vector/select_test.cu
+    bit_vector/size_test.cu)
```
sleeepyjack: We should move these to `trie/bit_vector/...` in the trie PR

tests/bit_vector/find_next_set_test.cu
```diff
@@ -0,0 +1,74 @@
...
+{
+  size_t index  = blockIdx.x * blockDim.x + threadIdx.x;
+  size_t stride = gridDim.x * blockDim.x;
+  while (index < n) {
+    output[index] = ref.find_next_set(index);
+    index += stride;
+  }
+}
+
+extern bool modulo_bitgen(uint64_t i);
```
sleeepyjack: Where is this declared?

tests/bit_vector/find_next_set_test.cu
```diff
@@ -0,0 +1,74 @@
...
+#include <cuco/detail/trie/bit_vector/bit_vector.cuh>
+
+#include <thrust/device_vector.h>
+#include <thrust/execution_policy.h>
+#include <thrust/host_vector.h>
+
+#include <catch2/catch_test_macros.hpp>
+
+template <class BitVectorRef>
+__global__ void find_next_set_kernel(BitVectorRef ref, size_t n, uint64_t* output)
```
sleeepyjack: If you add a tparam `ÒutputIt`, you won't need the `thrust::raw_pointer_cast` at the call site.

tests/bit_vector/get_test.cu
```diff
@@ -0,0 +1,70 @@
...
+{
+  size_t index  = blockIdx.x * blockDim.x + threadIdx.x;
+  size_t stride = gridDim.x * blockDim.x;
+  while (index < n) {
+    output[index] = ref.get(index);
+    index += stride;
+  }
+}
+
+bool modulo_bitgen(uint64_t i) { return i % 7 == 0; }
```
sleeepyjack: Ah, I see. So after linking this function should be available to the other test.

include/cuco/detail/trie/bit_vector/bit_vector_ref.inl
```diff
@@ -202,6 +202,37 @@ class operator_impl<op::bv_read_tag, bit_vector_ref<StorageRef, Operators...>> {
   }
 };
 
+template <typename StorageRef, typename... Operators>
+class operator_impl<op::bv_set_tag, bit_vector_ref<StorageRef, Operators...>> {
+  using ref_type                = bit_vector_ref<StorageRef, Operators...>;  ///< Bitvector ref type
+  using size_type               = typename StorageRef::size_type;            ///< Size type
+  using slot_type               = typename StorageRef::value_type;           ///< Slot type
+  const size_type bits_per_word = sizeof(slot_type) * 8;
```
sleeepyjack: There's one catch with these mixin operators: You cannot have any runtime members in the operator class because we're casting the operator to the `ref_type`, and after doing that, the operator's members occupy the same space as the `ref_type` members. The `static_cast<ref_type&>(*this)` and then accessing any of `ref_type`'s members will get you in trouble.
In this particular case, you can just declare `bits_per_word` as `constexpr` to get rid of the problem.

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: 9d5100d2c20347acf18d4d10a30ea76db159f13e

include/cuco/detail/trie/bit_vector/bit_vector.inl
amukkara: e248c9d26f40fcaeca19fa85ce9c313595c6bc03

include/cuco/detail/trie/bit_vector/bit_vector_ref.inl
amukkara: 9d5100d2c20347acf18d4d10a30ea76db159f13e

tests/bit_vector/find_next_set_test.cu
amukkara: Added comments to avoid confusion. dbd5313393815583a1a2766880cfda0f45d7bcab

tests/bit_vector/find_next_set_test.cu
amukkara: 6028a794dff6a1e3daf1a8b4d3980ee7c172601b

include/cuco/detail/trie/bit_vector/bit_vector.cuh
PointKernel: Using `aow_storage` would also benefit from its member functions like `initialize`. No strong preference here.

include/cuco/detail/trie/bit_vector/bit_vector.inl
PointKernel: This allocator should be passed to all storage data members.

include/cuco/detail/trie/bit_vector/bit_vector.inl
```diff
@@ -0,0 +1,277 @@
...
+    ranks0_{},
+    selects_{},
+    selects0_{},
+    n_bits_(0),
+    allocator_(allocator),
+    aow_words_(nullptr),
+    aow_ranks_(nullptr),
+    aow_selects_(nullptr),
+    aow_ranks0_(nullptr),
+    aow_selects0_(nullptr)
```
PointKernel:
```suggestion
    n_bits_{0},
    allocator_[allocator},
    aow_words_{nullptr},
    aow_ranks_{nullptr},
    aow_selects_{nullptr},
    aow_ranks0_{nullptr},
    aow_selects0_{nullptr}
```
Always prefer curly braces over parentheses in initialization list

include/cuco/detail/trie/bit_vector/bit_vector.inl
PointKernel: This should be fine. Unlike multimap or single map insertion where the output number doesn't necessarily match the input number, we always have one output for a given input

PointKernel (commented):
Partial review. Some of the comments could be out of date.

include/cuco/detail/trie/bit_vector/bit_vector.inl
amukkara: 8687daa9ca744bd61826df0a9147f395047a07fd

include/cuco/detail/trie/bit_vector/bit_vector.inl
amukkara: Added unique_ptrs eb7f9572fd61fc3d052e53736f4a2b47a7bf72a1

include/cuco/detail/trie/bit_vector/bit_vector.inl
amukkara: eb7f9572fd61fc3d052e53736f4a2b47a7bf72a1

include/cuco/detail/trie/bit_vector/bit_vector_ref.cuh
amukkara: should all bitvector code go under coco::detail?

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: f74bee9370517e91b1dbcd1af1afcfc9e8052f24

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: f74bee9370517e91b1dbcd1af1afcfc9e8052f24

include/cuco/detail/trie/bit_vector/bit_vector.inl
amukkara: 005bd5d42177a3239e4b23545e9f328e7bb07633

include/cuco/detail/trie/bit_vector/bit_vector.inl
amukkara: allocator is used only once, as an argument to aow_storage constructor.

include/cuco/detail/trie/bit_vector/bit_vector_ref.cuh
amukkara: agreed. by using a struct, ranks and words can use different storage types.

we can then delete rank_union struct, whose sole purpose is to be able to use same storage type for words and ranks.

amukkara: relevant changes in c5ec25452ea9a37d355c2eaca787574bea78ec01

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: 4451d7e19a6ee2041a6ae0522877abc474b2d97b

include/cuco/detail/trie/bit_vector/bit_vector_ref.cuh
amukkara: 4451d7e19a6ee2041a6ae0522877abc474b2d97b

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: This struct is not needed after c5ec25452ea9a37d355c2eaca787574bea78ec01

include/cuco/detail/trie/bit_vector/bit_vector.cuh
amukkara: 4451d7e19a6ee2041a6ae0522877abc474b2d97b

include/cuco/detail/trie/bit_vector/bit_vector.inl
amukkara: I prefer having implementations in .inl to avoid very large files (for readability)

include/cuco/detail/trie/bit_vector/bit_vector.cuh
```diff
@@ -0,0 +1,264 @@
...
+  thrust::device_vector<rank> ranks0_;         ///< Same as ranks_ but for `0` bits
+  thrust::device_vector<size_type> selects_;   ///< Block indices of (0, 256, 512...)th `1` bit
+  thrust::device_vector<size_type> selects0_;  ///< Same as selects_, but for `0` bits
+
+  allocator_type allocator_;  ///< Allocator used to (de)allocate temporary storage
+  std::unique_ptr<words_storage_type> aow_words_;  ///< Array of window storage structure
+  std::unique_ptr<ranks_storage_type> aow_ranks_;
+  std::unique_ptr<ranks_storage_type> aow_ranks0_;
+  std::unique_ptr<selects_storage_type> aow_selects_;
+  std::unique_ptr<selects_storage_type> aow_selects0_;
```
PointKernel: Maybe a leftover, one member group of those two should be redundant.

amukkara: We could remove the aow_storage versions. Using thrust::device_vector throughout would be much simpler. Bitvector class is not using any window functionality like say static_set would.

PointKernel: sure, works for me.

include/cuco/detail/trie/bit_vector/bit_vector_ref.cuh
PointKernel: Right, ref dynamic bitset would never "grow" the container itself thus having operator base class doesn't help here. I've moved ref as a nested type of `dynamic_bitset`.

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+  uint32_t abs_hi_;                    ///< Upper 32 bits of base
+  uint8_t abs_lo_;                     ///< Lower 8 bits of base
+  cuda::std::array<uint8_t, 3> rels_;  ///< Offsets for 64-bit sub-intervals
+
+  /**
+   * @brief Gets base rank of current 256-bit interval
+   *
+   * @return The base rank
+   */
+  __host__ __device__ constexpr uint64_t abs() const noexcept
```
PointKernel: Could we find a more descriptive name for this function and also `set_abs`?

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+  {
+    return (static_cast<uint64_t>(abs_hi_) << 8) | abs_lo_;
```
PointKernel:
```suggestion
    return (static_cast<uint64_t>(abs_hi_) << CHAR_BIT) | abs_lo_;
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+  /**
+   * @brief Sets base rank of current 256-bit interval
+   *
+   * @param abs Base rank
+   */
+  __host__ __device__ constexpr void set_abs(uint64_t abs) noexcept
+  {
+    abs_hi_ = static_cast<uint32_t>(abs >> 8);
```
PointKernel:
```suggestion
    abs_hi_ = static_cast<uint32_t>(abs >> CHAR_BIT);
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+  using slot_type = uint64_t;     ///< Slot type
+  /// Type of the allocator to (de)allocate words
+  using allocator_type = typename std::allocator_traits<Allocator>::rebind_alloc<slot_type>;
+
+  static constexpr size_type words_per_block = 4;  ///< Tradeoff between space efficiency and perf.
+  static constexpr size_type bits_per_word   = sizeof(slot_type) * CHAR_BIT;     ///< Bits in a word
+  static constexpr size_type bits_per_block  = words_per_block * bits_per_word;  ///< Trivial
+
+  /**
+   * @brief Constructs an empty bitvector
```
PointKernel:
```suggestion
   * @brief Constructs an empty biset
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+  __host__ __device__ constexpr void set_abs(uint64_t abs) noexcept
+  {
+    abs_hi_ = static_cast<uint32_t>(abs >> 8);
+    abs_lo_ = static_cast<uint8_t>(abs);
+  }
+};
+
+/**
+ * @brief Bitvector class with rank and select index structures
+ *
```
PointKernel, line 71:
```suggestion
 *
 * Current limitations:
 * - Stream controls are partially supported due to the use of `thrust::device_vector` as storage
 * - Device ref doesn't support modifiers like `set`, `reset`, etc. 
 *
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+ * In addition to standard bitvector get/set operations, this class provides
+ * rank and select operation API. It maintains index structures to make both these
+ * new operations close to constant time.
+ * Bitvector construction happens on host, after which the structures are moved to device.
+ * All subsequent read-only operations access device structures only.
```
PointKernel:
```suggestion
```
doc to be updated

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+   */
+  __host__ __device__ constexpr void set_abs(uint64_t abs) noexcept
+  {
+    abs_hi_ = static_cast<uint32_t>(abs >> 8);
+    abs_lo_ = static_cast<uint8_t>(abs);
+  }
+};
+
+/**
+ * @brief Bitvector class with rank and select index structures
```
PointKernel:
```suggestion
 * @brief Bitset class with rank and select index structures
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+  constexpr dynamic_bitset(Allocator const& allocator = Allocator{});
+
+  /**
+   * @brief adds a new bit at the end
+   *
+   * Grows internal storage if needed
+   *
+   * @param bit Boolean value of new bit to be added
+   */
+  constexpr void append(bool bit) noexcept;
```
PointKernel:
```suggestion
  /**
   * @brief Appends the given element `value` to the end of the bitset
   *
   * This API may involve data reallocation if the current storage is exhausted.
   *
   * @param value Boolean value of the new bit to be added
   */
  constexpr void push_back(bool value) noexcept;
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * @brief Modifies a single bit
+   *
+   * @param index position of bit to be modified
+   * @param bit new value of bit
+   */
+  constexpr void set(size_type index, bool bit) noexcept;
```
PointKernel:
```suggestion
  /**
   * @brief Sets the target bit indexed by `index` to a specified `value`.
   *
   * @param index Position of bit to be modified
   * @param value New value of the target bit
   */
  constexpr void set(size_type index, bool value) noexcept;
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * @brief Sets last bit to specified value
+   *
+   * @param bit new value of last bit
+   */
+  constexpr void set_last(bool bit) noexcept;
```
PointKernel:
```suggestion
  /**
   * @brief Sets the last bit to a specified value
   *
   * @param value New value of the last bit
   */
  constexpr void set_last(bool value) noexcept;
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * @brief Builds indexes for rank and select
+   */
+  constexpr void build() noexcept;
```
PointKernel: I'm thinking of adding a new data member called
```cuda
  bool is_built_; ///< Flag indicating whether the rank and select indices are built or not
```
We set the flag to `false` in ctor and all modifier functions and at the beginning of query/read functions, we can do something like:
```cuda
  if (not is_built_) {
    this->build();
    is_built_ = true;
  }
```
In this way,  this could be a private function and we can hide the build process from users.

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * @brief Bulk get operation
+   *
+   * @tparam KeyIt Device-accessible iterator to keys
+   * @tparam OutputIt Device-accessible iterator to outputs
```
PointKernel:
```suggestion
   * @tparam KeyIt Device-accessible iterator to keys whose `value_type` can be converted to bitset's `size_type`
   * @tparam OutputIt Device-accessible iterator to outputs whose `value_type` can be constructed from boolean type
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+   */
+  constexpr void set_last(bool bit) noexcept;
+
+  /**
+   * @brief Builds indexes for rank and select
+   */
+  constexpr void build() noexcept;
+
+  /**
+   * @brief Bulk get operation
```
PointKernel:
```suggestion
   * @brief For any element `keys_begin[i]` in the range `[keys_begin, keys_end)`, stores the boolean value at position `keys_begin[i]` to `output_begin[i]`.
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+   * @tparam KeyIt Device-accessible iterator to keys
+   * @tparam OutputIt Device-accessible iterator to outputs
+   *
+   * @param keys_begin Begin iterator to keys list whose values are queried
+   * @param keys_end End iterator to keys list
+   * @param outputs_begin Begin iterator to outputs of get operation
+   * @param stream Stream to execute get kernel
+   */
+  template <typename KeyIt, typename OutputIt>
+  constexpr void get(KeyIt keys_begin,
```
PointKernel:
```suggestion
  constexpr void test(KeyIt keys_begin,
```
`test` instead of `get` to align with STL/boost.

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+                     KeyIt keys_end,
+                     OutputIt outputs_begin,
+                     cuda_stream_ref stream = {}) const noexcept;
+
+  /**
+   * @brief Bulk rank operation
+   *
+   * @tparam KeyIt Device-accessible iterator to keys
+   * @tparam OutputIt Device-accessible iterator to output ranks
```
PointKernel: please be explicit about the iterator's `value_type`

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+  constexpr void ranks(KeyIt keys_begin,
+                       KeyIt keys_end,
+                       OutputIt outputs_begin,
+                       cuda_stream_ref stream = {}) const noexcept;
+
+  /**
+   * @brief Bulk select operation
+   *
+   * @tparam KeyIt Device-accessible iterator to keys
+   * @tparam OutputIt Device-accessible iterator to outputs
```
PointKernel: specify iterator's `value_type`

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+    [[nodiscard]] __device__ constexpr slot_type get_word(size_type word_id) const noexcept;
+
+    /**
+     * @brief Find position of first set bit starting from a given position (inclusive)
+     *
+     * @param key Position of starting bit
+     *
+     * @return Index of next set bit
+     */
+    [[nodiscard]] __device__ size_type find_next_set(size_type key) const noexcept;
```
PointKernel:
```suggestion
    [[nodiscard]] __device__ size_type find_next(size_type key) const noexcept;
```
https://www.boost.org/doc/libs/1_83_0/libs/dynamic_bitset/dynamic_bitset.html#find_next

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+    __host__ __device__ explicit constexpr reference(storage_ref_type storage) noexcept;
+
+    /**
+     * @brief Access value of a single bit
+     *
+     * @param key Position of bit
+     *
+     * @return Value of bit at position specified by key
+     */
+    [[nodiscard]] __device__ constexpr bool get(size_type key) const noexcept;
```
PointKernel:
```suggestion
    [[nodiscard]] __device__ constexpr bool test(size_type key) const noexcept;
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+
+    /**
+     * @brief Access a single word of internal storage
+     *
+     * @param word_id Index of word
+     *
+     * @return Word at position specified by index
+     */
+    [[nodiscard]] __device__ constexpr slot_type get_word(size_type word_id) const noexcept;
```
PointKernel:
```suggestion
    [[nodiscard]] __device__ constexpr slot_type word(size_type word_id) const noexcept;
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+ * All subsequent read-only operations access device structures only.
+ *
+ * @tparam Allocator Type of allocator used for device storage
+ */
+// TODO: have to use device_malloc_allocator for now otherwise the container cannot grow
+template <class Allocator = thrust::device_malloc_allocator<std::byte>>
+class dynamic_bitset {
+ public:
+  using size_type = std::size_t;  ///< size type to specify bit index
+  using slot_type = uint64_t;     ///< Slot type
```
PointKernel: Question: Should this be just `word_type`?

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+   * @tparam KeyIt Device-accessible iterator to keys
+   * @tparam OutputIt Device-accessible iterator to output ranks
+   *
+   * @param keys_begin Begin iterator to keys list whose ranks are queried
+   * @param keys_end End iterator to keys list
+   * @param outputs_begin Begin iterator to outputs ranks list
+   * @param stream Stream to execute ranks kernel
+   */
+  template <typename KeyIt, typename OutputIt>
+  constexpr void ranks(KeyIt keys_begin,
```
PointKernel:
```suggestion
  constexpr void rank(KeyIt keys_begin,
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+   * @tparam KeyIt Device-accessible iterator to keys
+   * @tparam OutputIt Device-accessible iterator to outputs
+   *
+   * @param keys_begin Begin iterator to keys list whose select values are queried
+   * @param keys_end End iterator to keys list
+   * @param outputs_begin Begin iterator to outputs selects list
+   * @param stream Stream to execute selects kernel
+   */
+  template <typename KeyIt, typename OutputIt>
+  constexpr void selects(KeyIt keys_begin,
```
PointKernel:
```suggestion
  constexpr void select(KeyIt keys_begin,
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+    [[nodiscard]] __device__ constexpr size_type select(size_type count) const noexcept;
+
+    /**
+     * @brief Find position of Nth not-set (0) bit counting from start of bitvector
+     *
+     * @param count Input N
+     *
+     * @return Position of Nth not-set bit
+     */
+    [[nodiscard]] __device__ constexpr size_type select0(size_type count) const noexcept;
```
PointKernel:
```suggestion
    [[nodiscard]] __device__ constexpr size_type select_false(size_type count) const noexcept;
```
`select_false` instead of `select0` based on the offline discussion

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,359 @@
...
+     * @brief Helper function for select operation that computes an initial rank estimate
+     *
+     * @param count Input count for which select operation is being performed
+     * @param selects Selects array
+     * @param ranks Ranks array
+     *
+     * @return index in ranks which corresponds to highest rank less than count (least upper bound)
+     */
+    template <typename SelectsRef, typename RanksRef>
+    [[nodiscard]] __device__ constexpr size_type get_initial_rank_estimate(
```
PointKernel:
```suggestion
    [[nodiscard]] __device__ constexpr size_type initial_rank_estimate(
```
`get_` prefixes rarely provide useful information so let's just remove it.

include/cuco/detail/trie/dynamic_bitset/kernels.cuh
```diff
@@ -0,0 +1,234 @@
...
+ * @param num_keys Number of input keys
+ */
+template <typename BitvectorRef, typename KeyIt, typename ValueIt, typename size_type>
+__global__ void bitvector_get_kernel(BitvectorRef ref,
+                                     KeyIt keys,
+                                     ValueIt outputs,
+                                     size_type num_keys)
+{
+  uint32_t const loop_stride = gridDim.x * blockDim.x;
+  uint32_t key_id            = blockDim.x * blockIdx.x + threadIdx.x;
```
PointKernel:
```suggestion
  cuco::detail::index_type const loop_stride = gridDim.x * blockDim.x;
  cuco::detail::index_type key_id            = blockDim.x * blockIdx.x + threadIdx.x;
```
Similar to
https://github.com/NVIDIA/cuCollections/blob/c0cce866368396ced18a37a0d30287729d6af6c4/include/cuco/detail/common_kernels.cuh#L74-L75, please use `cuco::detail::index_type` consistently to avoid potential overflow issue. The same to other kernels.

include/cuco/detail/trie/dynamic_bitset/kernels.cuh
```diff
@@ -0,0 +1,234 @@
...
+ * @tparam KeyIt Device-accessible iterator to input keys
+ * @tparam ValueIt Device-accessible iterator to values
+ * @tparam size_type Size type
+ *
+ * @param ref Bitvector ref
+ * @param keys Begin iterator to keys
+ * @param outputs Begin iterator to outputs
+ * @param num_keys Number of input keys
+ */
+template <typename BitvectorRef, typename KeyIt, typename ValueIt, typename size_type>
```
PointKernel:
```suggestion
template <typename BitsetRef, typename KeyIt, typename ValueIt, typename size_type>
```

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: How about replacing 'abs' with 'base'?

`rank` struct encodes a list of 4 integers say [1000, 1005, 1006, 1009] as [1000, 5, 6, 9]

what is the standard terminology for such encoding?

PointKernel: Yeah, `base` sounds good.

amukkara: a85b7b70eb22f7c1a03279c4a8674e8c949a6aba

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: a85b7b70eb22f7c1a03279c4a8674e8c949a6aba

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: ec6bb1117ed9e7d5489763549a0de818651e6c03

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: a85b7b70eb22

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: ec6bb1117ed9e7d5489763549a0de818651e6c03

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: ec6bb1117ed9e7d5489763549a0de818651e6c03

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: ec6bb1117ed9e7d5489763549a0de818651e6c03

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 59457e91c5eba5f93d45563c54336a9282f26b2b

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: c7faed0992d792f37919b8749949d0d3e7cf88ee

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 59457e91c5eba5f93d45563c54336a9282f26b2b

include/cuco/detail/trie/dynamic_bitset/kernels.cuh
amukkara: ec6bb1117ed9e7d5489763549a0de818651e6c03

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: c753f1082bf34b074023d6c80fe1e6ad59d6d075

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: c753f1082bf34b074023d6c80fe1e6ad59d6d075

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: Yes, slot_type is a vestige from aow_storage usage

include/cuco/detail/trie/dynamic_bitset/kernels.cuh
amukkara: b26f326b6929531b6517305031e8c816402fd05a

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: f96468df7b03a5ad8dd40eba9463f74fceb09142

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: f96468df7b03a5ad8dd40eba9463f74fceb09142

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: f96468df7b03a5ad8dd40eba9463f74fceb09142

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 0f1db0fae4cdedfcf7605f996400c647123ea13b

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 0f1db0fae4cdedfcf7605f996400c647123ea13b

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 0f1db0fae4cdedfcf7605f996400c647123ea13b

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 63ed55224354aacccde9054fc3ab698e91186b22

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 81b8e907e0acc19a5eabfaeeb927dd39db103739

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 81b8e907e0acc19a5eabfaeeb927dd39db103739

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: 81b8e907e0acc19a5eabfaeeb927dd39db103739

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: b2c88ded9df8bfbd45a98be9f88c3c9e51bcaffc

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -0,0 +1,377 @@
...
+
+  /**
+   * @brief Helper function to calculate grid size for simple kernels
+   *
+   * @param num_elements Elements being processed by kernel
+   *
+   * @return grid size
+   */
+  // TODO: to be moved to the CUDA utility header
+  constexpr size_type default_grid_size(size_type num_elements) const noexcept;
```
PointKernel: To be removed once #359 gets merged.

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
amukkara: This makes sense. We need to do this check within every bulk query operation and ref(). I will take a stab at making the change.

PointKernel: Unless it's needed by trie, let's just focus on bulk operations for now since such a flag in ref APIs would cause host/device synchronization issues that we tend to ignore for now.

amukkara: a629730842d08d41696b

sleeepyjack (approved):
LGTM

PR #353, opened 2023-08-18
Add `insert_or_assign` for map

### LINKED ISSUE #317: [ENHANCEMENT]: Add `insert_or_assign` for `static_map`

Depends on #314

Similar to `std::unordered_map::insert_or_assign`, we should add a new host function `static_map::insert_or_assign` that either inserts the kay-value pair or modifies the existing key with the given payload.

API reference: https://en.cppreference.com/w/cpp/container/unordered_map/insert_or_assign

_No response_

### Additional context

_No response_

PointKernel:
Closes #317

This PR adds `static_map::insert_or_assign`.

It also removes the `[[nodiscard]]` qualifier for `retrieve_all` since they shouldn't be necessary.

### DIFF
--- a/include/cuco/detail/static_map/static_map_ref.inl
+++ b/include/cuco/detail/static_map/static_map_ref.inl
@@ -320,11 +320,7 @@ class operator_impl<
         return detail::window_results{detail::equal_result::UNEQUAL, -1};
       }();
 
-      // If the key is already in the container, return false
-      if (group.any(state == detail::equal_result::EQUAL)) { return; }
-
       auto const group_contains_empty = group.ballot(state == detail::equal_result::EMPTY);
-
       if (group_contains_empty) {
         auto const src_lane = __ffs(group_contains_empty) - 1;
         auto const status =
@@ -334,6 +330,7 @@ class operator_impl<
                 value)
             : false;
 
+        // Exit if inserted or assigned
         if (group.shfl(status, src_lane)) { return; }
       } else {
         ++probing_iter;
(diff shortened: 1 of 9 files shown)


PointKernel:
@seunghwak Let me know if the current implementation can meet your needs.

include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -230,6 +230,168 @@ class operator_impl<
...
+  __device__ constexpr bool attempt_insert_or_assign(value_type* slot,
+                                                     value_type const& value) noexcept
+  {
+    ref_type& ref_          = static_cast<ref_type&>(*this);
+    auto const expected_key = ref_.impl_.empty_slot_sentinel().first;
+
+    auto old_key      = ref_.impl_.compare_and_swap(&slot->first, expected_key, value.first);
+    auto* old_key_ptr = reinterpret_cast<key_type*>(&old_key);
+
+    // if key success or key was alreay present in the map
```
sleeepyjack:
```suggestion
    // if key success or key was already present in the map
```

sleeepyjack (approved):
Looks good!

I have one question/concern: What happens if we concurrently issue a `insert_or_assign` and `erase` operation on the same key? Can the slot end up in an invalid state, e.g. `{erased_key_sentinel, updated_value}`?

PointKernel:
> What happens if we concurrently issue a insert_or_assign and erase operation on the same key?

Is that even legal? :)

Need to check erase detail but IIRC it will reset payload value during erasure so the invalid state you mentioned above may never happen.

sleeepyjack:
We could make it illegal to construct a device ref both with erase and insert_or_assign. That might be the safest solution. I can't even come up with a usecase for that combination.

PointKernel:
> We could make it illegal to construct a device ref both with erase and insert_or_assign.

I don't worry much here but would be eager to learn the use case if a user reports a bug about this scenario.

PR #356, opened 2023-08-21
Distinct code path for key equality checks in `packed_cas`

PointKernel:
This PR fixes a bug where we previously used the whole slot content for map key comparisons in `packed_cas`.


include/cuco/detail/open_addressing_ref_impl.cuh
```diff
@@ -140,7 +141,7 @@ class open_addressing_ref_impl {
    *
    * @return True if the given element is successfully inserted
    */
-  template <typename Predicate>
+  template <bool IsSet, typename Predicate>
```
sleeepyjack: `IsSet` sounds a bit confusing. How about `HasPayload`?

include/cuco/detail/open_addressing_ref_impl.cuh
```diff
@@ -657,20 +663,39 @@ class open_addressing_ref_impl {
...
-               ? insert_result::DUPLICATE
-               : insert_result::CONTINUE;
+      auto const res = [&]() {
+        if constexpr (IsSet) {
+          // If it's a set implementation, compare the whole slot content
+          return predicate.equal_to(*old_ptr, value);
+        }
+        if constexpr (not IsSet) {
+          // If it's a map implementation, compare keys only
+          return predicate.equal_to(old_ptr->first, value.first);
```
sleeepyjack: The return type in both cases is the same, so I think we can just use `if constexpr {...} else {...}`.

include/cuco/detail/open_addressing_ref_impl.cuh
PointKernel: I like `HasPayload`. Thanks!

sleeepyjack (approved):
LGTM
Thanks!

PR #359, opened 2023-08-29
Clean up detail utility functions

PointKernel:
This PR

- adds `grid_stride` and `global_thread_id` utilities in `detail/utility/cuda.cuh` (inspired by similar work in libcudf)
- adds `compute_grid_size` helper function in `detail/utility/cuda.hpp`
- renames `SDIV` as `ceiling_div` and moves it to `detail/utility/math.hpp`

### DIFF
--- a/benchmarks/hash_bench.cu
+++ b/benchmarks/hash_bench.cu
@@ -16,7 +16,6 @@
 
 #include <defaults.hpp>
 
-#include <cuco/detail/utility/math.hpp>
 #include <cuco/hash_functions.cuh>
 
 #include <nvbench/nvbench.cuh>
@@ -70,7 +69,7 @@ void hash_eval(nvbench::state& state, nvbench::type_list<Hash>)
   bool const materialize_result = false;
   constexpr auto block_size     = 128;
   auto const num_keys  = state.get_int64_or_default("NumInputs", cuco::benchmark::defaults::N * 10);
-  auto const grid_size = cuco::detail::ceiling_div(num_keys, block_size * 16);
+  auto const grid_size = (num_keys + block_size * 16 - 1) / block_size * 16;
 
   thrust::device_vector<typename Hash::result_type> hash_values((materialize_result) ? num_keys
                                                                                      : 1);
--- a/benchmarks/hash_bench.cu
+++ b/benchmarks/hash_bench.cu
@@ -16,7 +16,7 @@
 
 #include <defaults.hpp>
 
-#include <cuco/detail/utils.hpp>
+#include <cuco/detail/utility/math.hpp>
 #include <cuco/hash_functions.cuh>
 
 #include <nvbench/nvbench.cuh>
@@ -70,7 +70,7 @@ void hash_eval(nvbench::state& state, nvbench::type_list<Hash>)
   bool const materialize_result = false;
   constexpr auto block_size     = 128;
   auto const num_keys  = state.get_int64_or_default("NumInputs", cuco::benchmark::defaults::N * 10);
-  auto const grid_size = SDIV(num_keys, block_size * 16);
+  auto const grid_size = cuco::detail::ceiling_div(num_keys, block_size * 16);
 
   thrust::device_vector<typename Hash::result_type> hash_values((materialize_result) ? num_keys
                                                                                      : 1);
--- a/include/cuco/detail/extent/extent.inl
+++ b/include/cuco/detail/extent/extent.inl
@@ -81,7 +81,7 @@ template <int32_t CGSize, int32_t WindowSize, typename SizeType, std::size_t N>
     (static_cast<uint64_t>(std::numeric_limits<SizeType>::max()) < max_prime)
       ? std::numeric_limits<SizeType>::max()
       : static_cast<SizeType>(max_prime);
-  auto const size = cuco::detail::ceiling_div(
+  auto const size = cuco::detail::int_div_ceil(
     std::max(static_cast<SizeType>(ext), static_cast<SizeType>(1)), CGSize * WindowSize);
   if (size > max_value) { CUCO_FAIL("Invalid input extent"); }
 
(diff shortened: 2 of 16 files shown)


benchmarks/hash_bench.cu
```diff
@@ -70,7 +70,7 @@ void hash_eval(nvbench::state& state, nvbench::type_list<Hash>)
   bool const materialize_result = false;
   constexpr auto block_size     = 128;
   auto const num_keys  = state.get_int64_or_default("NumInputs", cuco::benchmark::defaults::N * 10);
-  auto const grid_size = SDIV(num_keys, block_size * 16);
+  auto const grid_size = cuco::detail::ceiling_div(num_keys, block_size * 16);
```
sleeepyjack: Since benchmarks are considered user-facing code, we shouldn't use `detail` namespace here.

include/cuco/detail/extent/extent.inl
```diff
@@ -80,8 +81,8 @@ template <int32_t CGSize, int32_t WindowSize, typename SizeType, std::size_t N>
     (static_cast<uint64_t>(std::numeric_limits<SizeType>::max()) < max_prime)
       ? std::numeric_limits<SizeType>::max()
       : static_cast<SizeType>(max_prime);
-  auto const size =
-    SDIV(std::max(static_cast<SizeType>(ext), static_cast<SizeType>(1)), CGSize * WindowSize);
+  auto const size = cuco::detail::ceiling_div(
```
sleeepyjack: Should we go with the stdlib naming scheme like `std::bit_ceil`?
Maybe `div_ceil` or `int_div_ceil`?

include/cuco/detail/utility/cuda.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include <cuco/detail/utility/math.hpp>
+
+namespace cuco {
+namespace detail {
+
+using index_type = int64_t;  ///< CUDA thread index type
+
+static constexpr int32_t CUCO_DEFAULT_BLOCK_SIZE = 128;  ///< Default block size
+static constexpr int32_t CUCO_DEFAULT_STRIDE     = 1;    ///< Default stride
```
sleeepyjack:
```suggestion
constexpr int32_t default_block_size() noexcept { return 128; }  ///< Default block size
constexpr int32_t default_stride() noexcept { return 1; }        ///< Default stride
```

The all-caps + `CUCO_` prefix naming scheme only applies to macros.
I tend to use a `constexpr` function over a variable here.
If we go with the variable instead, we should make them `inline constexpr` so all TUs see the same definition.
Not that for the function approach, `inline` is already implied.

include/cuco/detail/utility/cuda.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * @brief Computes the desired 1D grid size with the given parameters
+ *
+ * @param num Number of elements to handle in the kernel
+ * @param cg_size Number of threads per CUDA Cooperative Group
+ * @param stride Number of elements to be handled by each thread
+ * @param block_size Number of threads in each thread block
+ *
+ * @return The resulting grid size
+ */
+constexpr auto compute_grid_size(index_type num,
```
sleeepyjack: Maybe just `grid_size`?

include/cuco/detail/utility/cuda.hpp
```diff
@@ -0,0 +1,47 @@
...
+                                 int32_t cg_size    = 1,
+                                 int32_t stride     = CUCO_DEFAULT_STRIDE,
+                                 int32_t block_size = CUCO_DEFAULT_BLOCK_SIZE)
```
sleeepyjack:
```suggestion
                                 int32_t block_size = CUCO_DEFAULT_BLOCK_SIZE) noexcept
```

include/cuco/detail/utility/cuda.cuh
```diff
@@ -0,0 +1,50 @@
...
+*
+* @tparam CGSize Number of threads in each CUDA Cooperative Group
+*
+* @return The grid stride
+￼   */
+template <index_type CGSize = 1>
+__device__ static constexpr index_type grid_stride() noexcept
+{
+  return index_type{gridDim.x} * index_type{blockDim.x} / CGSize;
+}
```
sleeepyjack: Can we do the `.../ CGSize` part at the call site instead? It doesn't feel right to me for that to be part of these functions. The function names and docs doesn't reflect this optional behavior, and there's no performance benefit.

include/cuco/detail/static_multimap/static_multimap.inl
```diff
@@ -336,7 +326,6 @@ OutputIt static_multimap<Key, Value, Scope, Allocator, ProbeSequence>::retrieve_
 
   // Using per-warp buffer for vector loads and per-CG buffer for scalar loads
   constexpr auto buffer_size = uses_vector_load() ? (warp_size() * 3u) : (cg_size() * 3u);
```
sleeepyjack, line 327:
```suggestion
  constexpr auto buffer_size = this->uses_vector_load() ? (this->warp_size() * 3u) : (this->cg_size() * 3u);
```

~~I'm seeing this now but it's no subject for this PR. We can defer this to a separate cleanup PR.~~
Forget what I just said. I just realized this is the old multimap implementation. No need to fix it then.

sleeepyjack: Also, `warp_size()` feels like something that should be defined in `cuco/detail/utility/cuda.hpp`

include/cuco/detail/utility/cuda.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace cuco {
+namespace detail {
+
+using index_type = int64_t;  ///< CUDA thread index type
+
+/// Default block size
+static constexpr int32_t default_block_size() noexcept { return 128; }
+/// Default stride
+static constexpr int32_t default_stride() noexcept { return 1; }
```
sleeepyjack:
```suggestion
/// Default block size
constexpr int32_t default_block_size() noexcept { return 128; }
/// Default stride
constexpr int32_t default_stride() noexcept { return 1; }
```

For a constexpr function, giving it internal linkage via static is redundant.

sleeepyjack (changes requested):

sleeepyjack (approved):

PR #362, opened 2023-08-31
Fix hash to size conversion

sleeepyjack:
This PR fixes some issues in the helper function for converting a hash value into a valid size.

Hash values are always unsigned, but the size type is set by the user and can thus be any integer type.
The problem occurs when the size type is signed and the hash value exceeds the range of the size type (see [repro](https://godbolt.org/z/a9qnaPbeo)).

The proposed solution fixes this: https://godbolt.org/z/46fqEaaEx
The `thread_rank` parameter can be defaulted for non-CG probing, and I also double-checked that the compiler is able to use that information to apply optimizations: https://godbolt.org/z/x3b6c1aMh


PointKernel:
Can you please add a test exercising this?

PR #365, opened 2023-09-05
Add erase to new map and set

PointKernel:
This PR adds `erase` functionality in the new `static_set` and `static_map`.

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -640,6 +640,16 @@ class open_addressing_impl {
   }
 
   /**
+   * @brief Gets the sentinel value used to represent an empty key slot.
+   *
+   * @return The sentinel value used to represent an empty key slot
+   */
+  [[nodiscard]] constexpr key_type erased_key_sentinel() const noexcept
+  {
+    return erased_key_sentinel_;
+  }
+
+  /**
    * @brief Gets the key comparator.
    *
    * @return The comparator used to compare keys
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -641,9 +641,9 @@ class open_addressing_impl {
   }
 
   /**
-   * @brief Gets the sentinel value used to represent an empty key slot.
+   * @brief Gets the sentinel value used to represent an erased key slot.
    *
-   * @return The sentinel value used to represent an empty key slot
+   * @return The sentinel value used to represent an erased key slot
    */
   [[nodiscard]] constexpr key_type erased_key_sentinel() const noexcept
   {
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -413,16 +413,13 @@ class open_addressing_impl {
   /**
    * @brief Asynchronously erases keys in the range `[first, last)`.
    *
-   * For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
+   * @note For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
    * associated value from the container. Else, no effect.
-   *
-   *  Side-effects:
+   * @note Side-effects:
    *  - `contains(k) == false`
    *  - `find(k) == end()`
    *  - `insert({k,v}) == true`
    *  - `size()` is reduced by the total number of erased keys
-   *
-   * This function synchronizes `stream`.
    *
    * @tparam InputIt Device accessible input iterator whose `value_type` is
    * convertible to the container's `key_type`
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -415,6 +415,7 @@ class open_addressing_impl {
    *
    * @note For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
    * associated value from the container. Else, no effect.
+   *
    * @note Side-effects:
    *  - `contains(k) == false`
    *  - `find(k) == end()`
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -34,6 +34,8 @@
 
 #include <cuda/atomic>
 
+#include <cmath>
+
 namespace cuco {
 namespace experimental {
 namespace detail {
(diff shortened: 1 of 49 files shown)


include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -365,6 +410,48 @@ class open_addressing_impl {
...
+   * For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
+   * associated value from the container. Else, no effect.
+   *
+   *  Side-effects:
+   *  - `contains(k) == false`
+   *  - `find(k) == end()`
+   *  - `insert({k,v}) == true`
+   *  - `size()` is reduced by the total number of erased keys
+   *
+   * This function synchronizes `stream`.
```
sleeepyjack: Should these be `@note`?

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -365,6 +410,48 @@ class open_addressing_impl {
...
+   *
+   * @throw std::runtime_error if a unique erased key sentinel value was not
+   * provided at construction
+   */
+  template <typename InputIt, typename Ref>
+  void erase_async(InputIt first, InputIt last, Ref container_ref, cuda_stream_ref stream = {})
+  {
+    CUCO_EXPECTS(empty_key_sentinel_ != erased_key_sentinel_,
+                 "The empty key sentinel and erased key sentinel cannot be the same value.",
+                 std::logic_error);
```
sleeepyjack, line 441: So the reason this function can't be `noexcept` is this check. Do we need it here or can we move it to the ctor overload taking a `erased_key_sentinel`?

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -556,6 +643,16 @@ class open_addressing_impl {
     return empty_key_sentinel_;
   }
 
+  /**
+   * @brief Gets the sentinel value used to represent an empty key slot.
+   *
+   * @return The sentinel value used to represent an empty key slot
```
sleeepyjack:
```suggestion
   * @return The sentinel value used to represent an erased key slot
```

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -556,6 +643,16 @@ class open_addressing_impl {
     return empty_key_sentinel_;
   }
 
+  /**
+   * @brief Gets the sentinel value used to represent an empty key slot.
```
sleeepyjack:
```suggestion
   * @brief Gets the sentinel value used to represent an erased key slot.
```

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -130,7 +130,30 @@ class open_addressing_ref_impl {
...
+   *
+   * @param empty_slot_sentinel Sentinel indicating an empty slot
+   * @param erased_key_sentinel Sentinel indicating an erased key
+   * @param predicate Key equality binary callable
+   * @param probing_scheme Probing scheme
+   * @param storage_ref Non-owning ref of slot storage
+   */
+  __host__ __device__ explicit constexpr open_addressing_ref_impl(
+    value_type empty_slot_sentinel,
+    key_type erased_key_sentinel,
```
sleeepyjack, line 151: Having a dedicated struct for all the sentinels will clean this up. Let's make it a future PR.

include/cuco/static_map.cuh
```diff
@@ -367,6 +399,60 @@ class static_map {
...
+   * For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
+   * associated value from the map. Else, no effect.
+   *
+   *  Side-effects:
+   *  - `contains(k) == false`
+   *  - `find(k) == end()`
+   *  - `insert({k,v}) == true`
+   *  - `size()` is reduced by the total number of erased keys
+   *
+   * This function synchronizes `stream`.
```
sleeepyjack: Should use `@note`?
BTW I love the side-effects section!

include/cuco/static_map.cuh
```diff
@@ -367,6 +399,60 @@ class static_map {
...
```
sleeepyjack: Sam here: `@note`

sleeepyjack (commented):
First round of review. Nice work!

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -365,6 +410,48 @@ class open_addressing_impl {
...
+   * @brief Asynchronously erases keys in the range `[first, last)`.
+   *
+   * For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
+   * associated value from the container. Else, no effect.
+   *
+   *  Side-effects:
+   *  - `contains(k) == false`
+   *  - `find(k) == end()`
+   *  - `insert({k,v}) == true`
+   *  - `size()` is reduced by the total number of erased keys
```
PointKernel: @sleeepyjack how to properly "note" in this case?

include/cuco/detail/open_addressing/open_addressing_impl.cuh
PointKernel: We are doing this check in both `erase` and ctor to avoid misuse.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
sleeepyjack: Like this? https://stackoverflow.com/a/34016569

No clue how it looks after rendering.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
PointKernel: So we currently only apply the equal wrapper in container ref but not the container itself. Let me add a TODO here

PointKernel: [d2de184](https://github.com/NVIDIA/cuCollections/pull/365/commits/d2de184f27b9980c77af60e6691ebbcbda997255)

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -365,6 +410,45 @@ class open_addressing_impl {
...
+  /**
+   * @brief Asynchronously erases keys in the range `[first, last)`.
+   *
+   * @note For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
+   * associated value from the container. Else, no effect.
+   * @note Side-effects:
+   *  - `contains(k) == false`
+   *  - `find(k) == end()`
+   *  - `insert({k,v}) == true`
+   *  - `size()` is reduced by the total number of erased keys
```
sleeepyjack:
```suggestion
   * @note For each key `k` in `[first, last)`, if `contains(k)` returns true, removes `k` and it's
   * associated value from the container. Else, no effect.
   *
   * @note Side-effects:
   *  - `contains(k) == false`
   *  - `find(k) == end()`
   *  - `insert({k,v}) == true`
   *  - `size()` is reduced by the total number of erased keys
```

IIUC the doxygen docs correctly, `@note` will emit a new paragraph which is terminated by an empty line at the end.

include/cuco/detail/static_map/functors.cuh
```diff
@@ -84,7 +89,8 @@ struct slot_is_filled {
   template <typename Slot>
   __device__ constexpr bool operator()(Slot const& slot) const noexcept
   {
-    return not cuco::detail::bitwise_compare(empty_sentinel_, thrust::get<0>(slot));
+    return not(cuco::detail::bitwise_compare(empty_sentinel_, thrust::get<0>(slot)) or
+               cuco::detail::bitwise_compare(erased_sentinel_, thrust::get<0>(slot)));
```
sleeepyjack, line 93: Unrelated to the PR but FYI: We're using `thrust::get` here which is the right choice. In other places (including my recent PRs), we use `.first`/`.second` which is less generic. We should put this on our TODO list for minor cleanups.

include/cuco/static_set.cuh
```diff
@@ -298,6 +328,56 @@ class static_set {
...
+   * @brief Erases keys in the range `[first, last)`.
+   *
+   * @note For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
+   * associated value from the container. Else, no effect.
+   * @note This function synchronizes `stream`.
+   * @note Side-effects:
+   *  - `contains(k) == false`
+   *  - `find(k) == end()`
+   *  - `insert({k,v}) == true`
+   *  - `size()` is reduced by the total number of erased keys
```
sleeepyjack:
```suggestion
   * @note For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
   * associated value from the container. Else, no effect.
   *
   * @note This function synchronizes `stream`.
   *
   * @note Side-effects:
   *  - `contains(k) == false`
   *  - `find(k) == end()`
   *  - `insert({k,v}) == true`
   *  - `size()` is reduced by the total number of erased keys
```

include/cuco/static_set.cuh
```diff
@@ -298,6 +328,56 @@ class static_set {
...
+  /**
+   * @brief Asynchronously erases keys in the range `[first, last)`.
+   *
+   * @note For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
+   * associated value from the container. Else, no effect.
+   * @note Side-effects:
+   *  - `contains(k) == false`
+   *  - `find(k) == end()`
+   *  - `insert({k,v}) == true`
+   *  - `size()` is reduced by the total number of erased keys
```
sleeepyjack:
```suggestion
   * @note For each key `k` in `[first, last)`, if contains(k) returns true, removes `k` and it's
   * associated value from the container. Else, no effect.
   *
   * @note Side-effects:
   *  - `contains(k) == false`
   *  - `find(k) == end()`
   *  - `insert({k,v}) == true`
   *  - `size()` is reduced by the total number of erased keys
```

sleeepyjack (commented):
Last round. Only a few minor doc changes. LGTM

~~Can you also enable the commented-out section in the `tests/static_set/heterogeneous_lookup_test.cu`?
You can basically erase half of the keys, then rehash, and check if the size after the rehash still matches.~~
Edit: I got confused there...the test is in the rehash PR which is still open. No action item then.

sleeepyjack (approved):
LGTM 🚀

PR #374, opened 2023-09-27
Index and key

### LINKED ISSUE #373: [ENHANCEMENT]: Allow cuco to accept keys that are not the same values

The key that is being inserted does not necessarily have to have the same type as the values in the slots.  It just has to be castable to the type that is in the slot.

Modify templates so that you can insert any type that you want, so long as it can be cast to the slot type.

_No response_

### Additional context

Test case: https://godbolt.org/z/E338375bf

esoha-nvidia:
This fixes #373


esoha-nvidia:
@sleeepyjack Please take a look.

PR #377, opened 2023-10-02
Remove HasPayload tparam from OA impl classes

sleeepyjack:
This PR removes the `HasPayload` tparam from the member functions of the `open_addressing_impl` classes and instead dispatches the correct behavior implicitly. This PR additionally moves the `open_addressing_impl` headers to their own subdirectory.


sleeepyjack:
Based on offline discussion, we decided to split this work into multiple PRs.

PointKernel (approved):
Nice cleanups. Thanks!

PR #384, opened 2023-10-14
Add static_map::insert_or_apply aka reduce-by-key

sleeepyjack:
This PR adds a function `static_map::insert_or_apply`, which either inserts a new key-value pair into the map, or, in case the key already exists, applies a reduction function over the associated value.

This effectively replaces the former `cuco::static_reduction_map` and thus superseeds #98.

Closes #82


PointKernel:
Closing this as it's fixed by #515

PR #387, opened 2023-10-17
Fix problems with heterogeneous insert input types

sleeepyjack:
This PR fixes some of the problems with heterogeneous insert for `static_map`.
The input type to any of the `insert(_*)` operators is now `cuco::pair<ProbeKey, mapped_type>`.

Also updates docs for both map and set regarding type requirements of `ProbeKey`.

Also adds a small convenient abstraction aka `attempt_insert_stable`.


include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -248,7 +250,7 @@ class operator_impl<
     static_assert(cg_size == 1, "Non-CG operation is incompatible with the current probing scheme");
 
     ref_type& ref_       = static_cast<ref_type&>(*this);
-    auto const key       = value.first;
+    auto const key       = thrust::get<0>(thrust::raw_reference_cast(value));
```
PointKernel: can you please remind me why `thrust::get` is preferred?

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -1054,6 +1044,35 @@ class open_addressing_ref_impl {
...
   }
 
+  /**
+   * @brief Attempts to insert an element into a slot.
+   *
+   * @note Dispatches the correct implementation depending on the container
+   * type and presence of other operator mixins.
+   *
+   * @note `stable` here means that the payload will only be updated once from the sentinel value to
+   * the payload value
```
PointKernel: I'm confused by this note.

Maybe "from the sentinel value to the **desired** value"?

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: This is a leftover from the heterogeneous insert PR. The input type can now be any _tuple-like_ type that is convertible to the map's `value_type`. The `.first` and `.second` members are only defined for _pair-like_ types, so I had to use `get` here instead.

sleeepyjack: I went with the following convention: Use `.first/.second` where possible to indicate that the expected type is a `cuco::pair` - just for sanity checking, and only use `get` when needed, i.e., when dealing with a user-defined tuple-like type.

include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -196,16 +199,16 @@ class operator_impl<
...
    * @param group The Cooperative Group used to perform group insert
    * @param value The element to insert
    *
    * @return True if the given element is successfully inserted
    */
-  template <typename Value>
+  template <typename ProbeKey>
   __device__ bool insert(cooperative_groups::thread_block_tile<cg_size> const& group,
-                         Value const& value) noexcept
+                         cuco::pair<ProbeKey, mapped_type> const& value) noexcept
```
sleeepyjack: I went with this version since it requires less changes and is more sane.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -920,7 +928,7 @@ class open_addressing_ref_impl {
                                                               value_type const& expected,
                                                               Value const& desired) noexcept
   {
-    auto old      = compare_and_swap(address, expected, static_cast<value_type>(desired));
+    auto old      = compare_and_swap(address, expected, this->native_value(desired));
```
PointKernel, line 1012: `native_value` is used only once. Either it's not needed or it should be applied in other places as well.

include/cuco/detail/static_map/static_map_ref.inl
PointKernel: why is `ProbeKey` being convertible to `key_type` while `payload` has to be `mapped_type`

include/cuco/detail/traits.hpp
```diff
@@ -30,10 +29,11 @@ struct is_std_pair_like : cuda::std::false_type {
 
 template <typename T>
 struct is_std_pair_like<T,
-                        cuda::std::void_t<decltype(std::get<0>(cuda::std::declval<T>())),
-                                          decltype(std::get<1>(cuda::std::declval<T>()))>>
-  : cuda::std::
-      conditional_t<std::tuple_size<T>::value == 2, cuda::std::true_type, cuda::std::false_type> {
+                        cuda::std::void_t<decltype(cuda::std::get<0>(cuda::std::declval<T>())),
+                                          decltype(cuda::std::get<1>(cuda::std::declval<T>()))>>
```
PointKernel: Needs further check. This changes the trait behavior from std-like to libcudacxx cuda-std-like.

PointKernel (changes requested):

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
sleeepyjack: I'd say yes, but I think I will need that function for `insert_or_apply` soon. We can either introduce the function in this PR or the `insert_or_apply` PR.

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: The "heterogeneous" part only affects the key, i.e., when introducing a custom `key_equal` functor or key `hash`. So we only need the tparam for the custom key type, while the payload can be the same as the map's `mapped_type`

include/cuco/detail/traits.hpp
sleeepyjack: I thought the `cuda::std` version is a drop-in replacement for the host-side `std` implementation? Do we still need to support `std` namespace in this case?

include/cuco/detail/static_map/static_map_ref.inl
PointKernel: My point is that as long as the insert pair is convertible to the map's `value_type`, it should be fine. The payload can be any arbitrary types that can be convertible to `mapped_type`.

include/cuco/detail/traits.hpp
PointKernel: `cuda::std::pair` is not the same as `std::pair`. By doing this, we change the semantics of this trait.

The initial idea was to help effortless conversion from std pair to cuco pair. The proper way would be adding another trait like `is_cuda_std_pair_like` to apply the code you suggested.

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: Gotcha. Makes sense. Let me try out this change and see if it works.

include/cuco/detail/traits.hpp
sleeepyjack: You're right: https://godbolt.org/z/xx53ozdPE

Does that mean we should also need a `cuco::pair` ctor specialized for `is_cuda_std_pair_like`? I'm confused.

PointKernel: > Does that mean we should also need a cuco::pair ctor specialized for is_cuda_std_pair_like? 

That would be useful. We can do this in a separate PR.

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: IIRC based on offline discussion we decided to leave it as is. Can you confirm?

include/cuco/detail/traits.hpp
sleeepyjack: Already pushed the change when I saw your comment 🙈

sleeepyjack:
I fucked up the merge…investigating

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -938,7 +929,57 @@ class open_addressing_ref_impl {
...
+   * @brief Converts the given type to the container's native `value_type`.
+   *
+   * @tparam T Input type which is convertible to 'value_type'
+   *
+   * @param value The input value
+   *
+   * @return The converted object
+   */
+  template <typename T>
+  [[nodiscard]] __host__ __device__ constexpr value_type native_value(T const& value) const noexcept
```
PointKernel:
```suggestion
  [[nodiscard]] __device__ constexpr value_type native_value(T const& value) const noexcept
```
Is host really needed?

PointKernel (commented):
One question: when to use `heterogeneous_value` and when to use `native_value`?

include/cuco/detail/static_set/static_set_ref.inl
```diff
@@ -148,14 +148,14 @@ class operator_impl<op::insert_tag,
...
    *
-   * @tparam Value Input type which is implicitly convertible to 'value_type'
+   * @tparam ProbeKey Input key type which is convertible to 'key_type'
    *
    * @param value The element to insert
    *
    * @return True if the given element is successfully inserted
    */
-  template <typename Value>
-  __device__ bool insert(Value const& value) noexcept
```
PointKernel: I was using `value_type` mainly to follow STL convention https://en.cppreference.com/w/cpp/container/unordered_set/insert. They probably do this for a reason. The same for other insert-related functions.

tests/static_map/erase_test.cu
```diff
@@ -32,51 +30,49 @@ using size_type = int32_t;
...
 
-  thrust::sequence(thrust::device, d_keys.begin(), d_keys.end(), 1);
-  thrust::sequence(thrust::device, d_values.begin(), d_values.end(), 1);
+  auto keys_begin = thrust::counting_iterator<key_type>(1);
 
-  auto pairs_begin =
-    thrust::make_zip_iterator(thrust::make_tuple(d_keys.begin(), d_values.begin()));
+  auto pairs_begin = thrust::make_transform_iterator(keys_begin, [] __device__(key_type const& x) {
+    return cuco::pair<key_type, mapped_type>(x, static_cast<mapped_type>(x));
+  });
```
PointKernel: guard device lambda with `cuda::proclaim_return_type`

tests/static_map/rehash_test.cu
```diff
@@ -36,14 +34,11 @@ TEST_CASE("Rehash", "")
...
-  thrust::device_vector<key_type> d_keys(num_keys);
-  thrust::device_vector<mapped_type> d_values(num_keys);
+  auto keys_begin = thrust::counting_iterator<key_type>(1);
 
-  thrust::sequence(d_keys.begin(), d_keys.end());
-  thrust::sequence(d_values.begin(), d_values.end());
-
-  auto pairs_begin =
-    thrust::make_zip_iterator(thrust::make_tuple(d_keys.begin(), d_values.begin()));
+  auto pairs_begin = thrust::make_transform_iterator(keys_begin, [] __device__(key_type const& x) {
```
PointKernel: `proclaim_return_type`

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
sleeepyjack: Not really. I tend to do this in preparation for host-sided probing on G+H but this would need much more changes. I'll remove it for now.

sleeepyjack:
> One question: when to use `heterogeneous_value` and when to use `native_value`?

`heterogeneous_value` is what you use whenever you need to call the custom hasher or key equals functor. `native_value` is what's stored in the container. The only difference is that `heterogeneous_value` holds a potentially custom key type while `native_value` always holds the native key type of the container.

The workflow is as follows: Take the user-provided pair-like input type and convert it to `heterogeneous_value`. This is the value type that is used internally and works with the custom hasher/predicate. Once you want to store the value, you convert it into `native_value`. So it's `Value`->`heterogeneous_value`->`native_value`.

include/cuco/detail/static_set/static_set_ref.inl
sleeepyjack: Yeah, it's a feature that isn't included in the STL containers, so this is new terrain. Do you think the name `Value` is too vague?

PointKernel: The point is your input type change. STL requires it as `value_type` but you change the input type requirement as convertible to `key_type`

PointKernel:
> > One question: when to use `heterogeneous_value` and when to use `native_value`?
> 
> `heterogeneous_value` is what you use whenever you need to call the custom hasher or key equals functor. `native_value` is what's stored in the container. The only difference is that `heterogeneous_value` holds a potentially custom key type while `native_value` always holds the native key type of the container.
> 
> The workflow is as follows: Take the user-provided pair-like input type and convert it to `heterogeneous_value`. This is the value type that is used internally and works with the custom hasher/predicate. Once you want to store the value, you convert it into `native_value`. So it's `Value`->`heterogeneous_value`->`native_value`.

When dealing with hasher/comparator, can we use `extract_key`?

include/cuco/detail/static_set/static_set_ref.inl
sleeepyjack: > convertible to key_type
To `value_type`.

Having this not fixed to the native type of the container is a requirement for heterogeneous insert. So we can either follow STL or have that feature. Maybe I'm not getting the point. 🙃

sleeepyjack:
> When dealing with hasher/comparator, can we use `extract_key`?

You mean using the user-provided pair-like type instead of the heterogeneous intermediate type internally and use `extract_…` on-demand?
The problem with that is that we don't even know on how to extract the members if it's any arbitrary type. `cuda::std::get`? `thrust::get`? `.first/.second`? This is just not manageable. Thus, we convert it to a `cuco::pair` first using `heterogeneous_value`, so we know how to access the members in a uniform fashion. From there, we can access the members directly without using any `extract_…` functions.

tests/static_map/rehash_test.cu
sleeepyjack: Addressed in [6cc9040](https://github.com/NVIDIA/cuCollections/pull/387/commits/6cc9040acc17e1873d5b3976ac07438a7da3b644)

tests/static_map/erase_test.cu
sleeepyjack: Addressed in [6cc9040](https://github.com/NVIDIA/cuCollections/pull/387/commits/6cc9040acc17e1873d5b3976ac07438a7da3b644)

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
sleeepyjack: Addressed in [b7ccda0](https://github.com/NVIDIA/cuCollections/pull/387/commits/b7ccda0711b951d0b9ab6f57e853810a2810f3b1)

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: Addressed in [f5bf6d8](https://github.com/NVIDIA/cuCollections/pull/387/commits/f5bf6d839a313bfbec3be8820a0fc7daee4e8883)

PointKernel:
> > When dealing with hasher/comparator, can we use `extract_key`?
> 
> You mean using the user-provided pair-like type instead of the heterogeneous intermediate type internally and use `extract_…` on-demand? The problem with that is that we don't even know on how to extract the members if it's any arbitrary type. `cuda::std::get`? `thrust::get`? `.first/.second`? This is just not manageable. Thus, we convert it to a `cuco::pair` first using `heterogeneous_value`, so we know how to access the members in a uniform fashion. From there, we can access the members directly without using any `extract_…` functions.

So we could remove `extract_key` and just use `val.first` instead all over the place?

sleeepyjack:
> So we could remove `extract_key` and just use `val.first` instead all over the place?

`extract_key` also handles the distinction between `static_set` and `static_map`, i.e., returning `val` or `val.first`.

include/cuco/detail/static_set/static_set_ref.inl
sleeepyjack: Thanks for the offline clearification. Addressed in [f933f72](https://github.com/NVIDIA/cuCollections/pull/387/commits/f933f728d8477461233f5def165912cd0c6f8233)

PointKernel (approved):
LGTM

PR #343, opened 2023-08-02
Test updates of CCCL (thrust, cub, libcudacxx) to 2.1.0.

bdice:
This PR updates cuCollections to support CCCL 2.1.0. Needed for https://github.com/rapidsai/rapids-cmake/pull/399.

**Do not merge until the rapids-cmake PR is ready for all repos.**

Depends on https://github.com/NVIDIA/cuCollections/pull/342.

PR #393, opened 2023-11-15
Migrate tests to excercise the new map

PointKernel:
This PR migrates static map tests to test the new static map:

- Large type tests are not migrated since the new map doesn't support keys larger than 8 bytes
- Shared memory tests require related functions to be added into ref code thus would be in a separate PR

### DIFF
--- a/tests/static_map/key_sentinel_test.cu
+++ b/tests/static_map/key_sentinel_test.cu
@@ -24,6 +24,8 @@
 
 #include <catch2/catch_template_test_macros.hpp>
 
+#include <cuda/functional>
+
 #define SIZE 10
 __device__ int A[SIZE];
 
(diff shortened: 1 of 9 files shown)


tests/static_map/key_sentinel_test.cu
```diff
@@ -60,25 +64,23 @@ TEMPLATE_TEST_CASE_SIG(
...
     // All keys inserted via custom `key_equal` should be found
-    REQUIRE(cuco::test::all_of(
-      pairs_begin, pairs_begin + num_keys, [view] __device__(cuco::pair<Key, Value> const& pair) {
-        auto const found = view.find(pair.first);
-        return (found != view.end()) and
-               (found->first.load() == pair.first and found->second.load() == pair.second);
-      }));
+    REQUIRE(cuco::test::all_of(pairs_begin,
+                               pairs_begin + num_keys,
+                               [find_ref] __device__(cuco::pair<Key, Value> const& pair) {
```
sleeepyjack: Do we need to specify the return type here via `cuda::proclaim_return_type`?

PointKernel: good point. Let me do this to make #343 easier.

sleeepyjack (approved):

PR #394, opened 2023-11-16
Fix clang 17 compilation

AustinSchuh:
This makes it so I'm able to compile basic code which uses cuco with clang 17.

### DIFF
--- a/include/cuco/detail/pair/pair.inl
+++ b/include/cuco/detail/pair/pair.inl
@@ -50,10 +50,6 @@ __host__ __device__ constexpr bool operator==(cuco::pair<T1, T2> const& lhs,
 
 }  // namespace cuco
 
-namespace thrust {
-#include <cuco/detail/pair/tuple_helpers.inl>
-}  // namespace thrust
-
 namespace cuda::std {
 #include <cuco/detail/pair/tuple_helpers.inl>
 }  // namespace cuda::std
--- a/include/cuco/detail/static_map.inl
+++ b/include/cuco/detail/static_map.inl
@@ -443,7 +443,7 @@ __device__
                 "insert_and_find is not supported for unpackable data on pre-Volta GPUs.");
 #endif
 
-  auto current_slot{initial_slot(insert_pair.first, hash)};
+  auto current_slot{this->initial_slot(insert_pair.first, hash)};
 
   while (true) {
     key_type const existing_key = current_slot->first.load(cuda::std::memory_order_relaxed);
@@ -514,7 +514,7 @@ __device__
 
     // if we couldn't insert the key, but it wasn't a duplicate, then there must
     // have been some other key there, so we keep looking for a slot
-    current_slot = next_slot(current_slot);
+    current_slot = this->next_slot(current_slot);
   }
 }
 


sleeepyjack:
@AustinSchuh FYI if you sign all of your existing and future commits to this PR then CI will run automatically on every push.

AustinSchuh:
> @AustinSchuh FYI if you sign all of your existing and future commits to this PR then CI will run automatically on every push.

I thought I had, looks like I hadn't.  I'll work on getting the build to pass.

PointKernel:
> sign all of your existing and future commits to this PR then CI will run automatically on every push.

I think this is valid only for contributors who have write access to the repo.

AustinSchuh:
OK, I think I need to wait for https://github.com/NVIDIA/cuCollections/pull/343 to merge first.  There have been a lot of changes in `thrust` which change how the pieces interact quite a bit.  I'd rather debug those after the update, especially since some of the older versions of `thrust` and friends don't work with clang.

PointKernel:
> OK, I think I need to wait for https://github.com/NVIDIA/cuCollections/pull/343 to merge first.  There have been a lot of changes in `thrust` which change how the pieces interact quite a bit.  I'd rather debug those after the update, especially since some of the older versions of `thrust` and friends don't work with clang.

Make sense. I will ping you once it's ready.

sleeepyjack:
@AustinSchuh FYI #343 was superseeded by #404 which has been merged. We are now using CCCL 2.2.0.

AustinSchuh:
Thanks @sleeepyjack !  I rebased the pieces which still apply, hopefully CI will be happy now.

include/cuco/detail/pair/pair.inl
```diff
@@ -50,10 +50,6 @@ __host__ __device__ constexpr bool operator==(cuco::pair<T1, T2> const& lhs,
 
 }  // namespace cuco
 
-namespace thrust {
-#include <cuco/detail/pair/tuple_helpers.inl>
-}  // namespace thrust
```
PointKernel, line 55: @sleeepyjack

Just a reminder to ourselves, `thrust::pair` is the same as `cuda::std::pair` so the current pair trait logic can be simplified.

include/cuco/detail/static_map.inl
```diff
@@ -443,7 +443,7 @@ __device__
                 "insert_and_find is not supported for unpackable data on pre-Volta GPUs.");
 #endif
 
-  auto current_slot{initial_slot(insert_pair.first, hash)};
+  auto current_slot{this->initial_slot(insert_pair.first, hash)};
```
PointKernel, line 446: I'm trying to understand the logic here since we have many places missing `this->` but how can they be built properly?

sleeepyjack:
I'll enable llvm in CI later today. Unfortunately the newest version we can use in CI is llvm16. Is this a problem?

AustinSchuh:
> I'll enable llvm in CI later today. Unfortunately the newest version we can use in CI is llvm16. Is this a problem?

(answering the easy questions while I'm distracted)  16 should be fine.  My understanding is it doesn't support the latest and greatest CUDA version available today, but that statement will be true with 17 at some point soon anyways.

sleeepyjack:
> I'll enable llvm in CI later today.

I think I underestimated this project a little. Adding llvm tests requires more changes to you CI setup than I expected. I think we can proceed with this PR without CI coverage for now.

include/cuco/detail/static_map.inl
sleeepyjack: I would be genuinely surprised if this is the only place where we forgot to use `this->`. So yeah, not sure why clang is complaining here but not in the other cases.

sleeepyjack (commented):
Looks good so far. Since we have no CI coverage yet, we must make sure that we test everything, i.e., make sure all of the examples+tests+benchmarks compile without any problems.

include/cuco/detail/static_map.inl
AustinSchuh: My suspicion is that this is the only code I am calling in my application and exercising.  I need to go dig in and get you a real answer though, not a guess.

PointKernel: > this is the only code I am calling in my application and exercising. 

Make sense.


@sleeepyjack we can merge the current PR as it is and fix other uncovered issues in your clang CI PR. What do you think?

PointKernel (approved):
LGTM

include/cuco/detail/static_map.inl
sleeepyjack: I'm fine with that, especially if it unblocks @AustinSchuh 's work for now.

sleeepyjack (approved):
LGTM 🚀

PR #396, opened 2023-11-18
Fix potential logic issues in CG erase

PointKernel:
This PR updates the CG erase code to be the same as it is in the legacy implementation. It also increases the input size for the erase tests and thus has a higher chance of uncovering potential issues in the code.

Ran the new tests on both sm75 and sm61 for about 30 mins and no tests failed.

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
@@ -595,12 +595,12 @@ class open_addressing_ref_impl {
           case insert_result::DUPLICATE: return false;
           default: continue;
         }
-      } else if (group.any(state == detail::equal_result::EMPTY)) {
-        // Key doesn't exist, return false
-        return false;
-      } else {
-        ++probing_iter;
       }
+
+      // Key doesn't exist, return false
+      if (group.any(state == detail::equal_result::EMPTY)) { return false; }
+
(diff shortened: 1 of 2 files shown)


include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -595,12 +595,12 @@ class open_addressing_ref_impl {
...
-        // Key doesn't exist, return false
-        return false;
-      } else {
-        ++probing_iter;
       }
+
+      // Key doesn't exist, return false
+      if (group.any(state == detail::equal_result::EMPTY)) { return false; }
+
+      ++probing_iter;
```
sleeepyjack, line 603: I still don't get why the old logic was wrong but if it fixes the problem with the flaky unit test then I'm down with it.

sleeepyjack (approved):
Do we expect a perf regression from this change?

Otherwise LGTM

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
PointKernel: Me neither. If this doesn't fix the issue, at least the new test increases the chance of test failures thus we can take a closer look at the point.

PointKernel:
No perf regression has been observed.

PR #400, opened 2023-11-30
Expose `window_extent` to device ref

PointKernel:
This is the first step to migrate shared memory code from the legacy map to the new map. It exposes the `window_extent` ref API thus it can be used to construct container device ref on shared memory.

### DIFF
--- a/include/cuco/detail/static_map/static_map_ref.inl
+++ b/include/cuco/detail/static_map/static_map_ref.inl
@@ -122,6 +122,26 @@ template <typename Key,
           typename ProbingScheme,
           typename StorageRef,
           typename... Operators>
+__host__ __device__ constexpr static_map_ref<Key,
+                                             T,
+                                             Scope,
+                                             KeyEqual,
+                                             ProbingScheme,
+                                             StorageRef,
+                                             Operators...>::extent_type
+static_map_ref<Key, T, Scope, KeyEqual, ProbingScheme, StorageRef, Operators...>::window_extent()
+  const noexcept
+{
+  return impl_.window_extent();
+}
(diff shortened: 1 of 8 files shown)


include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -115,6 +115,26 @@ static_map_ref<Key, T, Scope, KeyEqual, ProbingScheme, StorageRef, Operators...>
...
+                                             T,
+                                             Scope,
+                                             KeyEqual,
+                                             ProbingScheme,
+                                             StorageRef,
+                                             Operators...>::extent_type
+static_map_ref<Key, T, Scope, KeyEqual, ProbingScheme, StorageRef, Operators...>::window_extent()
+  const noexcept
+{
+  return impl_.window_extent();
```
sleeepyjack, line 135: This returns a `window_extent` which can't be implicitly converted to `size_type`. Do we want to return a `cuco::extent` instead? On second thought, maybe it's a better choice to be consistent with the return type. What do you think?

sleeepyjack (commented):
Happy 400th issue/PR 🥳

include/cuco/detail/static_map/static_map_ref.inl
PointKernel: The return value of this function will be used to construct a container ref thus window extent is exactly what we need.

sleeepyjack: Ah, right. Good point. Approved!

sleeepyjack (approved):

PR #402, opened 2023-12-05
Migrate shared memory code

PointKernel:
Depends on #403

This PR migrates the shared memory tests to exercise the new experimental map.


PointKernel:
Ideally, `make_copy` should be available only if query operators are present and `initialize` can be exclusively invoked from insert refs. However, there is no such "category" concept in the current design. This PR is mainly to make the code work and the followup API trimming or cleanups are necessary.

Those two functions are no longer "static" simply because I haven't figured out how to pass the variadic operator parameters via a static function. Is that even possible?

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -268,6 +278,63 @@ class open_addressing_ref_impl {
    */
   [[nodiscard]] __host__ __device__ constexpr iterator end() noexcept { return storage_ref_.end(); }
 
+  /**
+   * @brief Makes a copy of the current device ref using non-owned memory
```
```suggestion
   * @brief Makes a copy of the current device reference using non-owned memory.
```

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -268,6 +278,63 @@ class open_addressing_ref_impl {
...
+    window_type const* const windows_ptr = this->storage_ref().data();
+    for (size_type i = g.thread_rank(); i < num_windows; i += g.size()) {
+      memory_to_use[i] = windows_ptr[i];
+    }
+    g.sync();
+#endif
+  }
+
+  /**
+   * @brief Initializes the container storage using the threads in the group `tile`.
```
```suggestion
   * @brief Initializes the container storage.
```

include/cuco/static_map_ref.cuh
```diff
@@ -166,6 +166,20 @@ class static_map_ref
...
+   * @return The sentinel value used to represent an erased key slot
+   */
+  [[nodiscard]] __host__ __device__ constexpr key_type erased_key_sentinel() const noexcept;
+
+  /**
+   * @brief Gets the key comparator.
+   *
+   * @return The comparator used to compare keys
+   */
+  [[nodiscard]] __host__ __device__ constexpr key_equal key_eq() const noexcept;
```
sleeepyjack, line 181: STL also exposes both `key_eq` and the hash function. We should do the same for the new set/map and the corresponding refs for API parity reasons. Could be another PR.

include/cuco/static_set_ref.cuh
```diff
@@ -161,6 +175,36 @@ class static_set_ref
   template <typename... NewOperators>
   [[nodiscard]] __host__ __device__ auto with(NewOperators... ops) && noexcept;
 
+  /**
+   * @brief Makes a copy of the current device ref using non-owned memory
```
sleeepyjack:
```suggestion
   * @brief Makes a copy of the current device reference using non-owned memory.
```

include/cuco/static_map_ref.cuh
```diff
@@ -183,6 +197,36 @@ class static_map_ref
   template <typename... NewOperators>
   [[nodiscard]] __host__ __device__ auto with(NewOperators... ops) && noexcept;
 
+  /**
+   * @brief Makes a copy of the current device ref using non-owned memory
+   *
+   * This function is intended to be used to create shared memory copies of small static maps,
+   * although global memory can be used as well.
+   *
+   * @tparam CG The type of the cooperative thread group
```
sleeepyjack:
```suggestion
   * although global memory can be used as well.
   *
   * @note This function synchronizes the group `tile`.
   *
   * @tparam CG The type of the cooperative thread group
```

include/cuco/static_set_ref.cuh
```diff
@@ -161,6 +175,36 @@ class static_set_ref
...
+   *
+   * This function is intended to be used to create shared memory copies of small static sets,
+   * although global memory can be used as well.
+   *
+   * @tparam CG The type of the cooperative thread group
```
sleeepyjack:
```suggestion
   * although global memory can be used as well.
   *
   * @note This function synchronizes the group `tile`.
   *
   * @tparam CG The type of the cooperative thread group
```

tests/static_map/shared_memory_test.cu
```diff
@@ -93,31 +98,35 @@ TEMPLATE_TEST_CASE_SIG("Shared memory static map",
...
-    maps.push_back(std::make_unique<MapType>(
-      map_capacity, cuco::empty_key<Key>{-1}, cuco::empty_value<Value>{-1}));
+    maps.push_back(std::make_unique<map_type>(
+      extent_type{}, cuco::empty_key<Key>{-1}, cuco::empty_value<Value>{-1}));
   }
 
   thrust::device_vector<bool> d_keys_exist(number_of_maps * elements_in_map);
   thrust::device_vector<bool> d_keys_and_values_correct(number_of_maps * elements_in_map);
 
+  using ref_type = decltype(maps.front()->ref(cuco::experimental::op::insert));
```
sleeepyjack: Can we instead use `map_type::ref_type`, i.e., ref without any operators and then in the kernel dispatch the ref with `insert`?

include/cuco/static_map_ref.cuh
PointKernel: Right. The only reason that I choose not to expose the hash function is "double hashing". :(

tests/static_map/shared_memory_test.cu
PointKernel:
```cpp
using ref_type = typename map_type::ref_type<cuco::experimental::op::insert_tag>;
```
Cannot be better. Good call!

sleeepyjack (approved):
LGTM 🚀

PR #403, opened 2023-12-05
Fix a static extent bug in probing scheme

PointKernel:
This PR fixes a bug in the probing scheme where the extent value must be explicitly converted to `size_type`. It also removes extent `value()` functions to avoid misuse.


include/cuco/detail/probing_scheme_impl.inl
```diff
@@ -100,7 +100,7 @@ __host__ __device__ constexpr auto linear_probing<CGSize, Hash>::operator()(
 {
   using size_type = typename Extent::value_type;
   return detail::probing_iterator<Extent>{
-    cuco::detail::sanitize_hash<size_type>(hash_(probe_key)) % upper_bound,
+    cuco::detail::sanitize_hash<size_type>(hash_(probe_key)) % static_cast<size_type>(upper_bound),
```
sleeepyjack: IIUC the `static_cast` demotes the `fast_int` to a simple integer in the dynamic extent case. This eliminates the fast modulo optimization.

include/cuco/detail/extent/extent.inl
```diff
@@ -31,8 +31,7 @@ template <typename SizeType, std::size_t N>
 struct window_extent {
   using value_type = SizeType;  ///< Extent value type
 
-  __host__ __device__ constexpr value_type value() const noexcept { return N; }
-  __host__ __device__ explicit constexpr operator value_type() const noexcept { return value(); }
+  __host__ __device__ explicit constexpr operator value_type() const noexcept { return N; }
```
sleeepyjack: I would keep the `value()` member function for the case where the `value_type` isn't known. For this case, downcasting would look like this: `static_cast<decltype(typename extent_type::value_type)>(extent)` which is a mouthful compared to `extent.value()`

include/cuco/detail/probing_scheme_impl.inl
PointKernel: Ah I see. Thanks for the clarification. Similarly to what you do in https://github.com/NVIDIA/cuCollections/blob/3fe5704b26a877f9321f07128b5764463e536012/include/cuco/utility/fast_int.cuh#L153-L157

I need to add a custom `%` operator for static extent then.

sleeepyjack: Yeah, that's a bit inconvenient since a static window extent is no longer implicitly convertible to a plain int but at least we get uniform behavior that way. I'm open to suggestions how to improve this.

include/cuco/detail/probing_scheme_impl.inl
```diff
@@ -149,9 +150,10 @@ __host__ __device__ constexpr auto double_hashing<CGSize, Hash1, Hash2>::operato
 {
   using size_type = typename Extent::value_type;
   return detail::probing_iterator<Extent>{
-    cuco::detail::sanitize_hash<size_type>(hash1_(probe_key) + g.thread_rank()) % upper_bound,
+    cuco::detail::sanitize_hash<size_type>(hash1_(probe_key) + g.thread_rank()) %
+      static_cast<size_type>(upper_bound),
     static_cast<size_type>((cuco::detail::sanitize_hash<size_type>(hash2_(probe_key)) %
-                              (upper_bound.value() / cg_size - 1) +
```
PointKernel, line 154: @sleeepyjack  I find there is a custom `/` for `fast_int` but the `fast_int` must be on the right side of the operator. Does `fast_int.value()` here affect any fast math optimization?

sleeepyjack: Yeah the / operator only works if the fast_int is in the denominator. Unfortunately we have to use the builtin operator here, thus the conversion to int.

sleeepyjack: I wonder if there's an identity which allows us to use the fast_int modulo operator but still gives the same result.

PointKernel: > use the fast_int modulo operator but still gives the same result.

the `fast_int` is on the left side this time so I'm not sure.

I've added a custom `/` operator in `fast_int` (line 153-157 in `fast_int.cuh`). Let me know if this fix the issue properly.

include/cuco/utility/fast_int.cuh
```diff
@@ -150,10 +150,16 @@ struct fast_int {
     return rhs.mulhi(rhs.magic_, mul) >> rhs.shift_;
   }
 
+  template <typename Rhs>
+  friend __host__ __device__ constexpr value_type operator/(fast_int const& lhs, Rhs rhs) noexcept
+  {
+    return lhs.value() / static_cast<value_type>(rhs);
```
sleeepyjack: Is the static cast required? I would instead let the compiler figure out the result type depending on `Rhs` rather than enforcing the `value_type`. The return type can be `auto` too in this case.

sleeepyjack (approved):

PR #406, opened 2023-12-14
Fix CTAD

sleeepyjack:
This PR fixes CTAD by allowing all class tparams to be inferred from ctor arguments. 

Changes:
- Add a `cuco::experimental::cuda_thread_scope` strong type and add it to the ctor argument list
- Add `cuco::experimental::storage` to the ctor argument list


include/cuco/cuda_thread_scope.cuh
PointKernel, line 1: maybe moving to the utility folder?

tests/static_map/shared_memory_test.cu
```diff
@@ -182,17 +182,13 @@ __global__ void shared_memory_hash_table_kernel(bool* key_found)
   using storage_ref_type = cuco::experimental::aow_storage_ref<slot_type, window_size, extent_type>;
 
   // CTAD doesn't work for container ref types
-  auto raw_ref = cuco::experimental::static_map_ref<
-    Key,
-    Value,
-    cuda::thread_scope_device,
```
sleeepyjack, line 188: We should use `thread_scope_block` here. Fixed it.

include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -231,6 +233,7 @@ static_map_ref<Key, T, Scope, KeyEqual, ProbingScheme, StorageRef, Operators...>
     cuco::erased_key<Key>{this->erased_key_sentinel()},
     this->key_eq(),
     this->impl_.probing_scheme(),
+    {},
```
PointKernel, line 236: Question: a default ctor will create an enum of value zero, but does zero correspond to our default `thead_scope_device`? But all of these will be changed based on our offline discussion.

include/cuco/detail/static_set/static_set_ref.inl
PointKernel, line 1: Similar to map ref `make_copy`, the set ref `make_copy` should also pass the scope around but it seems to be missed due to no set ref tests exercising this.

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: `{}` here calls the default ctor of the new `cuda_thread_scope<Scope>` strong type. If not explicitly specified, set/map will use the default `Scope` tparam `cuda::thread_scope_device`. That way we never actually have to deal with the underlying enum values and rely on the libcu++ definitions.

include/cuco/detail/static_set/static_set_ref.inl
sleeepyjack: Good catch! I missed that one. We should at some point aim for parity between map/set tests.

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: I just noticed this is the ctor call in `make_copy`. In this case `{}` will always use the scope of the parent container for the `cuco::cuda_thread_scope` parameter. ~~Let me changed that the way we discussed offline.~~ Better to do it in a separate small PR.

include/cuco/detail/static_set/static_set_ref.inl
sleeepyjack: Fixed in [188e973](https://github.com/NVIDIA/cuCollections/pull/406/commits/188e973a03debac2ff30c3cf5b914f69f916c19d)

include/cuco/detail/static_map/static_map_ref.inl
sleeepyjack: See #408 and commit [229df4](https://github.com/NVIDIA/cuCollections/pull/408/commits/3229df4ccfa31f4fa1aa94c2b8c8553f61ca7959)

PointKernel: I feel there are too many indirection layers to pass the thread scope down to the base implementation. The open addressing base class has a [thread_scope](https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/open_addressing/open_addressing_impl.cuh#L89) defined but the open addressing ref base class doesn't have this definition.

@sleeepyjack Can you please add `thread_scope` to ref base class as well and use it consistently wherever `Scope` is involved? Also, can you please add a dedicated unit test verifying `map_type::thread_scope`, `map_ref_type::thread_scope`, the thread scope of the return ref of `map_ref::make_copy`, etc have the proper thread scope?

sleeepyjack: > Can you please add thread_scope to ref base class

~~The OA impl class doesn't need this ctor parameter. The `cuda_thread_scope` type is just meant to make the user experience better (and also allow for passing the scope at compile time). Internally, we can still use the `cuda::thread_scope` type.~~ I think I misunderstood the question. See my answer in the next comment.

> Also, can you please add a dedicated unit test verifying map_type::thread_scope, map_ref_type::thread_scope, the thread scope of the return ref of map_ref::make_copy, etc have the proper thread scope?

Yes, let me think about how to best do this.

sleeepyjack: > The open addressing base class has a [thread_scope](https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/open_addressing/open_addressing_impl.cuh#L89) defined but the open addressing ref base class doesn't have this definition.

I had to fix that in #408. Now all classes have a `thread_scope` alias definition.

PointKernel: OK so this PR is to make CTAD work and #408 is to fix the thread scope issue. Got it.

PointKernel (approved):

include/cuco/static_map.cuh
```diff
@@ -179,6 +182,8 @@ class static_map {
                        empty_value<T> empty_value_sentinel,
                        KeyEqual const& pred                = {},
                        ProbingScheme const& probing_scheme = {},
+                       cuda_thread_scope<Scope> scope      = {},
+                       Storage storage                     = {},
```
PointKernel, line 186: @sleeepyjack I just realized this might be overkill for host ctors since CTAD worked just fine.

In all use cases in libcudf, I have to add two additional empty braces to adapt to this change. How can we improve this?

sleeepyjack: The variadic ctor idea would come in handy here. Since we now control almost all argument types but Allocator and KeyEqual, we can give it another shot.

Apart from that we could reorder the arguments but that would be tedious and would require downstream projects to adapt that change.

I would not advise to remove them, because then CTAD might no longer work in certain cases IIRC.

PointKernel: > Since we now control almost all argument types but Allocator and KeyEqual, we can give it another shot.

Allocator and KeyEqual are under control as well. So all arguments are of strong types.

> Apart from that we could reorder the arguments but that would be tedious and would require downstream projects to adapt that change.

The issue is that stream should be always the last argument but it is also the argument we custom the most...

sleeepyjack: > Allocator and KeyEqual are under control as well.

Not sure. Both can be user-defined types at this point.

Allocator is easy, since it has to fulfill the allocator interface. We can simply use a trait for that one.

KeyEqual gave me some headaches in the past, since the only requirement is that it defines a binary operator for the comparison. So if we use that as a check, every argument type that has a binary call operator defined will be considered as a KeyEqual concept.

PointKernel: > So if we use that as a check, every argument type that has a binary call operator defined will be considered as a KeyEqual concept.


It should be fine if the return type of the binary operator is bool and it's invokable with two arguments of the type `Key`

sleeepyjack: > It should be fine if the return type of the binary operator is bool and it's invokable with two arguments of the type Key

That was also my initial thoughts on this but maan was I wrong and wasted some hours trying to fix it :D

Turns out everything that can be implicitly converted to a `bool` (so basically any POD) will suffice the first constraint.
The second constraint is also no longer valid, since we introduced the heterogeneous insert/lookup concept. We actually compare `ProbeKey`and `Key`, where `ProbeKey` type isn't known in the ctor.

I can think of two ways we can reliably detect a `KeyEqual`-like type:

1. We require the user to apply a wrapper or a tag to their user-defined `KeyEqual` and we then use that in a trait. It could be a wrapper like `cuco::key_equal<MyKeyEqual>` or we define a macro to do the wrapping for you, e.g., `CUCO_DECLARE_KEY_EQUAL(MyKeyEqual)`. Tag member structs would also work, but I find them ugly from a design perspective. Then there's also the possibility of using inheritance, e.g., `struct MyKeyEqual : cuco::key_equal {…}`.
3. If we look at all the possible argument types, then there's only one other user-defined argument type (the `Allocator`) that _could possibly_ implement such a binary operator for whatever reason. So we can define the trait like this: `is_key_equal = is_binary_callable and not is_allocator_like`. There's one more thing required to make this actually work: Our probing scheme somehow also defines a binary callable operator. So the final trait would look like this: `is_key_equal = is_binary_callable and not is_allocator and not is_probing_scheme`.

PointKernel: > Turns out everything that can be implicitly converted to a bool

I'm inclined to restrict the requirement here: it should be bool instead of "convertible to bool"

> We actually compare ProbeKey and Key, where ProbeKey type isn't known in the ctor.

Is it fair to say that supporting the comparison between `Key` and `ProbeKey` is a fancy extra and the `KeyEqual` being able to compare two `Key`s is a basic requirement?

sleeepyjack: > I'm inclined to restrict the requirement here: it should be bool instead of "convertible to bool"

I totally agree. The problem was on the implementation side of the trait. I just couldn't think of a way to make this restriction work with the given STL tools. There might be a way.

> Is it fair to say that supporting the comparison between Key and ProbeKey is a fancy extra and the KeyEqual being able to compare two Keys is a basic requirement?

Fair point. I would say requiring a `Key<->Key` comparator (even it is actually never used) is reasonable.

Since we still need to check for '… and not is_allocator and not is_probing_scheme`, I would say proposal 2. plus the just discussed trait is the way to go here?

PR #407, opened 2023-12-15
Migrate static map examples

PointKernel:
This PR migrates all single map examples to use the new experimental map. It also fixes a bug where bitwise compare should be used to compare sentinel values.


examples/static_map/count_by_key_example.cu
```diff
@@ -71,13 +71,14 @@ __global__ void count_by_key(Map map_view,
   uint64_t thread_unique_keys = 0;
   while (idx < num_keys) {
     // insert key into the map with a count of 1
-    auto [slot, is_new_key] = map_view.insert_and_find({keys[idx], 1});
+    auto [slot, is_new_key] = map_ref.insert_and_find(cuco::pair{keys[idx], 1});
```
sleeepyjack, line 74: Side note: It would be nice if we would not have to specify the pair type but that's currently a limitation of the heterogeneous insert implementation. I wonder if it would work if we add an overload `insert_and_find(value_type)`.

examples/static_map/count_by_key_example.cu
```diff
@@ -125,34 +126,37 @@ int main(void)
...
-  // Constructs a map with "capacity" slots.
-  cuco::static_map<Key, Count> map{
-    capacity, cuco::empty_key{empty_key_sentinel}, cuco::empty_value{empty_value_sentinel}};
+  // Constructs a map with number of elements and desired load factor.
+  auto map = cuco::experimental::static_map{
+    num_elements,
+    load_factor,
+    cuco::empty_key{empty_key_sentinel},
+    cuco::empty_value{empty_value_sentinel},
+    thrust::equal_to<Key>{},
```
sleeepyjack, line 138: Side note 2: Spcifying `thrust::equal_to` feels weird. Should we add default values to the ref's tparams? Also, we should add an alias `cuco::default_key_equal = thrust::equal_to` of some sort.

examples/static_map/custom_type_example.cu
```diff
@@ -25,32 +25,13 @@
 #include <cuda/functional>
 
 // User-defined key type
-#if !defined(CUCO_HAS_INDEPENDENT_THREADS)
```
sleeepyjack, line 28: Are we removing this switch to make the example less complex?

examples/static_map/custom_type_example.cu
```diff
@@ -91,15 +67,20 @@ int main(void)
...
-      [] __device__(auto i) { return cuco::make_pair(custom_key_type{i}, custom_value_type{i}); }));
+      [] __device__(auto i) {
+        return cuco::pair{custom_key_type{i}, custom_value_type{i}};
+      }));
 
   // Construct a map with 100,000 slots using the given empty key/value sentinels. Note the
   // capacity is chosen knowing we will insert 80,000 keys, for an load factor of 80%.
-  cuco::static_map<custom_key_type, custom_value_type> map{
-    100'000, cuco::empty_key{empty_key_sentinel}, cuco::empty_value{empty_value_sentinel}};
+  auto map = cuco::experimental::static_map{cuco::experimental::extent<std::size_t, 100'000>{},
```
sleeepyjack, line 76: Love the idea of showing off the static extent concept here

include/cuco/detail/static_map/static_map.inl
```diff
@@ -570,7 +571,7 @@ auto static_map<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Stora
   Operators...) const noexcept
 {
   static_assert(sizeof...(Operators), "No operators specified");
-  return this->empty_key_sentinel() == this->erased_key_sentinel()
+  return cuco::detail::bitwise_compare(this->empty_key_sentinel(), this->erased_key_sentinel())
```
sleeepyjack, line 574: whups, that was a solid bug

examples/static_map/count_by_key_example.cu
PointKernel: > Should we add default values to the ref's tparams?

IIRC, we cannot add default parameters to ref templates because of the variadic parameter in the end.

> we should add an alias cuco::default_key_equal = thrust::equal_to of some sort.

No preference here. OK to add but not required.

examples/static_map/custom_type_example.cu
PointKernel: The new map doesn't support key types larger than 8 bytes.

examples/static_map/count_by_key_example.cu
PointKernel: A nice to have feature. Let me file an issue.

sleeepyjack (approved):
LGTM 🚀

examples/static_map/count_by_key_example.cu
PointKernel: #410

PR #408, opened 2023-12-18
Add option to change the thread scope when copying a container

sleeepyjack:
This PR adds an option to specify the thread scope of the newly created device ref when copying containers using `make_copy`.

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
@@ -111,7 +111,8 @@ class open_addressing_ref_impl {
 
   static constexpr auto cg_size = probing_scheme_type::cg_size;  ///< Cooperative group size
   static constexpr auto window_size =
-    storage_ref_type::window_size;  ///< Number of elements handled per window
+    storage_ref_type::window_size;             ///< Number of elements handled per window
+  static constexpr auto thread_scope = Scope;  ///< CUDA thread scope
   static constexpr auto has_payload =
     not std::is_same_v<key_type, value_type>;  ///< Determines if the container is a key/value or
                                                ///< key-only store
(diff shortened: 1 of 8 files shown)


include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -111,7 +111,8 @@ class open_addressing_ref_impl {
 
   static constexpr auto cg_size = probing_scheme_type::cg_size;  ///< Cooperative group size
   static constexpr auto window_size =
-    storage_ref_type::window_size;  ///< Number of elements handled per window
+    storage_ref_type::window_size;             ///< Number of elements handled per window
+  static constexpr auto thread_scope = Scope;  ///< CUDA thread scope
```
PointKernel, line 115: shall we use `thread_scope` consistently across the base ref class?

PointKernel (approved):
One suggestion. Otherwise LGTM

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
sleeepyjack: I tried to do that e.g. when we dispatch the CUDA-C atomics in the CAS funtion but I ran into some errors. I didn't dive deeper into it but it might have to do with the mixin stuff since inside the CAS we don't cast *this to the ref class and so theead_scope might not be available.

sleeepyjack: AFAIK apart from the sections mentioned, there are no other occurrences of Scope in the OA implementation

PointKernel: I recall I ran into similar issues as well.

PR #409, opened 2023-12-19
Migrate static map benchmarks

PointKernel:
Depends on #412 

This PR migrates all single map benchmarks to use the new experimental map.

PR #412, opened 2023-12-21
Fix key generator utility

PointKernel:
I ran into segfault issues multiple times when invoking specific benchmark cases (e.g. `./STATIC_MAP_BENCH -b 0`) instead of running the whole benchmark suite. The root cause was the host device lambda functions used during key generation/dropout. 

This PR fixes this issue by replacing lambda functions with functors.


PointKernel:
required by #409

benchmarks/hash_table/dynamic_map/contains_bench.cu
```diff
@@ -18,7 +18,7 @@
 #include <utils.hpp>
 
 #include <cuco/dynamic_map.cuh>
-#include <cuco/utility/key_generator.hpp>
+#include <cuco/utility/key_generator.cuh>
```
sleeepyjack, line 21: Oh noes, this might have been part of the culprit. A .hpp file with device code/lambdas in it.

sleeepyjack (approved):
LGTM 🚀 Thanks for catching this!

PR #414, opened 2023-12-28
Move old single map to `cuco::legacy` namespace

PointKernel:
This PR moves the old `static_map` to `cuco::legacy` namespace

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ Similar to how [Thrust](https://github.com/thrust/thrust) and [CUB](https://gith
 
 `cuCollections` is still under heavy development. Users should expect breaking changes and refactoring to be common.
 
+### Major Updates
+__01/02/2024__ The legacy `static_map` is moved to `cuco::legacy` namespace.
+
 ## Getting cuCollections
 
 `cuCollections` is header only and can be incorporated manually into your project by downloading the headers and placing them into your source tree.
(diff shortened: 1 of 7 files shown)


sleeepyjack:
Let's discuss the plan on how to phase out the old map implementation:

First of all, for users who can't/don't want to change their existing code, we should push a git tag `legacy/static_map` for the last commit before this PR, so people can use this as the last commit that works for them. We should also add a disclaimer to the readme and also inform users via Slack, e.g. by at-here in our cuco channel.

After this PR, we can remove the `experimental` namespace from the new implementation.

What would be the step after that? Remove the `legacy` map entirely from the repo?

PointKernel:
> First of all, for users who can't/don't want to change their existing code, we should push a git tag legacy/static_map for the last commit before this PR, so people can use this as the last commit that works for them. We should also add a disclaimer to the readme and also inform users via Slack, e.g. by at-here in our cuco channel.

In README, it's stated already that cuco is under active development thus changes are expected https://github.com/NVIDIA/cuCollections?tab=readme-ov-file#development-status. We are moving the old map to the legacy namespace but will still keep it for a while since the dynamic map still depends on the old static map. And also the new map doesn't support large keys/payloads.

> After this PR, we can remove the experimental namespace from the new implementation.

Yeah, I planned to inline the namespace first to avoid breaking RAPIDS projects. Once all RAPIDS are updated to be able to build with the inline experimental namespace, then we can remove it.


> What would be the step after that? Remove the legacy map entirely from the repo?

That's the plan. It also depends on when we could add lock table into libcudacxx to enable large keys/payloads and when we can migrate the dynamic map to use the new map.

sleeepyjack:
> In README, it's stated already that cuco is under active development thus changes are expected

True. I was just thinking since this is a breaking change for any projects that use the old static map, it would be helpful to have a document/paragraph on how to resolve the issue (i.e. switch to the new implementation or add the legacy namespace). Then we can just point users to this document.

sleeepyjack (approved):

PR #415, opened 2023-12-29
Move probe sequences to legacy namespace

PointKernel:
Similar to #414, this is a preparation for inlining `experimental` namespace. It moves the old probe sequences which are only used by the current multimap old into `cuco::legacy` namespace.

### DIFF
--- a/include/cuco/detail/probe_sequence_impl.cuh
+++ b/include/cuco/detail/probe_sequence_impl.cuh
@@ -25,8 +25,7 @@
 
 #include <utility>
 
-namespace cuco {
-namespace detail {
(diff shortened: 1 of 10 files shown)


sleeepyjack:
Shall we move this to the previous PR, so we have a single commit that places the old map into the legacy namespace?

sleeepyjack (approved):

PointKernel:
The change is required to avoid conflicts between the old and new probe schemes like `cuco::linear_probing` and `cuco::double_hashing`. This PR doesn't affect the legacy single map thus I'm inclined to keep them separate.

PR #416, opened 2023-12-29
Get rid of unused var warnings

PointKernel:
This PR adds `[[maybe_unused]]` attributes to avoid false-positive "variable set but not used" warnings (sample code https://godbolt.org/z/soPqcTdhP) in shared memory tests.

### DIFF
--- a/tests/static_map/shared_memory_test.cu
+++ b/tests/static_map/shared_memory_test.cu
@@ -123,7 +123,9 @@ TEMPLATE_TEST_CASE_SIG("Shared memory static map",
     }
     thrust::device_vector<ref_type> d_refs(h_refs);
 
-    auto constexpr num_windows = cuco::experimental::make_window_extent<ref_type>(extent_type{});
+    // maybe_unused to silence false positive "variable set but not used" warning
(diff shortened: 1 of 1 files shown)


tests/static_map/shared_memory_test.cu
```diff
@@ -123,7 +123,9 @@ TEMPLATE_TEST_CASE_SIG("Shared memory static map",
     }
     thrust::device_vector<ref_type> d_refs(h_refs);
 
-    auto constexpr num_windows = cuco::experimental::make_window_extent<ref_type>(extent_type{});
+    // maybe_unused to silence false positive "variable set but not used" warning
+    [[maybe_unused]] auto constexpr num_windows =
```
sleeepyjack, line 127: This workaround is fine. Should we get the compiler team involved so they can investigate?

sleeepyjack (approved):

tests/static_map/shared_memory_test.cu
PointKernel: Based on offline discussions, it's been a known gcc issue for a while.

PR #428, opened 2024-01-23
Fix the default erase sentinel bug in the base open addressing ctor

PointKernel:
This PR fixes a bug in the OA base class where the erased key sentinel value should have been initialized by the empty key sentinel if not specified.

Tests are updated to exercise this issue.

Needed by https://github.com/rapidsai/cudf/pull/14813

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -186,7 +186,7 @@ class open_addressing_impl {
                alloc}
   {
     CUCO_EXPECTS(desired_load_factor > 0., "Desired occupancy must be larger than zero");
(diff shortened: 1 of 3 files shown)


include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -178,14 +178,15 @@ class open_addressing_impl {
                                  cuda_stream_ref stream)
     : empty_key_sentinel_{empty_key_sentinel},
       empty_slot_sentinel_{empty_slot_sentinel},
+      erased_key_sentinel_{empty_key_sentinel},
```
sleeepyjack, line 181: Good catch!

sleeepyjack (approved):

PR #434, opened 2024-01-31
Get rid of container dependent functors by adding `has_payload` in OA base class

PointKernel:
Similar to the OA ref class, this PR adds a `has_payload` constexpr to the OA base class. The new constexpr is used to get rid of container-dependent functors like `slot_is_filled` and `get_slot`.


include/cuco/detail/open_addressing/functors.cuh
```diff
@@ -44,9 +47,64 @@ struct get_slot {
...
+        // required by thrust zip iterator in `retrieve_all`
+        if constexpr (cuco::detail::is_thrust_pair_like<S>::value) {
+          return thrust::get<0>(slot);
+        } else {
+          return slot.first;
+        }
+      } else {
+        return slot;
+      }
+    }();
```
PointKernel, line 104: Do we need to guard the lambda with `cuda::proclaim_return_type`? @bdice probably knows the answer the best.

include/cuco/detail/open_addressing/functors.cuh
```diff
@@ -44,9 +47,64 @@ struct get_slot {
   {
     auto const window_idx = idx / StorageRef::window_size;
     auto const intra_idx  = idx % StorageRef::window_size;
-    return storage_[window_idx][intra_idx];
+    if constexpr (HasPayload) {
+      auto const& [first, second] = storage_[window_idx][intra_idx];
+      return thrust::make_tuple(first, second);
```
sleeepyjack, line 52: Why do we wanna repack the slot's content which is already a `cuco::pair` into a `thrust::tuple`?

PointKernel: Good catch. This is only needed by `retrieve_all` where our output type is a zip iterator for map (equivalent to thrust::tuple):
https://github.com/NVIDIA/cuCollections/blob/fabd885327dc1662e1a63d9b8870ebf888893c09/include/cuco/detail/open_addressing/open_addressing_impl.cuh#L566-L573
And to assign a tuple with cub copy_if, the input type must be a tuple as well.

sleeepyjack: I see. If it's only used by `retrieve_all` then it should be fine. However, we should avoid using it anywhere in the hot path since repacking might hurt performance. BTW `cuco::pair` is somewhat tuple-like (for some reason the thrust getter doesn't work): https://godbolt.org/z/x8v7W78W4

PointKernel: > we should avoid using it anywhere in the hot path since repacking might hurt performance.

This functor is only used by `retrieve_all` so we should be good.

sleeepyjack (approved):

include/cuco/detail/open_addressing/functors.cuh
bdice: Now that we're on CCCL 2.2.0 as a minimum requirement, you will see build errors if you need `cuda::proclaim_return_type` and don't have it. In other words, if this is building, you're fine!

PointKernel: @bdice Realized it's the case after asking you. Thanks for the confirmation!

PR #437, opened 2024-02-02
Add `static_multiset` and its insert APIs

PointKernel:
This PR adds `static_multiset` and its insert-related APIs. This is a proof of concept of sharing implementation details between set and multiset containers.


include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -73,12 +73,14 @@ struct window_probing_results {
  * @tparam KeyEqual Binary callable type used to compare two keys for equality
  * @tparam ProbingScheme Probing scheme (see `include/cuco/probing_scheme.cuh` for options)
  * @tparam StorageRef Storage ref type
+ * @tparam SupportsDuplicateEntries Flag indicating whether duplicate entries are allowed or not
```
```suggestion
 * @tparam AllowDuplicates Flag indicating whether duplicate keys are allowed or not
```

Suggestion

include/cuco/static_set_ref.cuh
```diff
@@ -64,8 +64,16 @@ class static_set_ref
...
+  /// Flag indicating whether duplicate entries are allowed or not
+  static constexpr auto supports_duplicate_entries = false;
+
+  /// Implementation type
+  using impl_type = detail::open_addressing_ref_impl<Key,
+                                                     Scope,
+                                                     KeyEqual,
+                                                     ProbingScheme,
+                                                     StorageRef,
+                                                     supports_duplicate_entries>;
```
sleeepyjack: Instead of exposing this to the public API, we could use a type selector in detail namespace that selects the appropriate input type based on the public type, e.g., `detail::impl_type<static_set_ref>`.
`detail::impl_type` would then specialize for `static_set_ref` and emit the OA ref impl class.

sleeepyjack (approved):
Very nice PR!

include/cuco/static_set_ref.cuh
PointKernel: This `impl_type` is private. Are you suggesting using a detail type selector to avoid exposing the private `impl_type` alias in a public header?

sleeepyjack: Yeah that's basically it. It's just a minor remark and doesn't need to be dealt with in this PR.

sleeepyjack:
Reminder to include the ODR fix from #440 into this PR too

PR #438, opened 2024-02-06
Add shared memory example

sleeepyjack:
Demonstrates usage of the `static_set` in shared memory.


examples/static_set/shared_memory_example.cu
```diff
@@ -0,0 +1,107 @@
...
+  // Note that if you want to use a cg_size other then one, you have to use the cooperative
+  // overload of this function, i.e., insert(cg, key);
+  insert_ref.insert(key);
+
+  // Synchronize the cta to make sure all insert operations have finished.
+  block.sync();
+  // Next, we want to check if the keys can be found again using the `contains` function.
+  // We move the `insert_ref` to a new object that supports `contains` but no longer supports
+  // `insert`.
+  auto contains_ref = std::move(insert_ref).with(cuco::contains);
```
sleeepyjack:
```suggestion
  auto const contains_ref = std::move(insert_ref).with(cuco::contains);
```

include/cuco/detail/extent/extent.inl
```diff
@@ -30,11 +30,12 @@ template <typename SizeType, std::size_t N>
 struct window_extent {
   using value_type = SizeType;  ///< Extent value type
 
+  __host__ __device__ explicit constexpr window_extent() noexcept {}
```
PointKernel: Oh man, this is wild. `window_extent` ctor must be private. Imagine using this default ctor together with a dynamic extent...

sleeepyjack: Dynamic `window_extent` is a template specialization which has has the ctor still private. I just enabled the default ctor for the static case. This make the static shmem usecase a bit easier, since you can get the actual extent by calling `typename ref_type::extent_type{}`.

sleeepyjack: Oh, wow, I missed that. Although the dynamic extent isn't the problem here since it's a different specialization with the ctor still being private. The problem is that a user can specify any size and it will be assumed as a valid size. I'll roll back this commit.

PointKernel (approved):
LGTM.

Could add a line at the top-level README about this example as well.

PR #440, opened 2024-02-12
Fix ODR violation where end() might be defined by multiple mixin operators

sleeepyjack:
This PR fixes an ODR violation that occurs when mixing multiple operators that define `end()` in their body, e.g., `find()` and `insert_and_find().`

Solution: Move definition of `end()` to the `_ref` class instead of defining it in the operators.


PointKernel (approved):
LGTM

PR #441, opened 2024-02-21
Improve erased sentinel check logic

PointKernel:
This PR cleans up the equal wrapper logic in the base implementation. It makes insert and query operations use distinct code paths for equality checks. By introducing a new `AVAILABLE`equal result, all sentinel comparisons happened inside the equal wrapper and I believe this also fixes the subtle `erase` bug we try to target with #396.


sleeepyjack (approved):
I'm not 100% sold on the naming, e.g., the difference in meaning between `AVAILABLE` and `EMPTY` but I can't come up with a better name.

PR #442, opened 2024-02-25
Add set retrieve

PointKernel:
This PR adds host-bulk set retrieve APIs. For now, they use device `find` APIs to get matches since the benefit of creating a dedicated device `retrieve` is unclear.

It also adds a placeholder for an overload of `retrieve` that takes custom key_equal and hasher.


include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,193 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+                             cuco::pair<ProbeKey, Key>* buffer,
+                             AtomicT* counter,
+                             OutputIt1 probe_begin,
+                             OutputIt2 output_begin)
+{
+  auto i = block.thread_rank();
+  __shared__ Size offset;
+
+  if (i == 0) { offset = counter->fetch_add(buffer_size, cuda::std::memory_order_relaxed); }
+  block.sync();
```
wence-: *question*: Is this more future-proofly written as:

```
Size offset = cooperative_groups::invoke_one_broadcast(block, [&]() { return counter->fetch_add(buffer_size, cuda::std::memory_order_relaxed); });
```
?

Edit: no, because `invoke_one_broadcast` only works for `coalesced_group` and `thread_block_tile`, not general group types.

PointKernel: Good point.

I will do this in the tile-wise `flush_buffer`.

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+          typename Size,  // have to do this due to https://github.com/NVIDIA/cccl/issues/1523
+          typename ProbeKey,
+          typename Key,
+          typename AtomicT,
+          typename OutputIt1,
+          typename OutputIt2>
+__device__ void flush_buffer(CG const& tile,
+                             Size buffer_size,
+                             cuco::pair<ProbeKey, Key>* buffer,
+                             AtomicT* counter,
```
sleeepyjack:
```suggestion
                             cuda::atomic<Size, whateverthreadscope>& counter,
```

This way you can get rid of the `AtomicT` tparam and also make sure we pass a reference to a valid object instead of an arbitrary ptr.

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+  Size offset;
+  auto i = tile.thread_rank();
+
+#if defined(CUCO_HAS_CG_INVOKE_ONE)
+  offset = cooperative_groups::invoke_one_broadcast(
+    tile, [&]() { return counter->fetch_add(buffer_size, cuda::std::memory_order_relaxed); });
+#else
+  if (i == 0) { offset = counter->fetch_add(buffer_size, cuda::std::memory_order_relaxed); }
+  offset = tile.shfl(offset, 0);
+#endif
```
sleeepyjack: Is there a performance difference between the two? How is `invoke_one_broadcast` implemented?

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+  auto constexpr window_size = Ref::window_size;
+
+  auto idx          = cuco::detail::global_thread_id() / tile_size;
+  auto const stride = cuco::detail::grid_stride() / tile_size;
+  auto const block  = cg::this_thread_block();
+  auto const tile   = cg::tiled_partition<tile_size>(block);
+
+  auto constexpr flushing_tile_size = cuco::detail::warp_size() / window_size;
+  // random choice to tune
+  auto constexpr flushing_buffer_size = 2 * flushing_tile_size;
```
sleeepyjack, line 154: I'm curious. Why did you choose that particular size?

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+
+  auto const flushing_tile    = cg::tiled_partition<flushing_tile_size>(block);
+  auto const flushing_tile_id = block.thread_rank() / flushing_tile_size;
+
+  __shared__ cuco::pair<ProbeKey, Key> flushing_tile_buffer[num_flushing_tiles][flushing_tile_size];
+  // per flushing-tile counter to track number of filled elements
+  __shared__ cuda::atomic<size_type, cuda::thread_scope_block> flushing_counter[num_flushing_tiles];
+
+#if defined(CUCO_HAS_CG_INVOKE_ONE)
+  cg::invoke_one(flushing_tile, [&]() { flushing_counter[flushing_tile_id] = 0; });
```
sleeepyjack: Is this enough to initialize the atomic object inplace or do we need a placement-`new`?

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+    auto const active_flushing_tile =
+      cg::binary_partition<flushing_tile_size>(flushing_tile, active_flag);
+    if (active_flag) {
+      auto const found = ref.find(tile, *(first + idx));
+#if defined(CUCO_HAS_CG_INVOKE_ONE)
+      if (found != ref.end()) {
+        cg::invoke_one(tile, [&]() {
+          auto const offset =
+            flushing_counter[flushing_tile_id].fetch_add(1, cuda::std::memory_order_relaxed);
+          flushing_tile_buffer[flushing_tile_id][offset] = cuco::pair{*(first + idx), *found};
```
sleeepyjack:
```suggestion
          flushing_tile_buffer[flushing_tile_id][offset] = {*(first + idx), *found};
```

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+        });
+      }
+#else
+      if (tile.thread_rank() == 0 and found != ref.end()) {
+        auto const offset =
+          flushing_counter[flushing_tile_id].fetch_add(1, cuda::std::memory_order_relaxed);
+        flushing_tile_buffer[flushing_tile_id][offset] = cuco::pair{*(first + idx), *found};
```
sleeepyjack:
```suggestion
        flushing_tile_buffer[flushing_tile_id][offset] = {*(first + idx), *found};
```

Just a suggestion/my preference. Some find stating the type explicitly more readable.

include/cuco/detail/static_set/static_set.inl
```diff
@@ -307,6 +307,80 @@ void static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>
...
 
+template <class Key,
+          class Extent,
+          cuda::thread_scope Scope,
+          class KeyEqual,
+          class ProbingScheme,
+          class Allocator,
+          class Storage>
+template <typename InputIt, typename OutputIt1, typename OutputIt2>
+std::pair<OutputIt1, OutputIt2>
```
sleeepyjack:
```suggestion
cuda::std::pair<OutputIt1, OutputIt2>
```

include/cuco/detail/static_set/static_set.inl
```diff
@@ -307,6 +307,80 @@ void static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>
...
+
+template <class Key,
+          class Extent,
+          cuda::thread_scope Scope,
+          class KeyEqual,
+          class ProbingScheme,
+          class Allocator,
+          class Storage>
+template <typename InputIt, typename OutputIt1, typename OutputIt2>
+std::pair<OutputIt1, OutputIt2>
```
sleeepyjack:
```suggestion
cuda::std::pair<OutputIt1, OutputIt2>
```

Although I'm not sure if this is helpful or not

PointKernel: This is a host API thus returning a host device pair seems an overkill.

include/cuco/detail/static_set/kernels.cuh
PointKernel: > Is there a performance difference between the two?

No visible runtime difference but `invoke_one` indeed requires more instructions https://godbolt.org/z/8Y3eeT11j

include/cuco/detail/static_set/static_set.inl
sleeepyjack: Right, a plain `std::pair` will work here. My point is more about design guidelines. `cuda::std::` is a superset of `std::` as it works both on host and device. My understanding is that we should use `cuda::std::` as much as possible for this reason. Although I'm not sure if there's any automatic namespace aliasing `std:: -> cuda::std::` happening or not.

include/cuco/detail/static_set/kernels.cuh
sleeepyjack: Ha, if we `invoke_one` using a `thread_block_tile` instead of a whole CTA we actually emit fewer instructions. Notice the `ELECT` instruction: https://godbolt.org/z/q36dYEoTc

PointKernel: https://godbolt.org/z/G33zE5aGx

PointKernel: It saves more with `invoke_one_broadcast` since shared memory is no longer needed

PointKernel: OK, so for the new `invoke_one`/`invoke_one_broadcast` in general:

- tile works better than block
- newer arch like `sm_90` works better than old `sm_70`

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+
+  auto constexpr tile_size   = Ref::cg_size;
+  auto constexpr window_size = Ref::window_size;
+
+  auto idx          = cuco::detail::global_thread_id() / tile_size;
+  auto const stride = cuco::detail::grid_stride() / tile_size;
+  auto const block  = cg::this_thread_block();
+  auto const tile   = cg::tiled_partition<tile_size>(block);
+
+  auto constexpr flushing_tile_size = cuco::detail::warp_size() / window_size;
```
wence-, line 152: *suggestion*: `static_assert(flushing_tile_size > 0)` ? Unless it is asserted somewhere else (in the `Ref` constructor?) that the `window_size` parameter must always be at most `warp_size` large?

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+  // random choice to tune
+  auto constexpr flushing_buffer_size = 2 * flushing_tile_size;
+  auto constexpr num_flushing_tiles   = BlockSize / flushing_tile_size;
+  auto constexpr max_matches          = flushing_tile_size / tile_size;
+
+  auto const flushing_tile    = cg::tiled_partition<flushing_tile_size>(block);
+  auto const flushing_tile_id = block.thread_rank() / flushing_tile_size;
```
wence-:
```suggestion
  auto const flushing_tile_id = flushing_tile.meta_group_rank();
```

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+
+  while (flushing_tile.any(idx < n)) {
+    bool active_flag = idx < n;
+    auto const active_flushing_tile =
+      cg::binary_partition<flushing_tile_size>(flushing_tile, active_flag);
+    if (active_flag) {
+      auto const found = ref.find(tile, *(first + idx));
+#if defined(CUCO_HAS_CG_INVOKE_ONE)
+      if (found != ref.end()) {
+        cg::invoke_one(tile, [&]() {
```
wence-, line 187: *question*: `invoke_one` is logically collective over the group defined by `tile` and the hardware could select any thread in `[0, tile.num_threads())` to execute the functor. However, it seems to me that not all threads in `tile` could reach this line (because both `found` and `active_flag` are divergent to my understanding). Is this a problem?

include/cuco/detail/static_set/kernels.cuh
```diff
@@ -89,6 +90,257 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
...
+  auto i = block.thread_rank();
+  __shared__ Size offset;
+
+#if defined(CUCO_HAS_CG_INVOKE_ONE)
+  cooperative_groups::invoke_one(
+    block, [&]() { offset = counter->fetch_add(buffer_size, cuda::std::memory_order_relaxed); });
+#else
+  if (i == 0) { offset = counter->fetch_add(buffer_size, cuda::std::memory_order_relaxed); }
+#endif
+  block.sync();
```
wence-, line 256: *question*: In the CG `invoke_one` case, is this better written without the explicit `__shared__` offset as:
```
#if defined(CUCO_HAS_CG_INVOKE_ONE)
  Size offset = cg::invoke_one_broadcast(block, [&] { return counter->fetch_add(buffer_size, cuda::std::memory_order_relaxed) });
#else
  __shared__ Size offset;
  if (i == 0) { offset = counter->fetch_add(buffer_size, cuda::std::memory_order_relaxed); }
  block.sync()
#endif
```
?

include/cuco/detail/static_set/static_set.inl
PointKernel: > cuda::std:: is a superset of std:: as it works both on host and device.

That's exactly the reason I tent to use, e.g.  `cuda::std::pair` for host device API, `thrust::pair` (or is there a `cuda::pair`?) for device API and `std::pair` for host only.

PointKernel: Switching to `cuda::std::pair` for all APIs based on offline discussions

include/cuco/detail/static_set/kernels.cuh
PointKernel: No particular reason. Tested with 1, 2, 3 and 4 and there is no big difference between those options.

include/cuco/detail/static_set/kernels.cuh
PointKernel: Thanks for explaining your concern. The tile-based `ref.find(tile, ...)` guarantees that all threads of the same tile have the same `found`. `active_flag` could diverge between different tiles but not for threads of the same tile.

include/cuco/detail/static_set/kernels.cuh
PointKernel: Nice! Really love your comments like this. Please review more cuco PRs :)

include/cuco/detail/static_set/kernels.cuh
PointKernel: I see your point. `cg::invoke_one_broadcast` only works for tiles but not thread block thus it doesn't work in this particular case. However, your suggestion is valid for numerous other cases in cuco and I will make a PR to update them all. :+1: Love it.

include/cuco/detail/static_set/kernels.cuh
wence-: Thanks!

include/cuco/static_set.cuh
```diff
@@ -533,6 +534,108 @@ class static_set {
...
+  /**
+   * @brief Asynchronously retrieves the matched key in the set corresponding to all probe keys in
+   * the range `[first, last)`
+   *
+   * If key `k = *(first + i)` has a match `m` in the set, copies a `cuco::pair{k, m}` to
+   * unspecified locations in `[output_begin, output_end)`. Else, does nothing.
+   *
+   * @note Behavior is undefined if the size of the output range exceeds
+   * `std::distance(output_begin, output_end)`.
+   * @note Behavior is undefined if the given key has multiple matches in the set.
```
sleeepyjack, line 580: Is it undefined or do we return the first matching occurrence of the key?

include/cuco/detail/static_set/static_set.inl
```diff
@@ -307,6 +307,80 @@ void static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>
...
+OutputIt
+static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::retrieve_async(
+  InputIt first,
+  InputIt last,
+  OutputIt output_begin,
+  ProbeEqual const& probe_equal,
+  ProbeHash const& probe_hash,
+  cuda_stream_ref stream) const
+{
+  CUCO_FAIL("Unsupported code path: retrieve_async with custom hash/equal");
```
sleeepyjack, line 381: We should add a note about this in the inline docs

include/cuco/static_set.cuh
PointKernel: It's the first element for scalar probing but undefined behavior for CG-based algorithms so undefined behavior is accurate.

include/cuco/detail/static_set/static_set.inl
PointKernel: [1cb1b58](https://github.com/NVIDIA/cuCollections/pull/442/commits/1cb1b58edc671c9a5533495c864f61319dcbc8bc)

sleeepyjack (approved):
Awesome work! Thanks!

PR #443, opened 2024-03-08
Add mapping table example

PointKernel:
This PR adds a mapping table example demonstrating how to handle use cases where keys are larger than 8 bytes or it's impossible to determine sentinel values.


examples/static_set/mapping_table_example.cu
```diff
@@ -0,0 +1,154 @@
...
+   */
+  __device__ constexpr bool operator()(int32_t lhs, int32_t rhs) const
+  {
+    // Check all 4 elements of a tuple to determine if two tuples are identical
+    return cuda::std::get<0>(_data[lhs]) == cuda::std::get<0>(_data[rhs]) and
+           cuda::std::get<1>(_data[lhs]) == cuda::std::get<1>(_data[rhs]) and
+           cuda::std::get<2>(_data[lhs]) == cuda::std::get<2>(_data[rhs]) and
+           cuda::std::get<3>(_data[lhs]) == cuda::std::get<3>(_data[rhs]);
+  }
+  T const* _data;
```
sleeepyjack, line 79: Should we consider using `cuda::std::span` here?

examples/static_set/mapping_table_example.cu
```diff
@@ -0,0 +1,154 @@
...
+#include <thrust/device_vector.h>
+
+#include <iostream>
+
+/**
+ * @file mapping_table_example.cu
+ *
+ * @brief Demonstrates how to use hash set as a lookup table of the original data
+ *
+ * `cuco` hash tables such as `cuco::static_set` or `cuco::static_map` currently support only 4/8
```
sleeepyjack: I'd say supporting only native typed keys is by-design rather than a missing feature.

sleeepyjack (approved):

examples/static_set/mapping_table_example.cu
PointKernel: Using pointer for demonstration purposes since not everyone is familiar with the span concept

examples/static_set/mapping_table_example.cu
PointKernel: Updated the doc a bit based on offline discussions.

PR #446, opened 2024-03-18
Clean up strong type utility

PointKernel:
This PR removes the inline sentinel namespace and moves the strong type definition to its own header file.

It's the first step to clean up multiple unrelated definitions in the same `detail/utils.hpp/cuh` file.

### DIFF
--- a/include/cuco/detail/utility/strong_type.cuh
+++ b/include/cuco/detail/utility/strong_type.cuh
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2021-2024, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


include/cuco/detail/utility/strong_type.cuh
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ */
+
+#pragma once
+
+namespace cuco::detail {
```
sleeepyjack, line 18:
```suggestion
namespace cuco::utility {
```

The utility namespace is considered public so we don't want `detail` here.

PointKernel: This is a detail utility and we don't want to expose `strong_type` to the public. I can change to `cuco::utility::detail` as well but it sounds wordy.

sleeepyjack: Hmm, I tend to say `detail` namespace stuff should go into the `include/cuco/detail` directory if it's not intended to be used by a user. On the other hand, it's debatable if `strong_type` should be exposed or not. It could live in a public namespace. I'm okay with both versions.

PointKernel: > detail namespace stuff should go into the include/cuco/detail directory 

it is in `include/cuco/detail/utility`.

>  it's debatable if strong_type should be exposed or not. 

I don't know. When do users need to use `strong_type` directly?

sleeepyjack (approved):

PR #447, opened 2024-03-18
Clean up pair utilities

PointKernel:
This PR moves pair utilities to their own files.

### DIFF
--- a/include/cuco/detail/pair/traits.hpp
+++ b/include/cuco/detail/pair/traits.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


include/cuco/detail/pair/traits.hpp
```diff
@@ -15,11 +15,11 @@
...
 #pragma once
 
+#include <cuda/std/tuple>
+#include <cuda/std/type_traits>
 #include <thrust/device_reference.h>
 #include <thrust/tuple.h>
 
-#include <cuda/std/tuple>
-#include <cuda/std/type_traits>
 #include <tuple>
```
sleeepyjack, line 23: Not sure if we should group `cuda::std` facilities with `thrust` rather than STL.

PointKernel: We are using a new convention in cudf https://github.com/rapidsai/cudf/pull/15063.

The idea is that thrust/cub/libcu++ all belongs to CCCL thus they are as one single header group.

sleeepyjack: That makes alot of sense. I stand corrected.

sleeepyjack (approved):

PR #453, opened 2024-04-05
Fix illegal memory issues in multimap benchmarks

PointKernel:
This PR fixes bugs in the multimap benchmarks where the corresponding count operation should have determined the output size.


sleeepyjack (approved):
Good catch! 🎣

sleeepyjack:
Please update the copyright year of the changed files. Apart from that LGTM.

PR #454, opened 2024-04-11
Fix build errors in the multimap godbolt sample

PointKernel:
Update the multimap godbolt sample link to fix a breaking change leftover.


sleeepyjack (approved):
Thanks! At some point we need an automated way to generate these godbolt links to get rid of these kind of artifacts.

PointKernel:
The example is correct, only godbolt link is outdated

PR #456, opened 2024-04-12
Get rid of duplicate logic in the OA base implementations

PointKernel:
When testing the shared memory hash table optimization for set `contains`, I realized several duplicates exist in the OA base implementations.

This PR removes the redundant code by using one single code path for _certain_op_ and its _certain_op_if_ variant.


sleeepyjack (approved):
Can you verify that the constant `true` actually boils away the predicate logic in the SASS? I think looking at the benchmark diff should be enough. Apart from that LGTM.

PointKernel:
> Can you verify that the constant true actually boils away the predicate logic in the SASS?

Not really boil away but the runtime difference is about 0.5% or smaller when I tested it last time.

PR #458, opened 2024-04-15
Add shared memory contains

PointKernel:
TBD

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -456,13 +456,11 @@ class open_addressing_impl {
 
     if (bytes >= 48000) {
       open_addressing_ns::detail::contains_if_n<false, cg_size, cuco::detail::default_block_size()>
-        <<<grid_size,
-           cuco::detail::default_block_size(),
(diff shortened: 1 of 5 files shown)


include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -445,9 +445,24 @@ class open_addressing_impl {
...
     auto const always_true = thrust::constant_iterator<bool>{true};
-    this->contains_if_async(
-      first, last, always_true, thrust::identity{}, output_begin, container_ref, stream);
+
+    auto const bytes =
+      sizeof(typename storage_ref_type::window_type) * this->storage_.num_windows();
+
+    if (bytes >= 48000) {
+      open_addressing_ns::detail::contains_if_n<false, cg_size, cuco::detail::default_block_size()>
+        <<<grid_size, cuco::detail::default_block_size(), 0, stream>>>(
```
sleeepyjack, line 459: Here we specify `cuco::detail::default_block_size()` which is 128 but earlier we used 1024 to calculate the grid size.

PointKernel:
Closing this PR since the feature is probably no longer desired

Issue #462, opened 2024-04-17
[FEATURE]: Add multiset host-bulk count APIs


PointKernel:
### Is your feature request related to a problem? Please describe.

Add multiset host-bulk count APIs

### Describe the solution you'd like

**The basic API must have:**
```cuda
  /**
   * @brief Counts the occurrences of keys in `[first, last)` contained in the multiset.
   *
   * @tparam Input Device accessible input iterator
   * 
   * @param first Beginning of the sequence of keys to count
   * @param last End of the sequence of keys to count
   * @param stream CUDA stream used for count
   * 
   * @return The sum of total occurrences of all keys in `[first, last)`
   */
  template <typename InputIt>
  size_type count(InputIt first,
                  InputIt last,
                  cuda_stream_ref stream = {}) const;
```
_Note: There is no restriction for the iterator `value_type` due to heterogeneous lookup support._

**More specifically, libcudf requires the below `count` variants** 
```cuda
  /**
   * @brief Counts the occurrences of keys in `[first, last)` contained in the multiset.
   *
   * @tparam Input Device accessible input iterator
   * @tparam ProbeKeyEqual Binary callable
   * @tparam ProbeHash Unary hash callable
   * 
   * @param first Beginning of the sequence of keys to count
   * @param last End of the sequence of keys to count
   * @param probe_key_equal Binary function to compare two keys for equality
   * @param probe_hash Unary callable to hash a given key
   * @param stream CUDA stream used for count
   * 
   * @return The sum of total occurrences of all keys in `[first, last)`
   */
  template <typename InputIt, typename ProbeKeyEqual, typename ProbeHash>
  size_type count(InputIt first,
                  InputIt last,
                  ProbeKeyEqual probe_key_equal,
                  ProbeHash probe_hash, 
                  cuda_stream_ref stream = {}) const;

  /**
   * @brief Counts the occurrences of keys in `[first, last)` contained in the multiset.
   * 
   * @note: If a given key has no matches, its occurrence is 1.
   * 
   * @tparam Input Device accessible input iterator
   * @tparam ProbeKeyEqual Binary callable
   * @tparam ProbeHash Unary hash callable
   * 
   * @param first Beginning of the sequence of keys to count
   * @param last End of the sequence of keys to count
   * @param probe_key_equal Binary function to compare two keys for equality
   * @param probe_hash Unary callable to hash a given key
   * @param stream CUDA stream used for count
   * 
[...]

PR #472, opened 2024-04-24
Add multiset count APIs

PointKernel:
Contributes to #462

This PR adds a `multiset::count` host-bulk API.

### DIFF
--- a/include/cuco/detail/open_addressing/kernels.cuh
+++ b/include/cuco/detail/open_addressing/kernels.cuh
@@ -292,16 +292,14 @@ CUCO_KERNEL void count(InputIt first, cuco::detail::index_type n, AtomicT* count
   auto idx               = cuco::detail::global_thread_id() / CGSize;
 
   while (idx < n) {
-    typename std::iterator_traits<InputIt>::value_type const& insert_element{*(first + idx)};
-    /*
+    auto const key = *(first + idx);
     if constexpr (CGSize == 1) {
-      if (ref.insert(insert_element)) { thread_num_successes++; };
+      thread_count += ref.count(key);
     } else {
       auto const tile =
         cooperative_groups::tiled_partition<CGSize>(cooperative_groups::this_thread_block());
-      if (ref.insert(tile, insert_element) && tile.thread_rank() == 0) { thread_num_successes++; }
+      thread_count += ref.count(tile, key);
     }
-    */
     idx += loop_stride;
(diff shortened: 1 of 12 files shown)


PointKernel:
`cg::reduce` can generate more effective SASS but got an illegal memory access error when trying to read the final global output. To be investigated.

include/cuco/detail/open_addressing/kernels.cuh
```diff
@@ -329,6 +329,60 @@ CUCO_KERNEL void find(InputIt first, cuco::detail::index_type n, OutputIt output
   }
 }
 
+/**
+ * @brief Inserts all elements in the range `[first, first + n)` and returns the number of
+ * successful insertions if `pred` of the corresponding stencil returns true.
+ *
+ * @note If multiple elements in `[first, first + n)` compare equal, it is unspecified which element
+ * is inserted.
+ * @note The key `*(first + i)` is inserted if `pred( *(stencil + i) )` returns true.
```
sleeepyjack: This docs relate to `insert` rather than `count`

include/cuco/operator.hpp
```diff
@@ -50,6 +50,12 @@ struct erase_tag {
 struct contains_tag {
 } inline constexpr contains;  ///< `cuco::contains` operator
 
+/**
+ * @brief `count` operator tag
+ */
+struct count_tag {
```
sleeepyjack, line 56: Nit: Should we rename this to `count/find/*_operator` in a future PR?

include/cuco/detail/open_addressing/kernels.cuh
PointKernel: updated

include/cuco/operator.hpp
PointKernel: Are you suggesting to replace `*_tag` with `*_operator`?

sleeepyjack: Yeah, I find my intial choice with "_tag" might not be ideal

sleeepyjack: If course that's just a general remark - no action item for this PR

PointKernel: It's like the `empty_value` v.s. `empty_payload` discussion we just had. `*_tag` looks fine to me.

sleeepyjack (approved):

PR #473, opened 2024-05-03
Ensure that `tuple_size` is defined

miscco:
Some compilers have issues trying to instantiate it too early. We can work around this by adding another SFINAE condition.


miscco (commented):
This came up when building cuDF with latest CCCL ToT

PointKernel (approved):
LGTM. Thanks for the fix!

PR #478, opened 2024-05-07
Fix dynamic init warning for cuda barrier

PointKernel:
This PR silences the cuda barrier dynamic initialization warning by using the proper pragma.

The warning comes up when building with nvcc 12.4 but not seen with 12.3 or earlier.

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
@@ -302,9 +302,11 @@ class open_addressing_ref_impl {
   {
     auto const num_windows = static_cast<size_type>(this->window_extent());
 #if defined(CUCO_HAS_CUDA_BARRIER)
+#pragma nv_diagnostic push
 // Disables `barrier` initialization warning.
 #pragma nv_diag_suppress static_var_with_dynamic_init
     __shared__ cuda::barrier<cuda::thread_scope::thread_scope_block> barrier;
+#pragma nv_diagnostic pop
     if (g.thread_rank() == 0) { init(&barrier, g.size()); }
     g.sync();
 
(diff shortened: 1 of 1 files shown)


sleeepyjack:
Is this because the barrier internally uses dynamic shmem initialization or because we dynamically initialize a barrier in shmem? The latter can be solved through placement-new.

PointKernel:
This is a [known issue](https://nvidia.github.io/cccl/libcudacxx/extended_api/synchronization_primitives/barrier/init.html) for `cuda::barrier` and we are using the suggested `init` function to initialize the barrier. To clarify, I don't have this issue locally with 12.3 but got it when I built the code with devcontainer. Awkward enough, I don't see the warnings in CI either.

PointKernel:
So there is a similar issue here:
https://github.com/NVIDIA/cuCollections/actions/runs/8992496856/job/24702405011?pr=479#step:7:368

But not in https://github.com/NVIDIA/cuCollections/actions/runs/8992496856/job/24702405167?pr=479#step:7:371

What's changed between those two build tests?

sleeepyjack:
> What's changed between those two build tests?

The first link is an x86 build while the second link is an arm64 build. Not sure why only one of them triggers the warning.

> This is a [known issue](https://nvidia.github.io/cccl/libcudacxx/extended_api/synchronization_primitives/barrier/init.html) for cuda::barrier and we are using the suggested init function to initialize the barrier.

Ah, I see. My only concern is that when we disable the warning right after declaring the macro, then it will be disabled for the entire project. Instead, could we just selectively disable the line that performs the dynamic initialization and then re-enable the warning again?

PointKernel:
> could we just selectively disable the line that performs the dynamic initialization and then re-enable the warning again?

Good point. Done.

sleeepyjack (approved):
LGTM 🚀 Let's merge that puppy!

PR #479, opened 2024-05-07
Fix key equality check order

### LINKED ISSUE #474: [ENHANCEMENT]: Place the existing key to the right-hand side during equality checks

cuco hash tables always place the slot key on the left-hand side for key equality checks:
https://github.com/NVIDIA/cuCollections/blob/6cb6dbfe13b10109f74f3b5bedbe38f8c0eed687/include/cuco/static_map.cuh#L64-L66

This was a completely random choice when I started the open-addressing refactoring and I thought it didn't matter and was wrong.

Generally speaking, when we want to check if two variables are identical, we put the query value on the left-hand side and the "reference" or the existing value on the right-hand side. e.g. we do
```cpp
if (idx == 0) { ... }
```
instead of 
```cpp
if (0 == idx) { ... }
```

**_The new cuco data structures are unfortunately following the latter pattern._**

This works fine until we meet the hash join use case where the build table is the right table and the probe table is the left table. As a result, the left table is always on the right when doing comparisons in cuco while the right table is always on the left. In many places across libcudf, `build_table`/`right_table` and `probe_table`/`left_table` are interchangeable terms thus for a function `join_func` expecting the first argument to be the build table and the second argument to be the probe table, we may have to invoke it awkwardly:
```cpp
void join_func(right_table, left_table, ...)
```

This must be stopped.

Always place the existing value (either the sentinel value or the slot key) on the right side for equality checks.

e.g.
https://github.com/NVIDIA/cuCollections/blob/6cb6dbfe13b10109f74f3b5bedbe38f8c0eed687/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh#L370

We should do the following instead
```cpp
this->predicate_.operator()<is_insert::YES>(key, this->extract_key(slot_content));
```

PointKernel:
Closes #474 

This PR makes the reference value always the right-hand side for key equality checks.

The updates for heterogeneous lookup tests indicate that it will be probably a breaking change for libcudf byte pair encoding.

### DIFF
--- a/include/cuco/detail/equal_wrapper.cuh
+++ b/include/cuco/detail/equal_wrapper.cuh
@@ -78,9 +78,9 @@ struct equal_wrapper {
   /**
    * @brief Order-sensitive equality operator.
    *
-   * @note This function always compares the left-hand side element against sentinel values first
+   * @note This function always compares the right-hand side element against sentinel values first
    * then performs a equality check with the given `equal_` callable, i.e., `equal_(lhs, rhs)`.
-   * @note Container (like set or map) keys MUST be always on the left-hand side.
+   * @note Container (like set or map) buckets MUST be always on the right-hand side.
    *
    * @tparam IsInsert Flag indicating whether it's an insert equality check or not. Insert probing
    * stops when it's an empty or erased slot while query probing stops only when it's empty.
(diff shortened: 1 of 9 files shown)


include/cuco/detail/equal_wrapper.cuh
```diff
@@ -78,9 +78,9 @@ struct equal_wrapper {
   /**
    * @brief Order-sensitive equality operator.
    *
-   * @note This function always compares the left-hand side element against sentinel values first
+   * @note This function always compares the right-hand side element against sentinel values first
    * then performs a equality check with the given `equal_` callable, i.e., `equal_(lhs, rhs)`.
-   * @note Container (like set or map) keys MUST be always on the left-hand side.
+   * @note Container (like set or map) buckets MUST be always on the right-hand side.
```
sleeepyjack, line 83: buckets? slots?

sleeepyjack (approved):
LGTM. Can be merged once CI passes.

include/cuco/detail/equal_wrapper.cuh
PointKernel: you are the expert here :)
Is bucket a more proper term than slot in this case?

sleeepyjack: Open-addressing means buckets consist of a single table slot, so it's interchangeable. Let's leave it as is.

PR #482, opened 2024-05-14
Add ref with_operators

PointKernel:
Address https://github.com/NVIDIA/cuCollections/pull/467#discussion_r1580274993

This PR adds new `with_operators` ref APIs. They replace the legacy `with` APIs without moving the old ref object. The legacy APIs will be deprecated once all RAPIDS projects use the new APIs.

### DIFF
--- a/examples/static_set/shared_memory_example.cu
+++ b/examples/static_set/shared_memory_example.cu
@@ -60,6 +60,7 @@ __global__ void shmem_set_kernel(typename SetRef::extent_type window_extent,
   // Next, we want to check if the keys can be found again using the `contains` function. We create
   // a new non-owning object based on the `insert_ref` that supports `contains` but no longer
   // supports `insert`.
+  // CAVEAT: concurrent use of `insert_ref` and `contains_ref` is undefined behavior.
   auto const contains_ref = insert_ref.with_operators(cuco::contains);
 
   // Check if all keys can be found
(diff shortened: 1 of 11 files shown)


examples/static_set/shared_memory_example.cu
```diff
@@ -45,12 +45,9 @@ __global__ void shmem_set_kernel(typename SetRef::extent_type window_extent,
...
 
-  // The `set` object does not come with any functionality.
-  // We first have to transform it into an object that supports the function we need
-  // (in this case `insert`). This invalidates the old `set` object (move semantics).
-  // With this guard we can make sure that users don't mix operations that are not
-  // safe to use in a concurrent setup.
-  auto insert_ref = std::move(set).with(cuco::insert);
+  // The `set` object does not come with any functionality. We first have to transform it into an
+  // object that supports the function we need (in this case `insert`).
+  auto insert_ref = set.with_operators(cuco::insert);
```
sleeepyjack, line 50: In this example, we should still emphasize that concurrent use of refs with different configurations is UB.

include/cuco/static_set_ref.cuh
```diff
@@ -198,6 +198,22 @@ class static_set_ref
   template <typename... NewOperators>
   [[nodiscard]] __host__ __device__ auto with(NewOperators... ops) && noexcept;
```
sleeepyjack, line 201: There's a [\deprecated](https://www.doxygen.nl/manual/commands.html#cmddeprecated) keyword in Doxygen which we can add to the old function to inform users that it will be deleted soon.

sleeepyjack (approved):
LGTM
Very nice!

PR #470, opened 2024-04-23
Add multiset find APIs

PointKernel:
Closes #464 

This PR adds multiset host-bulk and device-singular find APIs

PR #483, opened 2024-05-15
Unify host-bulk find implementations in OA base class

PointKernel:
Address https://github.com/NVIDIA/cuCollections/pull/470#discussion_r1600151940

This PR unifies the host-bulk `find` in one place and removes duplicates. It turns out that we don't need the SFINAE helper to get the output type https://godbolt.org/z/aPo6xb3cW.

### DIFF
--- a/include/cuco/detail/open_addressing/kernels.cuh
+++ b/include/cuco/detail/open_addressing/kernels.cuh
@@ -18,15 +18,14 @@
 #include <cuco/detail/utility/cuda.cuh>
 
 #include <cub/block/block_reduce.cuh>
-
 #include <cuda/atomic>
+#include <cuda/functional>
 
 #include <cooperative_groups.h>
 
 #include <iterator>
 
-namespace cuco {
-namespace detail {
+namespace cuco::detail {
 CUCO_SUPPRESS_KERNEL_WARNINGS
 
 /**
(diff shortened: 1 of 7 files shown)


include/cuco/detail/open_addressing/kernels.cuh
```diff
@@ -255,6 +254,81 @@ CUCO_KERNEL void contains_if_n(InputIt first,
...
+
+  auto const block       = cg::this_thread_block();
+  auto const thread_idx  = block.thread_rank();
+  auto const loop_stride = cuco::detail::grid_stride() / CGSize;
+  auto idx               = cuco::detail::global_thread_id() / CGSize;
+
+  using output_type = typename std::iterator_traits<OutputIt>::value_type;
+  __shared__ output_type output_buffer[BlockSize / CGSize];
+
+  auto constexpr has_payload = not std::is_same_v<typename Ref::key_type, typename Ref::value_type>;
```
sleeepyjack, line 289: The same utility is defined [here](https://github.com/NVIDIA/cuCollections/blob/fb3d490ec8d564eba694d9919db92dbc50b635e8/include/cuco/detail/open_addressing/open_addressing_impl.cuh#L85) but is not accessible in this context. Maybe we should move it to a standalone facility so it can be used throughout the library.

PointKernel: Something like you pass a template container type and it will return whether or not it's map? Yeah, but this's a one-line expression so I'm not sure how much effort it could save.

sleeepyjack: Correct. I can see that it's not super useful. I will need the same mechanism to unify the retrieve operator, too. But then again it's just a single line of code. Lets leave it as is then.

sleeepyjack (approved):
LGTM
Nice work!

PR #486, opened 2024-05-16
Use empty value sentinel to improve readability

PointKernel:
This PR replaces the old `empty_slot_sentinel.second` expressions with `empty_value_sentinel()`.

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
@@ -513,7 +513,7 @@ class open_addressing_ref_impl {
         if (eq_res == detail::equal_result::EQUAL) {
           if constexpr (has_payload) {
             // wait to ensure that the write to the value part also took place
-            this->wait_for_payload((window_ptr + i)->second, this->empty_slot_sentinel_.second);
+            this->wait_for_payload((window_ptr + i)->second, this->empty_value_sentinel());
           }
           return {iterator{&window_ptr[i]}, false};
         }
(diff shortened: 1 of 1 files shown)


sleeepyjack:
Would `empty_payload_sentinel` resolve the old confusion between the terms `value_type` and `payload/mapped_type`?

PointKernel:
> Would `empty_payload_sentinel` resolve the old confusion between the terms `value_type` and `payload/mapped_type`?

It's a better name in all senses but it's also a breaking change. This is the only concern preventing me from doing this. What do you think?

PointKernel:
I was thinking of making `empty_value` a template alias of the new `empty_payload` but CTAD will no longer work until C++ 20 (https://godbolt.org/z/qzPzq33bc) And libcudf heavily uses CTAD for sentinel constructions.

PointKernel:
OK. The current could be merged as it is and I will open another PR to add the new `empty_payload` strong type. To avoid breaking changes, I plan to add a ctor overload for the new `static_map` taking `empty_payload` argument.

sleeepyjack:
> To avoid breaking changes, I plan to add a ctor overload for the new static_map taking empty_payload argument.

Do we plan to remove `empty_value` in the future?

sleeepyjack:
> The current could be merged as it is and I will open another PR to add the new empty_payload strong type. 

We should rename `map.empty_value_sentinel()` -> `map.empty_payload_sentinel()` then.

sleeepyjack (approved):

PointKernel:
Based on the offline discussion, replacing `empty_value` with `empty_payload` is nice but unnecessary. Let's leave `empty_value` as it is.

PR #490, opened 2024-05-22
Add multiset custom count APIs

### LINKED ISSUE #462: [FEATURE]: Add multiset host-bulk count APIs

### LINKED ISSUE #488: [FEATURE]: Add `static_multiset::count_outer`

`static_multiset` requires an equivalent function to `static_multimap::count_outer`.

After #462 is finished, add a new API `count_outer` and also add a template parameter to the kernel to switch between `count` and `count_outer` behavior.

Still an open discussion if this is the best approach to tackle this problem but we couldn't come up with a better solution so far.

### Additional context

_No response_

PointKernel:
Closes #462, closes #488

This PR adds custom `count` APIs desired by libcudf hash join.


include/cuco/detail/open_addressing/kernels.cuh
```diff
@@ -354,13 +362,21 @@ CUCO_KERNEL void count(InputIt first, cuco::detail::index_type n, AtomicT* count
   auto idx               = cuco::detail::global_thread_id() / CGSize;
 
   while (idx < n) {
-    auto const key = *(first + idx);
+    typename std::iterator_traits<InputIt>::value_type const& key = *(first + idx);
```
PointKernel, line 370: caveats: `auto const` should **_not_** be used in any kernels when dereferencing input since it cannot deal with thrust wrappers properly.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -544,22 +545,34 @@ class open_addressing_impl {
   [[nodiscard]] size_type count(InputIt first,
                                 InputIt last,
                                 Ref container_ref,
-                                cuda_stream_ref stream = {}) const noexcept
```
PointKernel, line 545: note to ourselves: no default parameters for detail APIs

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -809,7 +809,7 @@ class open_addressing_ref_impl {
   /**
    * @brief Counts the occurrence of a given key contained in the container
    *
-   * @tparam ProbeKey Input type
+   * @tparam ProbeKey Prob key type
```
```suggestion
   * @tparam ProbeKey Probe key type
```

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -844,7 +844,7 @@ class open_addressing_ref_impl {
   /**
    * @brief Counts the occurrence of a given key contained in the container
    *
-   * @tparam ProbeKey Input type which is convertible to 'key_type'
+   * @tparam ProbeKey Prob key type
```
```suggestion
   * @tparam ProbeKey Probe key type
```

include/cuco/detail/static_multiset/static_multiset_ref.inl
```diff
@@ -499,6 +499,27 @@ class operator_impl<
...
+   *
+   * @tparam ProbeKey Probe key type
+   *
+   * @param group The Cooperative Group used to perform group count
+   * @param key The key to count for
+   *
+   * @return Number of occurrences found by the current thread
+   */
+  template <typename ProbeKey>
+  __device__ size_type count_outer(cooperative_groups::thread_block_tile<cg_size> const& group,
```
sleeepyjack: I see there is a `count_outer` which takes a CG tile but no single-thread version. Is this intentional?

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -814,6 +827,43 @@ class open_addressing_impl {
...
+                                cuda_stream_ref stream) const noexcept
+  {
+    auto const num_keys = cuco::detail::distance(first, last);
+    if (num_keys == 0) { return 0; }
+
+    auto counter =
+      detail::counter_storage<size_type, thread_scope, allocator_type>{this->allocator()};
+    counter.reset(stream);
+
+    auto const grid_size = cuco::detail::grid_size(num_keys, cg_size);
```
sleeepyjack, line 856: Since `count` and `count_outer` are basically reductions, using a grid stride loop might help with performance since it reduces the number of global updates. Not an action item for this PR, but I think it would be helpful to introduce a facility that calculates the optimal grid size based on the given SM count.

include/cuco/detail/static_multiset/static_multiset.inl
```diff
@@ -289,6 +289,51 @@ static_multiset<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>
...
+static_multiset<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::count(
+  InputIt first,
+  InputIt last,
+  ProbeKeyEqual const& probe_key_equal,
+  ProbeHash const& probe_hash,
+  cuda_stream_ref stream) const noexcept
+{
+  return impl_->count(first,
+                      last,
+                      ref(op::count).with_key_eq(probe_key_equal).with_hash_function(probe_hash),
```
sleeepyjack, line 310: This is nice!

include/cuco/detail/static_multiset/static_multiset_ref.inl
```diff
@@ -499,6 +499,27 @@ class operator_impl<
     auto const& ref_ = static_cast<ref_type const&>(*this);
     return ref_.impl_.count(group, key);
   }
+
+  /**
+   * @brief Counts the occurrence of a given key contained in multiset
```
sleeepyjack:
```suggestion
   * @brief Counts the occurrence of a given key contained in the multiset
```

include/cuco/detail/static_multiset/static_multiset_ref.inl
PointKernel: Good point. I could probably do the "outer check" in the kernel directly so the `count_outer` device API is not required. In general, I feel the `count_outer` APIs wouldn't be used anywhere outside of libcudf so I wanted to expose a host-bulk API only.

sleeepyjack: Also, I think users can always emulate the `count_outer` behavior at the call site through `max(set.count(key), 1)`. The CG version needs a bit more thought but should work similarly.

PointKernel: Done.

Also compared two ways of "outer check" and they generate exactly the same SASS
```cuda
    if (tile.all(thread_sum == 0) and tile.thread_rank() == 0) {
        ++thread_sum;
    }
```
v.s.
```cuda
    if (tile.all(thread_sum == 0)) {
        thread_sum += static_cast<int>(tile.thread_rank() == 0);
    }
```
https://godbolt.org/z/sf7YPdzqa

include/cuco/detail/open_addressing/open_addressing_impl.cuh
PointKernel: Depending on how soon `cuda::launch` (https://github.com/NVIDIA/cccl/issues/1646) will be ready, I wanted to avoid temporary effort since eventually they will all be replaced by `cuda::launch`. Any preferences?

PointKernel: OK. Probably not anytime soon based on the offline discussion. Let me test it a bit.

sleeepyjack: `cuda::launch` is of course preferred but given that it might take another major version or so to be feature complete we might want to fall back to using the ol' driver API. I would say the only kernels that would benefit from a smaller grid in cuco are those that flush staging buffers from shmem to gmem or do some sort of reduction. So this would be a good place to start the investigation.

sleeepyjack (approved):
👏 Nice work!

PointKernel:
Added the max occupancy grid size utility at [506228b](https://github.com/NVIDIA/cuCollections/pull/490/commits/506228bc4458eca34aece075e5b26ad0fc465cad).

# static_multiset_count_uniform_occupancy

## [0] Quadro RTX 8000

|  Key  |  Distribution  |  Occupancy  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |      Diff |   %Diff |  Status  |
|-------|----------------|-------------|------------|-------------|------------|-------------|-----------|---------|----------|
|  I32  |    UNIFORM     |     0.1     |  39.571 ms |       0.04% |  49.468 ms |       0.08% |  9.898 ms |  25.01% |   FAIL   |
|  I32  |    UNIFORM     |     0.2     |  42.254 ms |       0.03% |  55.027 ms |       0.10% | 12.773 ms |  30.23% |   FAIL   |
|  I32  |    UNIFORM     |     0.3     |  45.503 ms |       0.02% |  59.976 ms |       0.04% | 14.474 ms |  31.81% |   FAIL   |
|  I32  |    UNIFORM     |     0.4     |  49.558 ms |       0.01% |  65.085 ms |       0.20% | 15.527 ms |  31.33% |   FAIL   |
|  I32  |    UNIFORM     |     0.5     |  54.977 ms |       0.01% |  71.128 ms |       0.07% | 16.152 ms |  29.38% |   FAIL   |
|  I32  |    UNIFORM     |     0.6     |  62.797 ms |       1.11% |  79.247 ms |       0.05% | 16.450 ms |  26.20% |   FAIL   |
|  I32  |    UNIFORM     |     0.7     |  73.613 ms |       0.01% |  90.862 ms |       0.06% | 17.249 ms |  23.43% |   FAIL   |
|  I32  |    UNIFORM     |     0.8     |  93.233 ms |       0.12% | 111.622 ms |       0.04% | 18.389 ms |  19.72% |   FAIL   |
|  I32  |    UNIFORM     |     0.9     | 146.283 ms |       0.01% | 166.684 ms |       0.08% | 20.401 ms |  13.95% |   FAIL   |
|  I64  |    UNIFORM     |     0.1     |  40.903 ms |       0.04% |  58.789 ms |       0.13% | 17.886 ms |  43.73% |   FAIL   |
|  I64  |    UNIFORM     |     0.2     |  45.784 ms |      10.08% |  65.997 ms |       7.83% | 20.213 ms |  44.15% |   FAIL   |
|  I64  |    UNIFORM     |     0.3     |  47.134 ms |       0.01% |  69.038 ms |       0.10% | 21.904 ms |  46.47% |   FAIL   |
|  I64  |    UNIFORM     |     0.4     |  51.379 ms |       0.02% |  74.370 ms |       0.14% | 22.991 ms |  44.75% |   FAIL   |
|  I64  |    UNIFORM     |     0.5     |  57.030 ms |       0.02% |  80.411 ms |       0.15% | 23.382 ms |  41.00% |   FAIL   |
|  I64  |    UNIFORM     |     0.6     |  64.987 ms |       0.18% |  88.957 ms |       0.07% | 23.970 ms |  36.88% |   FAIL   |
|  I64  |    UNIFORM     |     0.7     |  76.416 ms |       0.03% | 101.227 ms |       0.09% | 24.812 ms |  32.47% |   FAIL   |
|  I64  |    UNIFORM     |     0.8     |  96.865 ms |       0.01% | 123.250 ms |       0.07% | 26.385 ms |  27.24% |   FAIL   |
|  I64  |    UNIFORM     |     0.9     | 152.041 ms |       0.01% | 181.059 ms |       0.07% | 29.018 ms |  19.09% |   FAIL   |

The performance decreased by about 20% to 40% depending on the case. This utility could be helpful to other cases though. To be tested.

PR #494, opened 2024-05-29
Add multiset benchmarks

### LINKED ISSUE #484: [ENHANCEMENT]: Add benchmarks for `static_multiset`

`static_multiset` is missing proper benchmarks as of now.

We should adhere to the existing `static_multimap` benchmarks, as well as some existing cudf join benchmark scenarios.

_No response_

### Additional context

_No response_

PointKernel:
Close #484

This PR adds benchmarks for multiset

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -67,7 +67,10 @@ ConfigureBench(STATIC_MAP_BENCH
 ###################################################################################################
 # - static_multiset benchmarks --------------------------------------------------------------------
 ConfigureBench(STATIC_MULTISET_BENCH
-  hash_table/static_multiset/count_bench.cu)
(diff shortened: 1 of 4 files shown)


sleeepyjack (approved):
Very nice!

I guess at some point we want to add axes for `cg_size` and `window_size` and so on but that should be a separate PR.

Ready to merge 🚀

PointKernel:
We had such benchmarks at one point and it took about one hour to run the whole benchmark suite so I removed it.

sleeepyjack:
Yeah, we probably don't want to run the full sweep by default since it takes ages. However, it would be nice to define those axes but only run the default settings by-default. This way we can add more values via the command line in case we want to explore the parameter space.

PointKernel:
Make sense.

PR #498, opened 2024-06-03
Multimap find and next_iterator api bugs

lin0303-siyuan:
Hi cuco team,

I have implemented two APIs for `static_multimap`, i.e `find` and `next_iterator`. And I have written an example program `examples/static_multimap/find.cu` to test that: Insert 50000 key-value pairs into the map, with `key = i % 25000` and `value = i`, which means that there will be two matches for each key between 0 and 24999. But what I get from the count is indeterministic if I am using `double hashing`, and get `25000` from `linear_probing` as probing schemes. (some elements get 2 matches and some get 0 matches).

Thanks for your time to review this issue. I think I may have some understanding issues in cuco multimap implementations and need your help.

If you want to run the codes,
```
ci/build.sh -e
./build/local/examples/STATIC_MULTIMAP_FIND_EXAMPLE
```


PointKernel:
The culprit is [here](https://github.com/lin0303-siyuan/cuCollections/blob/4c68353fb1aaa81eac43b98656f4085dc652ba14/include/cuco/detail/static_multimap/static_multimap.inl#L455-L466). You could not rely on host-bulk [insert](https://github.com/lin0303-siyuan/cuCollections/blob/4c68353fb1aaa81eac43b98656f4085dc652ba14/examples/static_multimap/find.cu#L176) in this case since it uses the internal `uses_vector_load`.

lin0303-siyuan:
But I have written some extra kernels that rely on the host insert and they are getting correct results (50000). I wonder how will the host-bulk insert affect the `find` I am doing in the kernel. Really appreciate if you could elaborate.

PointKernel:
> But I have written some extra kernels that rely on the host insert and they are getting correct results (50000).

The beauty of coincidence :smiling_face_with_tear: 

To make sure the hash table works properly, one **must** follow exactly the same scheme for both insertions and queries, i.e., the same CG size, the same vector load pattern, the same probing sequence (either double hashing or linear probing), any of the unmatches from the above list are undefined behavior thus could result in wrong output.

lin0303-siyuan:
Thanks, get resolved. The key takeway is that when using hashing map, vector_load should be consistent to avoid undefined behaviors.

PR #500, opened 2024-06-04
Add function for generating random byte strings with random lengths

sleeepyjack:
This PR adds a benchmark helper function for generating random byte strings with random lengths which will be used to test the impact of unaligned input data passed to our hashers.

### DIFF
--- a/include/cuco/utility/key_generator.cuh
+++ b/include/cuco/utility/key_generator.cuh
@@ -430,8 +430,8 @@ class key_generator {
  * @param n_sequences Number of byte sequences to generate
  * @param min_sequence_length Minimum sequence length
  * @param max_sequence_length Maximum sequence length
- * @param stream CUDA stream in which this operation is executed in
  * @param seed Random seed
+ * @param stream CUDA stream in which this operation is executed in
  *
  * @return A pair consisting of (1) a vector of spans pointing to each sequence and (2) a byte
  * vector holding the actual data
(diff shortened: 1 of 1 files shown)


include/cuco/utility/key_generator.cuh
```diff
@@ -414,4 +420,90 @@ class key_generator {
...
+ * @return A pair consisting of (1) a vector of spans pointing to each sequence and (2) a byte
+ * vector holding the actual data
+ */
+template <typename RNG = thrust::default_random_engine>
+std::pair<thrust::device_vector<cuda::std::span<std::byte>>, thrust::device_vector<std::byte>>
+generate_random_byte_sequences(std::size_t n_sequences,
+                               std::size_t min_sequence_length,
+                               std::size_t max_sequence_length,
+                               cudaStream_t stream = 0,
+                               std::size_t seed    = 0)
```
PointKernel:
```suggestion
                               std::size_t seed    = 0,
                               cudaStream_t stream = 0)
```
nit: Place stream as the last argument following cudf's styling. question: Do we actually want to expose the stream parameter?

sleeepyjack: It's not strictly neccessary since this function isn't async. However, I find it's best practice to always expose the stream parameter, in cases where the environment doesn't allow for using the default stream.

include/cuco/utility/key_generator.cuh
```diff
@@ -414,4 +420,90 @@ class key_generator {
...
+ *
+ * @tparam RNG Pseudo-random number generator
+ *
+ * @throws If the minimum/maximum sequence lengths are implausible
+ *
+ * @param n_sequences Number of byte sequences to generate
+ * @param min_sequence_length Minimum sequence length
+ * @param max_sequence_length Maximum sequence length
+ * @param stream CUDA stream in which this operation is executed in
+ * @param seed Random seed
```
sleeepyjack:
```suggestion
 * @param seed Random seed
 * @param stream CUDA stream in which this operation is executed in
```

PointKernel (approved):

PR #501, opened 2024-06-05
make murmurhash3_x64_128 compatible with existing cuco data structures

srinivasyadav18:
Make **murmurhash3_x64_128** compatible with existing **cuco data structures**

-  Moves `sanitize_hash` function from `cuco::detail::` **namespace** to `probing_scheme_base` **class** as **protected memeber**.
- Modifies the `santize_hash` function to handle `cuda::std::array<uint64_t, 2>`, which is returned from `murmurhash3_x64_128` hash function.
- Adds new `hash_test.cu` to test `static_map` API with all **hash functions.**

### DIFF
--- a/include/cuco/detail/probing_scheme_base.cuh
+++ b/include/cuco/detail/probing_scheme_base.cuh
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <cuco/detail/__config>
+
 #include <cstdint>
 
 namespace cuco {
@@ -30,6 +32,34 @@ namespace detail {
  */
 template <int32_t CGSize>
 class probing_scheme_base {
+ private:
+  template <typename SizeType, typename HashType>
+  __host__ __device__ constexpr SizeType sanitize_hash_positive(HashType hash) const noexcept
+  {
+    if constexpr (cuda::std::is_signed_v<SizeType>) {
+      return cuda::std::abs(static_cast<SizeType>(hash));
+    } else {
+      return static_cast<SizeType>(hash);
+    }
+  }
+
+ protected:
+  template <typename SizeType, typename HashType>
+  __host__ __device__ constexpr SizeType sanitize_hash(HashType hash) const noexcept
+  {
+    if constexpr (cuda::std::is_same_v<HashType, cuda::std::array<std::uint64_t, 2>>) {
+#if !defined(CUCO_HAS_INT128)
+      static_assert(false,
+                    "CUCO_HAS_INT128 undefined. Need unsigned __int128 type when sanitizing "
+                    "cuda::std::array<std::uint64_t, 2>");
+#endif
+      unsigned __int128 ret{};
+      memcpy(&ret, &hash, sizeof(unsigned __int128));
+      return sanitize_hash_positive<SizeType>(static_cast<SizeType>(ret));
+    } else
+      return sanitize_hash_positive<SizeType>(hash);
+  }
+
  public:
   /**
    * @brief The size of the CUDA cooperative thread group.
--- a/include/cuco/detail/probing_scheme_base.cuh
+++ b/include/cuco/detail/probing_scheme_base.cuh
@@ -32,34 +32,6 @@ namespace detail {
  */
 template <int32_t CGSize>
 class probing_scheme_base {
- private:
-  template <typename SizeType, typename HashType>
-  __host__ __device__ constexpr SizeType sanitize_hash_positive(HashType hash) const noexcept
-  {
-    if constexpr (cuda::std::is_signed_v<SizeType>) {
-      return cuda::std::abs(static_cast<SizeType>(hash));
-    } else {
-      return static_cast<SizeType>(hash);
-    }
-  }
-
- protected:
-  template <typename SizeType, typename HashType>
-  __host__ __device__ constexpr SizeType sanitize_hash(HashType hash) const noexcept
-  {
-    if constexpr (cuda::std::is_same_v<HashType, cuda::std::array<std::uint64_t, 2>>) {
-#if !defined(CUCO_HAS_INT128)
-      static_assert(false,
-                    "CUCO_HAS_INT128 undefined. Need unsigned __int128 type when sanitizing "
-                    "cuda::std::array<std::uint64_t, 2>");
-#endif
-      unsigned __int128 ret{};
-      memcpy(&ret, &hash, sizeof(unsigned __int128));
-      return sanitize_hash_positive<SizeType>(static_cast<SizeType>(ret));
-    } else
-      return sanitize_hash_positive<SizeType>(hash);
-  }
-
  public:
   /**
    * @brief The size of the CUDA cooperative thread group.
--- a/include/cuco/detail/probing_scheme_base.cuh
+++ b/include/cuco/detail/probing_scheme_base.cuh
@@ -16,8 +16,6 @@
 
 #pragma once
 
-#include <cuco/detail/__config>
-
 #include <cstdint>
 
 namespace cuco {
--- a/include/cuco/detail/probing_scheme_impl.inl
+++ b/include/cuco/detail/probing_scheme_impl.inl
@@ -120,9 +120,8 @@ __host__ __device__ constexpr auto linear_probing<CGSize, Hash>::operator()(
   Extent upper_bound) const noexcept
 {
   using size_type = typename Extent::value_type;
-  using cg_type   = cooperative_groups::thread_block_tile<cg_size>;
   return detail::probing_iterator<Extent>{
-    cuco::detail::sanitize_hash<cg_type, size_type>(g, hash_(probe_key)) % upper_bound,
+    cuco::detail::sanitize_hash<size_type>(g, hash_(probe_key)) % upper_bound,
     cg_size,
     upper_bound};
 }
@@ -164,9 +163,8 @@ __host__ __device__ constexpr auto double_hashing<CGSize, Hash1, Hash2>::operato
   Extent upper_bound) const noexcept
 {
   using size_type = typename Extent::value_type;
-  using cg_type   = cooperative_groups::thread_block_tile<cg_size>;
   return detail::probing_iterator<Extent>{
-    cuco::detail::sanitize_hash<cg_type, size_type>(g, hash1_(probe_key)) % upper_bound,
+    cuco::detail::sanitize_hash<size_type>(g, hash1_(probe_key)) % upper_bound,
     static_cast<size_type>(
       (cuco::detail::sanitize_hash<size_type>(hash2_(probe_key)) % (upper_bound / cg_size - 1) +
        1) *
--- a/include/cuco/detail/probing_scheme_impl.inl
+++ b/include/cuco/detail/probing_scheme_impl.inl
@@ -121,7 +121,7 @@ __host__ __device__ constexpr auto linear_probing<CGSize, Hash>::operator()(
 {
   using size_type = typename Extent::value_type;
   return detail::probing_iterator<Extent>{
-    cuco::detail::sanitize_hash<size_type>(hash_(probe_key), g.thread_rank()) % upper_bound,
+    cuco::detail::sanitize_hash<size_type>(hash_(probe_key), g) % upper_bound,
     cg_size,
     upper_bound};
 }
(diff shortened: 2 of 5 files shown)


include/cuco/detail/probing_scheme_base.cuh
```diff
@@ -30,6 +32,34 @@ namespace detail {
  */
 template <int32_t CGSize>
 class probing_scheme_base {
+ private:
+  template <typename SizeType, typename HashType>
+  __host__ __device__ constexpr SizeType sanitize_hash_positive(HashType hash) const noexcept
```
PointKernel:
```suggestion
  __device__ constexpr SizeType sanitize_hash_positive(HashType hash) const noexcept
```
nit: all probing happen on device thus `__host__` qualifier is not needed

include/cuco/detail/probing_scheme_base.cuh
```diff
@@ -30,6 +32,34 @@ namespace detail {
...
+    if constexpr (cuda::std::is_signed_v<SizeType>) {
+      return cuda::std::abs(static_cast<SizeType>(hash));
+    } else {
+      return static_cast<SizeType>(hash);
+    }
+  }
+
+ protected:
+  template <typename SizeType, typename HashType>
+  __host__ __device__ constexpr SizeType sanitize_hash(HashType hash) const noexcept
```
PointKernel:
```suggestion
  __device__ constexpr SizeType sanitize_hash(HashType hash) const noexcept
```

include/cuco/detail/probing_scheme_impl.inl
```diff
@@ -107,7 +107,7 @@ __host__ __device__ constexpr auto linear_probing<CGSize, Hash>::operator()(
 {
   using size_type = typename Extent::value_type;
   return detail::probing_iterator<Extent>{
-    cuco::detail::sanitize_hash<size_type>(hash_(probe_key)) % upper_bound,
+    probing_scheme_base_type::template sanitize_hash<size_type>(hash_(probe_key)) % upper_bound,
```
PointKernel: This could be simplified a bit, e.g.
```cpp
struct linear_probing {
  ...
  using probing_scheme_base_type::sanitize_hash;
  ...
};
```
Then here 
```suggestion
    sanitize_hash<size_type>(hash_(probe_key)) % upper_bound,
```
should be working just fine.

include/cuco/detail/utils.cuh
```diff
@@ -17,6 +17,7 @@
 
 #include <cuco/detail/bitwise_compare.cuh>
 
+#include <cuda/std/array>
```
PointKernel, line 20:
```suggestion
```
suggestion: A leftover to be removed. could you also remove the inclusion of `#include <cuda/std/cmath>` as well since we no longer invoke ` cuda::std::abs` in this file?

PointKernel (changes requested):
Some minor cleanups.

Based on the offline discussions, we are not happy with the fact that `sanitize_hash` has to be invoked twice in CG-based probing. @sleeepyjack Any idea how to improve the situation?

include/cuco/detail/probing_scheme_impl.inl
srinivasyadav18: sanitize_hash being a template, we still need it :).

sleeepyjack:
I have some high-level questions/suggestions regarding this PR:

- Why do we need to move the `sanitize_hash` function to the `probing_scheme_base`?
- In my opinion, there is no real world use in having a size type larger than `sizeof(size_t)`. So why not simply truncate the hash value to that max bit width and then call `sanitize_hash` on it?
- The `sanitize_hash` function still has some unresolved problems (see draft PR #362) which I haven't been able to fix yet. Any chance we can also get this resolved in this PR? Should be a small but fun challenge @srinivasyadav18 ;)
- nit: GH regularly gets confused and loses the discussion history when rebasing+force-pushing to the PR branch. We're fine with just adding or merging on top of the current HEAD...we will squash-merge the branch anyway so the intermediate commits will not be visible in the default branch.

PointKernel:
> Why do we need to move the sanitize_hash function to the probing_scheme_base?

I suggested to do so since probing is the only place using `sanitize_hash`.

> So why not simply truncate the hash value to that max bit width and then call sanitize_hash on it?

It works but technically we are not using the hash value in a proper way, is it fair to say so?

sleeepyjack:
> I suggested to do so since probing is the only place using sanitize_hash.

And I agree that it makes sense to put utils with the class where they have one-time use. However, since it's a template function the syntax doesn't get much cleaner doing so: `probing_scheme_base::template sanitize_hash<SizeType>(...)` vs. `cuco::detail::sanitize_hash<SizeType>(...)`.

> It works but technically we are not using the hash value in a proper way, is it fair to say so?

`static_cast<size_t>(some_uint128)` will also just truncate the upper bits so I'd say it's valid. We may want to add a `static_assert` to our `extent` type to limit the bit width to 64.

include/cuco/detail/utils.cuh
```diff
@@ -92,12 +103,33 @@ struct slot_is_filled {
...
+/**
+ * @brief Converts a given hash value and cg_rank, into a valid (positive) size type.
+ *
+ * @tparam SizeType The target type
+ * @tparam HashType The input type
+ *
+ * @return Converted hash value
+ */
+template <typename SizeType, typename HashType>
+__host__ __device__ constexpr SizeType sanitize_hash(HashType hash, std::uint32_t cg_rank) noexcept
```
sleeepyjack:
```suggestion
/**
 * @brief Converts a given hash value into a valid (positive) size type during cooperative probing.
 *
 * @tparam SizeType The target type
 * @tparam HashType The input type
 * @tparam CG Cooperative group type
 *
 * @return Converted hash value
 */
template <typename SizeType, typename HashType, typename CG>
__host__ __device__ constexpr SizeType sanitize_hash(HashType hash, CG group) noexcept
```

I would suggest passing in the CG itself rather that the rank. This is a common pattern for CG operations.

include/cuco/detail/utils.cuh
```diff
@@ -92,12 +103,33 @@ struct slot_is_filled {
...
+{
+  return sanitize_hash<SizeType>(sanitize_hash<SizeType>(hash) + cg_rank);
```
sleeepyjack: There exists a scenario where this approach fails and it's the reason CI in my initial draft PR failed.
Consider the following example:

- `SizeType` is `int32_t` aka a signed type
- The value of `sanitize_hash<SizeType>(hash)` is very close to `numeric_limits<SizeType>::max()`

In this scenario, if we compute `sanitize_hash<SizeType>(hash) + cg_rank` there's chance the result oxceeds `numeric_limits<SizeType>::max()` which would result in a signed integer overflow which is undefined behavior under the C++ abstract machine. Thus the compiler is free to produce any garbage code around this call.

To solve this we need check if `sanitize_hash<SizeType>(hash) > (numeric_limits<SizeType>::max() - group.size())` (be careful with `>` and `>=`, I'm infamous for my off-by-one errors) and then compute the wrapped-around value manually in case this expression evaluates to `true`

sleeepyjack (changes requested):

include/cuco/detail/utils.cuh
srinivasyadav18: Thanks for the detailed explanation! 
I think [3ae1afe](https://github.com/NVIDIA/cuCollections/pull/501/commits/3ae1afeb0e2b55d1b2ede3496c5de82a74541ce0) covers this case now.

include/cuco/detail/utils.cuh
```diff
@@ -92,12 +106,39 @@ struct slot_is_filled {
...
+ * @brief Converts a given hash value and cg_rank, into a valid (positive) size type.
+ *
+ * @tparam SizeType The target type
+ * @tparam HashType The input type
+ * @tparam CG Cooperative group type
+ *
+ * @return Converted hash value
+ */
+template <typename SizeType, typename HashType, typename CG>
+__host__ __device__ constexpr SizeType sanitize_hash(HashType hash, CG group) noexcept
```
PointKernel:
```suggestion
template < typename CG, typename SizeType, typename HashType>
__device__ constexpr SizeType sanitize_hash(CG const& group, HashType hash) noexcept
```
styling nits:
- CG is device-only
- CG APIs always take CG as the first argument
- Pass by const ref

include/cuco/detail/probing_scheme_base.cuh
```diff
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <cuco/detail/__config>
+
```
PointKernel:
```suggestion
```
Is this a leftover?

PointKernel (changes requested):
some styling nits otheriwse good to go

PointKernel (approved):
Great work as always! Thanks!

include/cuco/detail/utils.cuh
```diff
@@ -122,13 +125,19 @@ __host__ __device__ constexpr SizeType sanitize_hash(HashType hash) noexcept
...
-__host__ __device__ constexpr SizeType sanitize_hash(HashType hash, std::uint32_t cg_rank) noexcept
+template <typename SizeType, typename HashType, typename CG>
+__host__ __device__ constexpr SizeType sanitize_hash(HashType hash, CG group) noexcept
 {
-  return sanitize_hash<SizeType>(sanitize_hash<SizeType>(hash) + cg_rank);
+  auto const base_hash = sanitize_hash<SizeType>(hash);
+  auto const max_size  = cuda::std::numeric_limits<SizeType>::max();
+  auto const cg_rank   = static_cast<SizeType>(group.thread_rank());
+
+  if (base_hash > (max_size - cg_rank)) return cg_rank - (max_size - base_hash);
```
sleeepyjack: I might be confused but `cg_rank - (max_size - base_hash)` can easily become a negative number which is not what we are looking for as the return value of this function. Am I missing something?

include/cuco/detail/probing_scheme_impl.inl
```diff
@@ -120,8 +120,9 @@ __host__ __device__ constexpr auto linear_probing<CGSize, Hash>::operator()(
   Extent upper_bound) const noexcept
 {
   using size_type = typename Extent::value_type;
+  using cg_type   = cooperative_groups::thread_block_tile<cg_size>;
   return detail::probing_iterator<Extent>{
-    cuco::detail::sanitize_hash<size_type>(hash_(probe_key) + g.thread_rank()) % upper_bound,
+    cuco::detail::sanitize_hash<cg_type, size_type>(g, hash_(probe_key)) % upper_bound,
```
sleeepyjack: I would move the `size_type` to the front of the tparam list so you don't have to specify the `cg_type` as it can be inferred from `g`.

include/cuco/detail/utils.cuh
```diff
@@ -92,12 +106,39 @@ struct slot_is_filled {
...
+ * @return Converted hash value
+ */
+template <typename CG, typename SizeType, typename HashType>
+__device__ constexpr SizeType sanitize_hash(CG const& group, HashType hash) noexcept
+{
+  auto const base_hash = sanitize_hash<SizeType>(hash);
+  auto const max_size  = cuda::std::numeric_limits<SizeType>::max();
+  auto const cg_rank   = static_cast<SizeType>(group.thread_rank());
+
+  if (base_hash > (max_size - cg_rank)) return cg_rank - (max_size - base_hash);
```
PointKernel:
```suggestion
  if (base_hash > (max_size - cg_rank)) { return cg_rank - (max_size - base_hash); }
```
always use curly braces

include/cuco/detail/probing_scheme_impl.inl
PointKernel: It's about the intention of the API and the syntax consistency.

sleeepyjack: Sure, the `g` parameter should be the first one for consistency reasons but we can still use a different ordering for the tparam list, i.e., the one that lets us make use of automatic type inference. The only tparam that cannot be inferred is the result size type so specifying the CG type is redundant.

include/cuco/detail/utils.cuh
sleeepyjack: Resolved during offline discussion.

include/cuco/detail/probing_scheme_impl.inl
sleeepyjack: It's an internal API so I don't want to bikeshed too much. I'm okay with merging it as is.

sleeepyjack (approved):
🚀 🚀 🚀

PR #503, opened 2024-06-12
Fix a build warning with ARM

PointKernel:
Fixes a build warning with ARM
```bash
/home/coder/cuCollections/include/cuco/detail/static_map_kernels.cuh(446): warning #20054-D: dynamic initialization is not supported for a function-scope static __shared__ variable within a __device__/__global__ function
395
              Value writeBuffer[block_size / tile_size];


sleeepyjack:
Why on earth does the compiler think this is dynamic initialization?

sleeepyjack (approved):
Can we add a small note that this is due to ARM compiler complaining about dynamic init aka false-positive? Just for future reference.

Apart from that LGTM

PR #505, opened 2024-06-14
Issue 487 [WIP]

njones93531:
[WIP]
closes 487
- Adds a new class `perfect_probing<class Hash>`
- Adds template checking to `open_addressing_ref_impl.cuh` to tell if a probing scheme is perfect
- For perfect probing in `open_addressing_ref_impl.cuh`:
  - for all `insert` functions, do not check the existing value. Instead, simply overwrite with the new value.
  - Replaces CAS operations with non-atomic operations
  - Instead of using `KeyEqual`, `contains` and `find` are completed by checking if the slot value is the sentinel or not. 
- Adds a new hasher identityhash.cuh
  - Example hash function to test perfect probing 
  - Is perfect if the hash table has size >= number of possible keys
- Adds a new hasher sigbithash.cuh (WIP)
  - A more realistic perfect hash function, the basis of other proposed perfect hashes
  - Reads the entire input on construction to determine a significant bits mask
  - Hashes a key into key ^ sigbit_mask
  - Is perfect as long as the introspection at construction is complete

Supporting Documentation: 
1. The identity hasher behaves as expected. 
Included in the code is `cuco/examples/other/identity_test.cu`, an example that shows the identity hasher behaves as expected. Below is the output of this example:
image

2.  Perfect probing successfully avoids atomics. 
Below is a `diff` of the `.ptx` code generated by `cuco/examples/static_set/host_bulk_insert.cu` when using the default set construction and when using a set with perfect probing. Notice that the atomic operation `atom.global.cas.b32` disappears. Perfect probing is in blue. 
image
Additionally, some of the non-atomic calculations become much simpler, as seen below. 
image

3. Perfect probing improves performance. 
Included in the code is a version of the `static_set` benchmarks modified to support perfect probing. After running the benchmark with `xxhash` and `double_hashing` compared to `identityhash` and `perfect_probing`, I found that perfect probing matched or exceeded the performance of `xxhash` in all of the tests. 
In the following tests, perfect probing performed significantly better, with an average speedup of 1.17x and a maximum speedup of 
1.36x:
- static_set_constains_unique_capacity
- static_set_find_unique_capacity
The following tests saw a speedup of 1.015x - 1.06x:
- static_set_insert_unique_occupancy
- static_set_find_unique_matching_rate
- static_set_find_unique_occupancy
The remaining 7 tests saw no change due to perfect probing.


njones93531:
After using `c++filt`, I can say the first `.ptx` screenshot where atomics are avoided is from `insert_if_n()`. The second, where a large amount of code is replaced/condensed, is from `contains_if_n()`.

benchmarks/benchmark_defaults.hpp
```diff
@@ -23,10 +23,12 @@
...
 
-using KEY_TYPE_RANGE   = nvbench::type_list<nvbench::int32_t, nvbench::int64_t>;
-using VALUE_TYPE_RANGE = nvbench::type_list<nvbench::int32_t, nvbench::int64_t>;
+using KEY_TYPE_RANGE   = nvbench::type_list<nvbench::int32_t>;
+using VALUE_TYPE_RANGE = nvbench::type_list<nvbench::int32_t>;
+// using KEY_TYPE_RANGE   = nvbench::type_list<nvbench::int32_t, nvbench::int64_t>;
+// using VALUE_TYPE_RANGE = nvbench::type_list<nvbench::int32_t, nvbench::int64_t>;
 
-auto constexpr N             = 100'000'000;
+auto constexpr N             = 10'000'000;
```
PointKernel: probably a test leftover, can you please revert unrelated changes throughout the PR?

benchmarks/hash_table/static_set/insert_bench.cu
```diff
@@ -47,8 +47,13 @@ void static_set_insert(nvbench::state& state, nvbench::type_list<Key, Dist>)
...
              [&](nvbench::launch& launch, auto& timer) {
-               cuco::static_set<Key> set{
-                 size, cuco::empty_key<Key>{-1}, {}, {}, {}, {}, {}, {launch.get_stream()}};
+               cuco::static_set<Key,
+                                cuco::extent<std::size_t>,
+                                cuda::thread_scope_device,
+                                thrust::equal_to<Key>,
+                                cuco::double_hashing<4,  // CG size
+                                                     cuco::default_hash_function<Key>>>
+                 set{size, cuco::empty_key<Key>{-1}, {}, {}, {}, {}, {}, {launch.get_stream()}};
```
PointKernel: question: is this change necessary, please revert if not.

benchmarks/hash_table/static_set_perfect/contains_bench.cu
PointKernel, line 1: instead of creating a new file, we could just make the original `contains` benchmark code taking a template probing scheme parameter.

benchmarks/hash_table/static_set_perfect/insert_bench.cu
```diff
@@ -0,0 +1,87 @@
...
+             [&](nvbench::launch& launch, auto& timer) {
+               cuco::static_set<Key,
+                                cuco::extent<std::size_t>,
+                                cuda::thread_scope_device,
+                                thrust::equal_to<Key>,
+                                perfect_static_set::defaults::PROBER<Key>>
+                 set{size, cuco::empty_key<Key>{-1}, {}, {}, {}, {}, {}, {launch.get_stream()}};
+
+               timer.start();
+               set.insert(keys.begin(), keys.end(), {launch.get_stream()});
```
PointKernel, line 60:
```suggestion
               set.insert_async(keys.begin(), keys.end(), {launch.get_stream()});
```
`insert` includes atomic counting operations to track the number of successful insertions, let's benchmark `insert_async` instead.

benchmarks/hash_table/static_set_perfect/insert_bench.cu
```diff
@@ -0,0 +1,87 @@
...
+               timer.stop();
+             });
+}
+
+NVBENCH_BENCH_TYPES(static_set_insert,
+                    NVBENCH_TYPE_AXES(defaults::KEY_TYPE_RANGE,
+                                      nvbench::type_list<distribution::uniform>))
+  .set_name("static_set_insert_uniform_multiplicity")
```
PointKernel, line 68: this could be a naming conflict with the original insert bench.

examples/other/identity_test.cu
```diff
@@ -0,0 +1,94 @@
...
+
+#define cudaCheckError(call)                                                              \
+  {                                                                                       \
+    cudaError_t err = call;                                                               \
+    if (err != cudaSuccess) {                                                             \
+      std::cerr << "CUDA error in file '" << __FILE__ << "' in line " << __LINE__ << ": " \
+                << cudaGetErrorString(err) << "." << std::endl;                           \
+      exit(EXIT_FAILURE);                                                                 \
+    }                                                                                     \
+  }
```
PointKernel:
```suggestion
```
please use `CUCO_CUDA_TRY` instead

examples/static_set/host_bulk_example.cu
```diff
@@ -48,7 +48,13 @@ int main(void)
...
 
   // Constructs a set with at least `capacity` slots using -1 as the empty keys sentinel.
-  cuco::static_set<Key> set{capacity, cuco::empty_key{empty_key_sentinel}};
+  cuco::static_set<Key,
+                   cuco::extent<std::size_t>,
+                   cuda::thread_scope_device,
+                   thrust::equal_to<Key>,
+                   cuco::double_hashing<4,  // CG size
+                                        cuco::default_hash_function<Key>>>
+    set{capacity, cuco::empty_key{empty_key_sentinel}};
```
PointKernel: to be reverted

examples/static_set/host_bulk_perfect_example.cu
```diff
@@ -0,0 +1,78 @@
...
+
+#include <iostream>
+#include <limits>
+
+/**
+ * @file host_bulk_example.cu
+ * @brief Demonstrates usage of the static_set "bulk" host APIs.
+ *
+ * The bulk APIs are only invocable from the host and are used for doing operations like `insert` or
+ * `contains` on a set of keys.
```
PointKernel, line 32: Can you please update the doc to describe the purpose of this example better?

examples/static_set/host_bulk_perfect_example.cu
```diff
@@ -0,0 +1,78 @@
...
+  std::size_t const capacity = std::ceil(num_keys / load_factor);
+
+  // Constructs a set with at least `capacity` slots using -1 as the empty keys sentinel.
+  cuco::static_set<Key,
+                   cuco::extent<std::size_t>,
+                   cuda::thread_scope_device,
+                   thrust::equal_to<Key>,
+                   cuco::perfect_probing<4,  // CG size
+                                         cuco::default_hash_function<Key>>>
+    set{capacity, cuco::empty_key{empty_key_sentinel}};
```
PointKernel, line 57: Based on our offline discussion, we probably don't want to expose `perfect_probing` publicly. Instead, it should be only available via a `create_perfect_probing` helper.

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,144 @@
...
+#pragma once
+
+#include <cuco/detail/hash_functions/utils.cuh>
+#include <cuco/extent.cuh>
+
+#include <cstddef>
+#include <cstdint>
+
+namespace cuco::detail {
+
```
PointKernel, line 26: I feel the whole `IdentityHash_*` logic can be just 
```cpp
template<typename Key>
using identity_hash = thrust::identity<Key>;
```
Make sense to you? Probably no need to have `_32`/`_64` variants since cuco internals will convert them to `size_type` anyway.

include/cuco/detail/hash_functions/sigbithash.cuh
```diff
@@ -0,0 +1,153 @@
...
+  using result_type   = std::uint32_t;  ///< The type of the hash values produced
+
+  /**
+   * @brief Constructs a SigBitHash_32 hash function by finding the significant bits of
+   * a given span of keys
+   *
+   * @param start 	A pointer/iterator to the first key in the list
+   * @param numKeys 	The number of keys in the span
+   */
+  __host__ __device__ constexpr SigBitHash_32(Key* start, std::uint32_t numKeys)
```
PointKernel, line 46: question: is there a seed parameter for sighash?

include/cuco/detail/hash_functions/utils.cuh
```diff
@@ -16,6 +16,13 @@
 
 #pragma once
 
+#include <cub/cub.cuh>
+
+#include <cuda_runtime.h>
+
```
PointKernel:
```suggestion
```
nit: this is only needed when building with host compilers like gcc, `.cuh` files are built with `nvcc` by default thus this is redundant

include/cuco/probing_scheme.cuh
```diff
@@ -175,6 +177,159 @@ class double_hashing : private detail::probing_scheme_base<CGSize> {
...
+     * @brief Prefix increment operator
+     *
+     * @return Current iterator
+     */
+    __host__ __device__ constexpr auto operator++() noexcept
+    {
+      // TODO: This is still getting called, and should never really be used
+      curr_index_ = (curr_index_ + 1) % upper_bound_;
+      // curr_index_ = 0;
+      return *this;
```
PointKernel, line 310:
```suggestion
  static_assert(false, "perfect probing doesn't increment");
```
question: do we need to probe (increment the pointer) in the case of perfect hashing?

PointKernel (changes requested):
@njones93531 Thanks for the work!

Just a quick pass to get some cleanups first:

- Shall we remove all benchmarks but keep `insert` and `contains` only?
- We probably see more speedups by benchmarking `insert_async` instead of `insert`
- A good first example will be demonstrating the use of `create_perfect_probing`. `perfect_probing` ctor should be private
- The identity hash seems to be `thrust::identity`
- Please revert unrelated changes

include/cuco/detail/hash_functions/sigbithash.cuh
esoha-nvidia: No.  And it doesn't need one, actually!

The point of a seed in a traditional hash is so that a hash will not behave the same each time.  This is to prevent denial-of-service attacks based on hash collisions, like this: https://stackoverflow.com/questions/52184366/why-does-hashmap-need-a-cryptographically-secure-hashing-function

However, a perfect hash has no collisions.  (Collision is defined as `hash(x) == hash(y) && x != y`.)  So with no collisions, there is no need for the seed.

include/cuco/detail/hash_functions/sigbithash.cuh
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * @brief An SigBit hash function to hash the given argument on host and device.
+ *
+ * -----------------------------------------------------------------------------
+ *
+ * @tparam Key The type of the values to hash
+ */
+template <typename Key>
+struct SigBitHash_32 {
```
esoha-nvidia, line 35: I think that this can be in a separate PR from the perfect probing PR.

include/cuco/detail/hash_functions/sigbithash.cuh
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * @brief Returns a hash value for its argument, as a value of type `result_type`.
+   *
+   * @param key The input argument to hash
+   * @return The resulting hash value for `key`
+   */
+  constexpr result_type __host__ __device__ operator()(Key const& key) const noexcept
+  {
+    if constexpr (sizeof(Key) <= 16) {
+      Key const key_copy = key;
```
esoha-nvidia, line 60: I worry that this will introduce extra copies unnecessarily if the `Key` is not contiguous in memory.  It would be nice to be able to hash non-contiguous memory without copying.  And for that, I think that we need a recursive template.  Let's discuss this in Slack with @sleeepyjack and @PointKernel

include/cuco/detail/hash_functions/sigbithash.cuh
```diff
@@ -0,0 +1,153 @@
...
+   *
+   * @param bytes The input argument to hash
+   * @param size The extent of the data in bytes
+   * @return The resulting hash value
+   */
+  template <typename Extent>
+  constexpr result_type __host__ __device__ compute_hash(std::byte const* bytes,
+                                                         Extent size) const noexcept
+  {
+    std::uint64_t h1 = load_chunk<std::uint64_t>(bytes, 0) & sigbit_mask_;
```
esoha-nvidia, line 145: Of course, this is not the real significant bits hash because it's missing all the shifting and stuff.  I think that we can move this file entirely into a different PR and then, in that one, implement the real deal.

include/cuco/detail/hash_functions/sigbithash.cuh
esoha-nvidia, line 46: This argument will actually need to be `Key**` because we need to compute the significant bits of everything that might be inserted or search for.  For example, imagine this code:

```c++
vector<int> insert_list, lookup_list;
hash.insert(insert_list.data(), insert_list.size());
hash.lookup(lookup_list.data(), lookup_list.size());
```
You need the significant bits hasher to take into account *everything* that might ever be hashed, not just the stuff that is inserted.  Your significant bits hasher needs to compute the significant bits of the `insert_list` and the `lookup_list`.  Those might not be contiguous, so best to allow the user to provide multiple arrays.

I know this because I did it wrong and ended up with errors that were very hard to reason about!  :-(

include/cuco/detail/hash_functions/utils.cuh
```diff
@@ -27,4 +34,100 @@ constexpr __host__ __device__ T load_chunk(U const* const data, Extent index) no
...
+  AndOrPairCombiner combiner_op;
+  auto init = AndOrPair<T>(T(-1), T(0));
+
+  // Determine temporary device storage requirements
+  void* d_temp_storage      = NULL;
+  size_t temp_storage_bytes = 0;
+  cub::DeviceReduce::Reduce(
+    d_temp_storage, temp_storage_bytes, inputs, output, inputs_size, combiner_op, init);
+  // Allocate temporary storage
+  cudaMalloc(&d_temp_storage, temp_storage_bytes);
```
esoha-nvidia, line 78: We definitely must allow the user to provide an allocator.  Much of cuco already has an allocator as a template param.  You can probably find examples in cuco that are already doing this and you can copy those ideas.

include/cuco/detail/hash_functions/utils.cuh
```diff
@@ -27,4 +34,100 @@ constexpr __host__ __device__ T load_chunk(U const* const data, Extent index) no
...
+  // Run reduction
+  cub::DeviceReduce::Reduce(
+    d_temp_storage, temp_storage_bytes, inputs, output, inputs_size, combiner_op, init);
+  cudaFree(d_temp_storage);
```
esoha-nvidia, line 82: Perhaps nicer to use a smart pointer for the recently allocated memory so that you don't accidentally leak anything if, for example, `Reduce` throws an exception.

If the rest of cuco is already not doing this then I suppose you don't have to either.  Though, if I saw this, I would open an issue and ask cuco to user smart pointers everywhere!

include/cuco/detail/hash_functions/utils.cuh
```diff
@@ -27,4 +34,100 @@ constexpr __host__ __device__ T load_chunk(U const* const data, Extent index) no
...
+void copy_to_host(T* host_data, const T* device_data, int64_t size)
+{
+  cudaMemcpy(host_data, device_data, size * sizeof(T), cudaMemcpyDeviceToHost);
+}
+
+// The first element of and_or gets the binary and reduction of all
+// the inputs. The second element of and_or gets the binary or
+// reduction of all the inputs.
+template <typename T>
+AndOrPair<T> reduce_and_or_cuda(const T* host_input, int64_t num_elements)
```
esoha-nvidia, line 115: The data will often already be on the GPU.  Perhaps just write the version that assumes that all the data is already on the GPU and then leave it up to the user to call a memcpy as needed.

Also, we'll often want the reduction over multiple arrays.  You can do it in two phases: First one reduction per input to generate a single `AndOrPair` per input.  And then one more reduction of those per-input results to get a complete result.  And have a special case for when there is only one array.

For now, you can have the result go into GPU memory, as you are doing, and the user can memcpy it to CPU memory to read it there, which is what the user will eventually need in order to compute the sigbits.  In a future PR, you can look into making this function even more flexible to allow the user to select pinned memory as an output in order to avoid memcpy.

This is pretty technical and it took me months to understand it in the first place.  Go ahead and ask questions.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -504,35 +522,47 @@ class open_addressing_ref_impl {
...
-              if constexpr (has_payload) {
-                // wait to ensure that the write to the value part also took place
-                this->wait_for_payload((window_ptr + i)->second, this->empty_value_sentinel());
-              }
-              return {iterator{&window_ptr[i]}, true};
+          return {iterator{&window_ptr[i]}, true};
+        } else {
+          // Default behavior: eq check
+          auto const eq_res =
+            this->predicate_.operator()<is_insert::YES>(key, this->extract_key(window_slots[i]));
```
esoha-nvidia, line 537: Again, this will just be asking, "does the slot contain anything at all?".  Because if the slot is occupied, then it's equal.

Given that this change is the same as above, I feel like the correct place to make the change is actually in the predicate itself!  The equality predicate itself can be changed and then the correct behavior simply falls through in the rest of the code?  Maybe?  This will make the diffs smaller, too!

We sort of want the change to be here:

https://github.com/NVIDIA/cuCollections/blob/057d3e05cf16d7df59e346b05246754bf0729802/include/cuco/detail/equal_wrapper.cuh#L72-L76

If perfect hashing then equal_to is always true.  However, this only works if `equal_to` is only called when neither `lhs` nor `rhs` are empty/erased.  One way to ensure that would be to make it a private function so that it is only called by `operator()` at https://github.com/NVIDIA/cuCollections/blob/057d3e05cf16d7df59e346b05246754bf0729802/include/cuco/detail/equal_wrapper.cuh#L95-L108 .  If it were me, I would try to make that `equal_to` function a private one and see if everything still compiles.  If so, then you can just change the `equal_to` function.  That would be best.  If you are unable to make it private because it is called elsewhere, then there is more work to do but it can be done.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -757,10 +787,16 @@ class open_addressing_ref_impl {
       auto const window_slots = storage_ref_[*probing_iter];
 
       for (auto& slot_content : window_slots) {
-        switch (this->predicate_.operator()<is_insert::NO>(key, this->extract_key(slot_content))) {
-          case detail::equal_result::UNEQUAL: continue;
-          case detail::equal_result::EMPTY: return false;
-          case detail::equal_result::EQUAL: return true;
+        if constexpr (is_perfect_hashing<decltype(probing_scheme_)>::value) {
+          // Perfect hashing only needs to compare to sentinel
+          return !(this->extract_key(slot_content) ^ empty_slot_sentinel_);
```
esoha-nvidia, line 792: Better to just use a regular equality operator and let the compiler optimize this stuff for you.  Compilers are really good at these mini optimizations.  Also, it might be that what you've written is currently the fastest way to an equality check but maybe in future hardware `!(a-b)` will be faster and you will have unnecessarily written slower code.  A compiler, however, must keep up with all possible hardware because it needs to compile for multiple targets.  And it will know the best way to do an equality comparison on various platforms.

All this is moot if we just fix the predicate instead.  Also, the predicate uses `cuco::detail::bitwise_compare`.  I don't know if that's better than xor or just the same thing but lets trust that they did it right and just use that!

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -1208,14 +1253,21 @@ class open_addressing_ref_impl {
...
                                                         Value const& desired) noexcept
   {
-    if constexpr (sizeof(value_type) <= 8) {
-      return packed_cas(address, expected, desired);
+    if constexpr (is_perfect_hashing<decltype(probing_scheme_)>::value) {
+      // Do it without atomics.
+      *address = desired;
+      return insert_result::SUCCESS;  // Perfect probing guarantees success
     } else {
+      // Default: use atomics
```
esoha-nvidia, line 1261: This code appears a few times below.  We should aim to remove the code duplication for DRY principles. https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

I think that we should make a new function that gets `is_perfect_hashing` as a template parameter and only invokes `packed_cas` as necessary.  And then use that function instead of `packed_cas`, here.  You could call it `maybe_packed_cas<is_perfect>(address, expected, desired)`.  

If it were me doing this, I would be afraid of missing a call to `packed_cas` on accident so I would probably intentionally rename `packed_cas` to `packed_cas_foo` so that the compiler will throw an error for every call to `packed_cas`.  And then I can examine each one to make sure that it truly is a candidate for conversion to `maybe_packed_cas`.

include/cuco/detail/probing_scheme_impl.inl
```diff
@@ -163,12 +163,50 @@ __host__ __device__ constexpr auto double_hashing<CGSize, Hash1, Hash2>::operato
...
+}
+
+template <int32_t CGSize, typename Hash>
+template <typename ProbeKey, typename Extent>
+__host__ __device__ constexpr auto perfect_probing<CGSize, Hash>::operator()(
+  ProbeKey const& probe_key, Extent upper_bound) const noexcept
+{
+  using size_type = typename Extent::value_type;
+  return cuco::perfect_probing<CGSize, Hash>::probing_iterator<Extent>{
+    cuco::detail::sanitize_hash<size_type>(hash_(probe_key)) % upper_bound, upper_bound};
```
esoha-nvidia, line 196: Does perfect probing need a modulo?

You can only use perfect probing if the hash output is guaranteed to be smaller than the size of the hash table, otherwise, it will not work.  Perfect probing is basically two conditions:

1. The hash is perfect: `hash(x) == hash(y) => x == y`.  Identity and sigbits hash qualify.  xxhash does not.
2. The hash table slot count is greater than the maximum possible output of the hash function: `hash(x) < slot_count for all x`.

Given number 2, the modulo is going to have no affect, right?  Modulo by non-constant in code is actually quite expensive because it is a division operation so removing this could be a big deal!  Consider measuring it.

include/cuco/detail/probing_scheme_impl.inl
```diff
@@ -163,12 +163,50 @@ __host__ __device__ constexpr auto double_hashing<CGSize, Hash1, Hash2>::operato
...
+template <int32_t CGSize, typename Hash>
+template <typename ProbeKey, typename Extent>
+__host__ __device__ constexpr auto perfect_probing<CGSize, Hash>::operator()(
+  cooperative_groups::thread_block_tile<cg_size> const& g,
+  ProbeKey const& probe_key,
+  Extent upper_bound) const noexcept
+{
+  using size_type = typename Extent::value_type;
+  return cuco::perfect_probing<CGSize, Hash>::probing_iterator<Extent>{
+    cuco::detail::sanitize_hash<size_type>(hash_(probe_key) + g.thread_rank()) % upper_bound,
```
esoha-nvidia, line 208: Ah damnit.  The result of the hash, plus the thread_rank, *can* exceed the size of the hash table.  So we must modulo.  So here, you must keep the modulo...

Wait, no, the answer is that perfect probing doesn't need CGSize because it will always be `1`.  Remove it as a template param and get rid of all mention of cooperative groups in perfect probing.

Some background:

This whole cooperative groups thing in cuco is for inserting a single key with multiple threads.  The idea is that instead of having a single thread devoted to the insert of each key, you perhaps devote x threads per insert, like, maybe 4 threads per key.  And then all 4 of those threads check 4 slots simultaneously and then those 4 threads figure out which is the first one and only that one does the insert.  The idea is that, imagine you have a bunch of collisions, rather than scan through the slots one a time looking for an empty slot, you are scanning them four at a time.

But with perfect probing, there are no collisions!  So there's never a need to use more than one thread at a time.  Perfect probing with cooperative groups makes no sense.  Remove the option to use CG with perfect probing please.

include/cuco/hash_functions.cuh
```diff
@@ -67,6 +69,52 @@ using xxhash_32 = detail::XXHash_32<Key>;
 template <typename Key>
 using xxhash_64 = detail::XXHash_64<Key>;
 
+/**
+<<<<<<< HEAD
+=======
```
esoha-nvidia, line 74: oops!

include/cuco/probing_scheme.cuh
```diff
@@ -175,6 +177,159 @@ class double_hashing : private detail::probing_scheme_base<CGSize> {
...
+ **/
+template <typename T, typename = void>
+struct is_perfect_hashing : cuda::std::false_type {};
+
+/**
+ * @brief wrapper to test if a probing scheme is perfect
+ * @tparam T is a probing scheme that does have a method "is_perfect".
+ **/
+template <typename T>
+struct is_perfect_hashing<T, cuda::std::void_t<decltype(std::declval<T>().is_perfect())>>
```
esoha-nvidia, line 192: I don't think that it's sufficient to just check if a class has a method `is_perfect`, I would also like to check that `is_perfect` returns `true`!

Let's rename these to `has_is_perfect` and then make another function that is called `is_perfect_hashing` which does a constexpr check on the value of `has_is_perfect` and also a constexpr check on the result of `is_perfect`.  We want the check to be constexpr because we want the compiler to throw an error if we're unable to determine whether or not a hashing scheme is perfect at compile time.  We want to know it at compile time.  Run time is too late.  This will force the `is_perfect` method of the hashing functor to be `constexpr`, which is what we want.  You already did make it `constexpr` above so that'll work.

Just to be sure, I would try removing the `constexpr` of the `is_perfect` method and see that the compiler correctly throws an error.

We want, as much as possible, for the compiler to catch stuff like this.

include/cuco/probing_scheme.cuh
```diff
@@ -175,6 +177,159 @@ class double_hashing : private detail::probing_scheme_base<CGSize> {
...
+  : cuda::std::true_type {};
+
+/**
+ * @brief Public perfect probing scheme class.
+ *
+ * @note Perfect hash functions guarantee no collisions. User is responsible for supplying a perfect
+ * hash
```
esoha-nvidia, line 199: You can use a `static_assert` in the code to guarantee that the hash function provided is indeed perfect.  `static_assert` doesn't need to be inside a function, either.  You could put it right before the `public:`.

https://stackoverflow.com/questions/25904145/static-assert-inside-outside-class-definition

```
static_assert(is_perfect_hashing<Hash>::value);
```

Like that maybe?

include/cuco/probing_scheme.cuh
```diff
@@ -175,6 +177,159 @@ class double_hashing : private detail::probing_scheme_base<CGSize> {
...
+   *
+   * @tparam NewHash New hasher type
+   *
+   * @param hash Hasher
+   *
+   * @return Copy of the current probing method
+   */
+  template <typename NewHash>
+  [[nodiscard]] __host__ __device__ constexpr auto with_hash_function(
+    NewHash const& hash) const noexcept;
```
esoha-nvidia, line 237: Also here, static_assert that the new hash function is perfect.  this will *not* catch users that provide a perfect hash function that has a range which is bigger than the number of slots in the hash table!  That's up to the user to be smart enough to do that properly.  We should probably put it in the doxygen that the hash function should have a range smaller than the number of slots in the hash table and it's up to the user to ensure that.  We can't ensure on the user's behalf.

We could do some tricks to help the user out.  For example, we could have the user provide, along with the hash, the maximum possible value of the hash.  And then we would do a one-time check on the CPU to see that the maximum possible output of the hash is smaller than the number of slots.  This is just taking a very small check that the user is going to have to do anyway and forcing it to happen by forcing the user to provide us with the inputs to do it.  Is that a good idea?  I dunno, maybe?  It may prevent user errors.  And the user is going to have to know that maximum value anyway in order to construct a hash table that is large enough.

If you want, this could be a later refinement.  It would cause the user to get a runtime error explaining that perfect probing was used incorrectly, rather than the code silently corrupting memory.

In my implementation, in addition to the `is_perfect` method, I have a `max()` method that returns the maximum possible output of a hash function.  I use that metaprogramming trick from above to detect if a hasher has a `max()` method.  If the method exists, return that value.  If it does not then return [`std::numeric_limits<T>::max`](https://en.cppreference.com/w/cpp/types/numeric_limits/max).  `max()` is not constexpr.  Maybe call it `range_max` to differentiate it from being interpreted as the max of the domain?

include/cuco/probing_scheme.cuh
esoha-nvidia: Never.  Good catch @PointKernel !

If this function is getting called called, there is a programming error at the call site?  Hmm, except for multisets and multihashs...

I could see how this function might get called if the cooperative group size is not 1.  perfect hashing should disallow cooperative groups, too.

I like the `static_assert` here instead of deleting the function.  It provides more information to the programmer.  I would like more detail, either in the blurb or in a comment, about why perfect probing doesn't even need incrementing.

```cuda
static_assert(false, "perfect probing doesn't support incrementing because incrementing is only necessary when a slot is found to be full but not matching, which is impossible");
```

Go through the code and figure out why this is getting called.  If it's due to the cooperative groups stuff then those test cases must be deleted.  perfect probing is not compatible with CG.  If it's due to multiset or multihash then delete those test cases, too.  Not compatible.

For all other cases, there is a bug in the code and it should be fixed.

include/cuco/probing_scheme.cuh
```diff
@@ -175,6 +177,159 @@ class double_hashing : private detail::probing_scheme_base<CGSize> {
...
+    }
+
+    /**
+     * @brief Postfix increment operator
+     *
+     * @return Old iterator before increment
+     */
+    __host__ __device__ constexpr auto operator++(int32_t) noexcept
+    {
+      auto temp = *this;
```
esoha-nvidia, line 320: also static_assert that this is not allowed.

esoha-nvidia (changes requested):
Thanks for your efforts!

include/cuco/detail/hash_functions/sigbithash.cuh
PointKernel: OK, got it. In that sense, we need to figure out a way to guarantee that the sig hash is used exclusively with perfect probing.

include/cuco/detail/hash_functions/sigbithash.cuh
esoha-nvidia: Hmm, yeah, maybe it's okay.  I was thinking about the case with the row hasher like cuDF does but in that case, there's going to be an adaptor and the key will actually just be an index to a key...

Yeah, nevermind me here.  This is fine.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -391,6 +384,31 @@ class open_addressing_ref_impl {
...
+                             val)) {
+              case insert_result::DUPLICATE: {
+                if constexpr (allows_duplicates) {
+                  [[fallthrough]];
+                } else {
+                  return false;
+                }
+              }
+              case insert_result::CONTINUE: continue;
+              case insert_result::SUCCESS: return true;
```
wence-, line 409: This insertion part is identical to lines 373-386, I think? Can it be factored out?

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -1208,14 +1253,21 @@ class open_addressing_ref_impl {
                                                         value_type const& expected,
                                                         Value const& desired) noexcept
   {
-    if constexpr (sizeof(value_type) <= 8) {
-      return packed_cas(address, expected, desired);
+    if constexpr (is_perfect_hashing<decltype(probing_scheme_)>::value) {
+      // Do it without atomics.
+      *address = desired;
```
wence-, line 1258: *question*: Can it be the case that two different threads might try to `attempt_insert` at the same address without synchronisation? Because if yes, even if they are writing the same `desired` value, this is UB in the memory model.

wence- (commented):
One question

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
PointKernel: @wence- is right, this should be atomic store.

include/cuco/detail/hash_functions/sigbithash.cuh
esoha-nvidia: Well, kind of.  To be more accurate, we need to make sure that a perfect probing scheme is **always** used with a hasher that is perfect.  For example, use **can** use `SigBitHash` with perfect probing but you can also use it with double probing or linear probing, no problem!  What we want to avoid is someone trying to use a non-perfect hash together with perfect probing.  Like perfect probing with xxhash.  That's not allowed, it will give invalid results or crash!

I'll try to write up the whole idea here:

# Background

A hash table, when constructed, gets a probing scheme that is a [functor](https://en.wikipedia.org/wiki/Function_object).  For example, linear or double or perfect, etc.  The probing scheme describes how the hasher will scan through the slots looking for a place to insert/lookup.

That probing scheme has inside it a hasher which is also a functor.  That hasher might be xxhash or murmur3 or Identity or SigBitHash, etc.  The hasher does the math to help the probing scheme get the slot to look at.

If:

* the hasher is perfect (`hash(x) == hash(y) => x==y`) **and**
* the maximum output of the hasher's `operator()` is less than the number of slots in the hash table **and** 
* the minimum output of the hasher's `operator()` is at least 0

then the hashtable or hashset can:

* Skip the use of atomics in most cases (because all collisions will just be the same value).
* Run a faster equality operator (because all collisions are sure to be equal).
* Skip the modulo operator (because the hasher's output is always less than the slot count).

These will give a speed improvement.

# Goal

We want to make it as easy as possible for the programmer using cuco to get this right and not screw it up.

# Solutions

## Detecting perfect probing scheme.

One way is that we could accomplish this is to have a prober class that is called `perfect_probing`, just like we already have one called `linear_probing`.  And in the code for the hash table, in the places where we use modulo or check equality or use atomics, we just do a little `if constexpr` to check if the type of the probing class is `perfect_probing`.  And if it is, then we skip those and do things the fast way.

The problem with this is that, what if the user wants to invent their own perfect probing scheme?  Because we are testing against a specific class `perfect_probing` then the user won't be able to create a class called `my_perfect_probing` and use it as if it is perfect.  This is not good.

So instead, we put a method inside the `perfect_probing` class that is called `is_perfect` that returns `true` only for perfect probing schemes.  And the `if constexpr` above is now testing that the probing scheme's `is_perfect` method returns `true`.

```c++
if constexpr (this.probing_scheme_.is_perfect()) {
  // don't use atomics
} else {
  // do it the normal way with atomics.
}
```

### Detecting if a class has a method.

The problem with this is that the other probing schemes don't even have a function called `is_perfect` so they will fail to compile.

```c++
 // If this.probing_scheme_ is linear probing, compile-time error.
if constexpr (this.probing_scheme_.is_perfect()) {
  // don't use atomics
} else {
  // do it the normal way with atomics.
}
```


But we can solve this using an idea similar to [c++ type traits](https://accu.org/journals/overload/9/43/frogley_442/).  We can use [SFINAE](https://en.wikipedia.org/wiki/Substitution_failure_is_not_an_error) to [detect if a class has a method at compile time](https://stackoverflow.com/questions/87372/check-if-a-class-has-a-member-function-of-a-given-signature).  It is done [here](https://github.com/NVIDIA/cuCollections/pull/505/files#diff-2d1ed9c31db1250809c8d1f33971551ab468bc7521f037c2e61291848158ad1aR184-R192).

So the test is now, if:
* the hash table's probing scheme has the method `is_perfect` **and**
* the `is_perfect` method returns `true`

then:
* do all the perfect hashing optimization.

else:
* do regular hash table stuff.

This will allow anyone to make a new perfect probing scheme just by defining `is_perfect()` to return `true`.  It also means that we *don't* have to put an `is_perfect` method into all the other probing schemes because they all get treated as if their `is_perfect` returns `false`.

We're going to need to do this a few times so we make a utility that can check if a class has a method and, if so, it calls that method and returns the result.  If it *doesn't* have the method then it returns a default value that we provide.   A code example is below and you can see it [on godbolt here](https://godbolt.org/z/5nnYM3ccW).

<details>
  <summary>Code to call a class method and use a default if the method doesn't exist.</summary>

```c++
#include <stdio.h>

/*** The utility code ***/
template <typename T, typename = void>
struct has_foo_method_helper : std::false_type {};

template <typename T>
struct has_foo_method_helper<T, std::void_t<decltype(std::declval<T>().foo())>> : std::true_type {};;

template <typename T>
bool constexpr has_foo_method() {
    return has_foo_method_helper<T>::value;
}

template <typename T>
bool constexpr call_foo_with_default(bool default_value) {
    if constexpr (has_foo_method<T>()) {
        return T::foo();
    }
    return default_value;
}

/*** Example ***/

struct HasFooTrue {
    static bool constexpr foo() { return true; }
};

struct HasFooFalse {
    static bool constexpr foo() { return false; }
};

struct DoesNotHaveFoo {
};

int main() {
  printf("%d\n", call_foo_with_default<HasFooTrue>(false));
  printf("%d\n", call_foo_with_default<HasFooFalse>(false));
  printf("%d\n", call_foo_with_default<DoesNotHaveFoo>(false));

  printf("%d\n", call_foo_with_default<HasFooTrue>(true));
  printf("%d\n", call_foo_with_default<HasFooFalse>(true));
  printf("%d\n", call_foo_with_default<DoesNotHaveFoo>(true));

  return 0;
}
```
</details>

We can do this for any method.  We'll call functions using this technique `call_XXXX_with_default` and `has_XXXX_method`.

The hash table, to check if we can skip atomics or modulo or equality comparison, will use:

```c++
if constexpr (call_is_perfect_with_default<probing_scheme>(false)) {
  // it is perfect!  don't use atomics, modulo, etc
} else {
  // not perfect hashing so do it the normal way, with atomics, modulo, etc
}
```

## Checking that the hasher is perfect.

With the above done, now we can write code that will work one way for perfect probing and a different way for linear and double probing.  And we don't need to modify our other probing schemes.  If a user wants to make a new probing scheme, they can make it the normal way and it will not be perfect.  Or they can define `is_perfect` to return `true` and it will use the optimizations of perfect hashing.  It's all pretty good.

The next thing that we need to check is that the hasher provided to the perfect probing scheme is indeed perfect.  If the user constructs a `perfect_probing` instance and supplies to it `xxhash`, which isn't perfect, it'll fail!

One solution to this would be to just have a comment in the docs that says that if you use a `perfect_probing`, make sure you also use a perfect hasher functor.  But we can do better.

What we'll do is put a method in the perfect hashers (such as `SigbitHash` and `IdentityHash`) called `is_perfect` that returns `true`.  And in the constructor of `perfect_probing`, we will have:

```c++
static_assert(call_is_perfect_with_default<Hasher>(false));
```

With this, if the user accidentally tries to construct a `perfect_probing` with a hasher that is not perfect, it'll be a compile time error.

## Checking that the range of the hasher is less than the number of slots.

With the techniques above, we are now ensuring that only when the probing is perfect do we skip atomics/modulo/etc.  And the perfect probing always has a perfect hasher.  The last thing to check is that the max output of the hasher is less than the number of slots.  In mathematics, the inputs to a function is called its "domain" and the output is the "range".  So we'll say that we want the `range_max` of the hasher to be less than number of slots in the hash table.

For this, the hasher will have a method called `range_max` that returns that maximum possible value of the output of the hasher's `operator()`.  For hashers that don't have the `range_max` method, we'll just assume that the maximum is the maximum of the output type of `operator()`.  We can get that maximum using [code like this on godbolt](https://godbolt.org/z/G3qvjjq9z).

<details>
  <summary>Code to get the return type max of `operator()`.</summary>

```c++
#include <stdio.h>
#include <type_traits>
#include <limits>

struct FooReturnsInt {
    int operator()() { return 0; }
};

struct FooReturnsLong {
    long operator()() { return 0; }
};

int main() {
  printf("%lld\n", std::numeric_limits<std::invoke_result_t<FooReturnsLong>>::max());
  printf("%lld\n", std::numeric_limits<std::invoke_result_t<FooReturnsInt>>::max());

  return 0;
}
```
</details>

Let's call that `get_invoke_result_t_max<T>()`.  So now we will add to the constructor of our hashtable code like this:

```c++
if constexpr (call_is_perfect_with_default<probing_scheme>(false)) {
  // Okay, this is a perfect probing scheme.  Is the range of the hasher valid?
  constexpr auto max_of_output_type = get_invoke_result_t_max<probing_scheme>();
  if (call_range_max_with_default<probing_scheme>(max_of_output_type) >= slots_.size()) {
    // Uh oh, error, the perfect probing scheme is valid
    // but this hash table doesn't have enough slots to use it!
  }
}
```

And the `probing_scheme::range_max` function will call its hasher's `range_max` function.  

With this, now we can check at the time of constructing the hash table that the probing scheme provided will indeed use a hasher that does not return values greater than the number of the slots.  This lets us get rid of the modulo operator and that is necessary for removing atomics, too.

All the above should also be done for `range_min`.  (However, `range_min` is usually not going to be defined because the `operator()` functions will usually return an unsigned value and the `call_range_min_with_default` will already be returning a default of 0, which is correct.  So we need the test but the hasher probably doesn't need to define `range_min` for unsigned outputs.)

## Reporting a range_max error.

We won't know the `range_max` of a hasher at compile time so we can only do the `call_range_max_with_default` at runtime, and the `range_max` functions are not `constexpr`, either, for the same reason.

The test above happens inside the constructor of the hash table.  Constructors have only one way to return errors, by throwing exceptions.  It's [perfectly acceptable to throw an exception in a constructor.  ](https://isocpp.org/wiki/faq/exceptions#ctors-can-throw).  We can use the existing `CUCO_FAIL` macro for this.

```c++
if constexpr (call_is_perfect_with_default<probing_scheme>(false)) {
  // Okay, this is a perfect probing scheme.  Is the range of the hasher valid?
  constexpr auto max_of_output_type = get_invoke_result_t_max<probing_scheme>();
  auto max_hasher_output =
      call_range_max_with_default<probing_scheme>(max_of_output_type);
  if (max_hasher_output >= slots_.size()) {
    std::stringstream error_string;
    error_string << "Can't use perfect probing because the number of slots is only "
                 << slots_size()
                 << " but the hasher may return the value "
                 << max_hasher_output
                 << " which was determined because ";
    if (has_range_max<hasher>()) {
      error_string << "we called range_max and that's what it told us"
    } else {
      error_string << "there's no range_max method so we assumed the maximum of the data type"
    }
     CUCO_FAIL(error_string.str(), std::runtime_error);
  }
}
```

(The exception message above is extra verbose just to make clear the logic.)

As before, do the same for `range_min`.

## Avoiding the runtime error.

In practice, a smart user of cuco's perfect hashing is never going to see this exception because they are going to first construct the hasher and then look at the range_max of it and use that info to determine if they want to use perfect hashing or not.  In psuedocode, here's what the user will do:

```c++
vector <int> data_to_insert;
SigBitHash my_hasher(data_to_insert); // The SigBitHash looks at the data to make a perfect hash.
auto range_max = my_hasher.range_max();
constexpr auto max_hash_memory = 1000000; // Based on how much memory we want to allow our hash table to use.
if (range_max < max_hash_memory) {
  // Cool!  We can use perfect hashing!
  auto my_probing_scheme = perfect_probing(my_hasher);
  // remember to add one because of fencepost error https://en.wikipedia.org/wiki/Off-by-one_error
  auto my_hash_table = hash_table(my_probing_scheme, range_max + 1);
  // use the hash table
} else {
  // Okay, we'll just use regular hashing.
  auto my_probing_scheme = linear_probing(my_hasher);
  auto occupancy = 2; // We'll have twice as many slots as item to insert.
  auto my_hash_table = hash_table(my_probing_scheme, data_to_insert.size() * occupancy);
  // use the hash table
}
```

To make it ever tidier, you can put the above into a function.  Instead of using the hash table, just return the hash table.  The return type of the function will have to be `std::variant<...>` where the variants are either a hash table with perfect probing or a hash table with linear probing.  And then you can use the `std::visit` with a [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) to do `insert` and `lookup` on the hash table.

Or if you don't like the visitor pattern, just make a function called `make_perfect_hash_table` that has return type `std::optional<...>` that returns a perfect hash table when `range_max < max_hash_memory` and otherwise returns `std:nullopt`.  And then call `make_perfect_hash_table`.  If it returns a hash table, use it.  If not, fall back to using linear probing.

Both of these solutions will result in the same compiled asm code but it's up to the user which one seems more ergonomic.  Doesn't really matter, they work the same.

# Conclusion

With all of this, we'll have a `perfect_probing` that gets us all the optimizations we want and provides a lot of safety so that the user won't use perfect_probing incorrectly.  None of the existing probing schemes and hash functions need to be modified and if the user doesn't want to use perfect probing, they don't have to change anything!

If the user *does* want to use perfect probing, they can.  It will do all the compile-time and run-time checks to make sure that that the hasher is valid to use and the hash table size is correct.  The only overhead will be a single test for `range_max < slots_.size()`, once, on the CPU, at construction time.  It's trivially fast.  That test throws an exception but they can avoid it by doing the test themselves before construction of the hash table.  In fact, it'll be the same test so the optimizer will optimize it away and remove internal overhead in the constructor!

If the user wants to design their own hasher that isn't perfect, they can do it like they are already doing it.

If they want to design a perfect one, they just have to remember to put the `is_perfect` method in there that returns `true`.  And if they forget to do that but try to use it with `perfect_probing`, they will get a compile-time error that explains the problem.

If they forget to define `range_max` in their perfect hasher then the default will be the maximum of the output type.  If that's too big then they will get a runtime error explaining the that the range is too big for the number of slots and they can fix that.  

## Possibilities for errors

It's pretty foolproof.

One way to break it would be if the user defines a hasher that isn't perfect and they put in it a method called `is_perfect` that returns `true`.  I mean, that would be pretty unlikely to do on accident!

Another way to break it is if they define the `range_max` function of the hasher and then return a max is the smaller than what the hasher's `operator()` sometimes returns.  This is only possible if the user's hash function is broken.

To prevent these, we'll have comments in the code around these functions that explain their purpose so that a user who is designing a perfect hasher will know how to write the above functions correctly.

# Notes

* Perfect hashing is only possible for hashset and hashtable.  It won't work for multihash nor multiset.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
esoha-nvidia: Yes, @wence- is right.  It is UB, even if `desired` is the same for all threads writing to the same location.

It's kind of like jaywalking in New York.  Technically, it's illegal.  In practice, everyone is doing it.  If you want, you can avoid jaywalking and stand at the corner and tell everyone next to you, "Look, I am not a criminal because I wait here at the corner!"  But no one will hear you say that because all of them are on the other side of the road already.

Let's look at the pros and cons of the options:

* Don't use atomics
  * Pros:
    * Hash table is now way faster.
    * It works.
  * Cons:
    * It uses UB, which works now and lots of people do it but maybe in the future it won't.
* Use atomics
  * Pros:
    * No UB.  The code is guaranteed to keep working into the future.
    * It works.
  * Cons:
    * The hashing is slower than the competitors so people stop using cuco.

I'm not sure how to deal with this issue.  If we want to be very strict and avoid UB, we can.  And we'll be able to tell all the users that there is no UB in cuco.  But maybe there won't be any users because they all went to go use a faster solution made by the competitor.  (I am the one that made that competitor's solution, ironically.)

There is still some benefit.  Even without removing the atomic here, we still have the removal of the modulo operator and the faster equality operation.  Those are kind of the side-show, though.  The main attraction is removing atomics.

I propose:
1. Get the perfect hashing and perfect probing PRs put into cuco, without this UB but with the modulo and equality optimizations, then
2. Test the performance with/without the UB.
  1. If the performance difference **is not** significant, don't do UB.
  2. If the performance difference **is** significant then...?

What if we let the user decide?  It would be easy to provide a template parameter to the perfect probing called `ub_allowed` that is by default `false`.  And if the user is willing to take the risk then let them.  This makes code that by default has no UB but also cuco's users don't have to go switch to the competitor's solution when they want more performance.

---

It'll be a while until we can measure this.  In the meanwhile, can we get an estimate on how big the performance difference is?  My [TPC-H](https://www.tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) (Q1, page 29) test case is doing 300,000,000 inserts into a hash table where the cardinality of the data is only 4.  (This is actually considered a small test case for TPC-H).  At best case, there are 75,000,000 `atomicCAS` operations going into the same memory location.  We could measure that with `atomicCAS` versus simple write and see what happens.

We should check with all `desired` equal and also all `desired` not equal because, for the `row_hasher` that databases like cuDF use, it's going to be all not equal.

wence-: > This makes code that by default has no UB but also cuco's users don't have to go switch to the competitor's solution when they want more performance.

I am not a cuco core dev, so perhaps my opinion here carries no real weight, but I would find it very hard to justify a mode in which you can use the library where you can provide _no guarantees_ about whether a result is correct. Particularly, just because you can check that the "no atomics" code works for some test case, that is not a guarantee that it is producing the correct code for any other execution (even of the same test case again).

wence-: FWIW, the cheapest valid thing to do here is, I think, to use a `relaxed` store. This is a strong (in the memory order sense) operation, rather than a weak one, and therefore removes any UB: if there are different values of `desired` being written, a relaxed store says "I don't care which one I see, or in which order", but I have to see one of them. Whereas a standard store doesn't give the guarantee that you have to see one of the writes.

esoha-nvidia: Let's first measure with a relaxed store and see how it goes.

> I would find it very hard to justify a mode in which you can use the library where you can provide no guarantees about whether a result is correct

It's possible that you'll lose customers if you don't at least provide the option because the customer already has the faster version that doesn't use atomics.  Why would they switch to slower code?

wence-: > [...] faster version that doesn't use atomics. Why would they switch to slower code?

Perhaps I am too idealistic, but to me, a library which produces the wrong answer (sometimes, you can't know!) very fast is about as useful as a chocolate teapot.

benchmarks/benchmark_defaults.hpp
njones93531: Will do!

benchmarks/hash_table/static_set/insert_bench.cu
njones93531: Not necessary, this is a testing leftover.

examples/other/identity_test.cu
njones93531: Sounds good!

examples/static_set/host_bulk_example.cu
njones93531: Agreed

examples/static_set/host_bulk_perfect_example.cu
njones93531: Yes, this can be more explicit

examples/static_set/host_bulk_perfect_example.cu
njones93531: Sounds good!

include/cuco/detail/hash_functions/identityhash.cuh
njones93531: Broken out into [PR 514]( https://github.com/NVIDIA/cuCollections/pull/514)

include/cuco/detail/hash_functions/utils.cuh
njones93531: Makes sense, thanks!

include/cuco/probing_scheme.cuh
njones93531: Sounds good!

include/cuco/detail/hash_functions/sigbithash.cuh
njones93531: Agreed, this will be split off in the future.

include/cuco/detail/hash_functions/sigbithash.cuh
njones93531: Agreed

include/cuco/detail/hash_functions/sigbithash.cuh
njones93531: This makes sense. In this case, would it make sense to take both an array of pointers to each of the lists, as well as an array of sizes for each of the lists?

include/cuco/detail/hash_functions/utils.cuh
njones93531: Sounds good!

include/cuco/detail/hash_functions/utils.cuh
njones93531: Sounds good! I'll check to stay consistent with cuCo

include/cuco/detail/hash_functions/utils.cuh
njones93531: > Also, we'll often want the reduction over multiple arrays. You can do it in two phases: First one reduction per input to generate a single AndOrPair per input.

This makes sense. I will certainly ask about the rest as I get back into this functionality

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
njones93531: Sounds like a good way to stay DRY! I'll work on it

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
njones93531: Sounds good!

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
njones93531: This makes sense. I would much rather write this branch once!

include/cuco/detail/probing_scheme_impl.inl
njones93531: Yes, modulo should only matter if the `perfect` hash isn't perfect after all. I will happily remove it! It is especially unnecessary after the protections you mention [here](https://github.com/NVIDIA/cuCollections/pull/505#discussion_r1646605810)

include/cuco/detail/probing_scheme_impl.inl
njones93531: Thanks for the CG background! It does make sense to remove all CG references from `perfect probing`.

include/cuco/hash_functions.cuh
njones93531: Good catch!

include/cuco/probing_scheme.cuh
njones93531: I do like this idea, and the details added in your other [comment](Wait, no, the answer is that perfect probing doesn't need CGSize because it will always be 1. Remove it as a template param and get rid of all mention of cooperative groups in perfect probing.).

include/cuco/probing_scheme.cuh
njones93531: Sounds good!

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
njones93531: Good catch, yes it can!

benchmarks/hash_table/static_set_perfect/contains_bench.cu
njones93531: This makes a lot of sense!

benchmarks/hash_table/static_set_perfect/insert_bench.cu
njones93531: Sounds good!

benchmarks/hash_table/static_set_perfect/insert_bench.cu
njones93531: Makes sense, I'll make the names more specific.

include/cuco/detail/hash_functions/sigbithash.cuh
PointKernel: To simplify the problem, let's narrow down the discussion to some key points:

- From cuco's perspective, we don't allow user-defined probing schemes
- Whether it's perfect or not, that's the information carried by probing scheme, not hasher
- The steps checking perfect hasher, checking the range, reporting range max error and avoiding runtime errors should be the responsibility of a helper function like `create_perfect_hashing` whose return type is an optional: if the input keys can be hashed perfectly, returns the perfect probing. Otherwise, returns `nullopt`.  
- `perfect_probing` ctor should be private to avoid misuse and the only way to get a perfect probing is via the above helper function. By doing so, we can avoid all the potential errors mentioned in your `Possibilities for errors` section

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
PointKernel: Relaxed atomic store is the right way to go.

No library wants to tolerate undefined behavior.

examples/other/identity_test.cu
njones93531: Removed file entirely as its purpose (testing validity of identity_hash) is better served by the unit tests in PR #514 .

benchmarks/hash_table/static_set_perfect/contains_bench.cu
```diff
@@ -0,0 +1,90 @@
+/*
```
esoha-nvidia, line 1: Is it possible to use the existing file `../static_set/contains_bench.cu` and then template it so that it receives the probing scheme as a parameter?  And then add to `NVBENCH_TYPE_AXES` the probing scheme?

I don't know anything about how nvbench works.  Maybe it's not possible?  Or maybe the output would look too wonky?

benchmarks/hash_table/static_set_perfect/insert_bench.cu
```diff
@@ -0,0 +1,87 @@
+/*
```
esoha-nvidia, line 1: Again, can we template `../static_set/insert_bench.cu`?

examples/static_set/host_bulk_perfect_example.cu
esoha-nvidia: Remind me please.  Is this because creating the perfect prober is potentially impossible so we want to have a `create_perfect_probing` function that returns `std::optional<.....>`?

You could also make the constructor public and just have it throw in case of failure.  That's a normal way for constructors to fail.

Or is there another good reason?  I don't remember.

examples/static_set/host_bulk_perfect_example.cu
```diff
@@ -0,0 +1,78 @@
+/*
```
esoha-nvidia, line 1: Unlike in the benchmark code, for example code, I like the idea of having this copy of the file rather than use templates.  Templates are harder to read and examples should be kept simple.  So I'm cool with this file, even though it looks very similar to the other `host_bulk_example.cu`.

include/cuco/detail/hash_functions/identityhash.cuh
esoha-nvidia: If this is in another PR then it is, for this PR, resolved, yes?

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,144 @@
...
+  template <typename Extent>
+  constexpr result_type __host__ __device__ compute_hash(std::byte const* bytes,
+                                                         Extent size) const noexcept
+  {
+    std::uint32_t h1 = load_chunk<std::uint32_t>(bytes, 0);
+    return h1;
+  }
+
+ private:
+  std::uint32_t seed_;
```
esoha-nvidia, line 82: This is unused, right?  Let's delete it.

If the compiler complains that you didn't use a variable in the constructor, the solution is to leave it unnamed.

```cuda
 __host__ __device__ constexpr IdentityHash_32(std::uint32_t seed = 0) : seed_{seed} {}
```

becomes:

```cuda
 __host__ __device__ constexpr IdentityHash_32(std::uint32_t) {}
```

If you like.  Better yet, get rid of that argument entirely!

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * @brief An Identity hash function to hash the given argument on host and device.
+ *
+ * -----------------------------------------------------------------------------
+ *
+ * @tparam Key The type of the values to hash
+ */
+template <typename Key>
+struct IdentityHash_64 {
```
esoha-nvidia, line 93: This is the same as the previous one just with a different type.

Let's unify them and add a param:

```cuda
template <typename Key, typename Result>
struct IdentityHash {
```

And then replace all the `std::uint64_t` below with `Result`.  And then you can just have a single struct instead of one for 32-bit and one for 64-bit.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -369,14 +369,7 @@ class open_addressing_ref_impl {
...
       for (auto& slot_content : window_slots) {
-        auto const eq_res =
-          this->predicate_.operator()<is_insert::YES>(key, this->extract_key(slot_content));
-
-        if constexpr (not allows_duplicates) {
-          // If the key is already in the container, return false
-          if (eq_res == detail::equal_result::EQUAL) { return false; }
-        }
-        if (eq_res == detail::equal_result::AVAILABLE) {
+        if constexpr (is_perfect_hashing<decltype(probing_scheme_)>::value) {
```
esoha-nvidia, line 372: I would like to check if the function `is_perfect` exists, which you are doing here, and *also* check that `is_perfect()` returns `true`.  Because we want to allow someone to define the function as false.  Like, say you are inheriting from a class that defines the function.  You can't easily undefine it.  So you might choose to override the inheritance with a version that returns false.

(Technically, you could also do `private` inheritance so that the `is_perfect` function doesn't get defined in the class that extends but that also has other side-effects that the user might not want!)

esoha-nvidia (changes requested):
A few more comments.  I realize that this is just the WIP code and we'll actually be doing the work in other, smaller PRs.  I just wanted to capture the ideas here.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
esoha-nvidia: First, let's measure the performance difference between relaxed atomic store and simply writing.

If it's not very different then we can just use relaxed atomic store.

If the difference is significant then you'll have to decide if you even want to provide the clients of cuco the option to do UB.  Keep in mind that if you don't provide the option, they might not consider using cuco.

include/cuco/probing_scheme.cuh
esoha-nvidia: Actually, you can't `static_assert` in this function because that will just cause compilation to fail even if the function isn't called.  I think that you'll actually just be deleting the function.  If you want, you can put a comment about the fact that incrementing is not allowed.

If the fact that this function is missing causes a compilation error then it is the caller that is wrong.

PR #506, opened 2024-06-15
Add static_multiset::for_each and its OA impl

### LINKED ISSUE #499: [ENHANCEMENT]: Add `for_each` device API

There's currently no way of iterating over all matching values for a given key from the device.
The current `retrieve` API isn't exposed to the public since it is highly optimized for bulk `retrieve` and thus not well-suited for general use.

Implement a device function that allows for iterating over all values found and then process them according to some user-defined callback function:

```cpp
template <class CG, class ProbeKey, class Callback>
void for_each(CG tile, ProbeKey key, Callback f) {...}
```

During probing for `key`, if a thread in `tile` hits a matching slot, it will pass a non-incrementable iterator to said slot to the callback which then handles the downstream processing.

This API can also be used to implement the aforementioned optimized `retrieve`.

`for_each` handles all of the probing logic internally and just calls the callback with the found values.
The counter design to this would consist of a "heavy" iterator which is returned from a new `retrieve` device function`
The iterator will handle the probing internally, returning `end()` once there are no more matching values for the given `key`.
I have some concerns with this design:

- If we use window loads, i.e., load multiple slots per thread at once, then the iterator has to store this window internally
- The iterator needs to be kept in the same CG context. If we can't ensure that the iterator is incremented in a CG context we might loose some of the values. There's no way we can ensure this never happens with this design. The `for_each` function, however, ensures that probing is always performed in the same CG context

### Additional context

WarpCore had a `for_each` function although my design back in the day wasn't perfect: https://github.com/sleeepyjack/warpcore/blob/ba5a6a84a8797464e2ce8dd03b269907ee70e7c2/include/warpcore/multi_value_hash_table.cuh#L607

sleeepyjack:
closes #499

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
@@ -970,13 +970,13 @@ class open_addressing_ref_impl {
    * `key` to the callback.
    *
    * @tparam ProbeKey Input type which is convertible to 'key_type'
-   + @tparam Callback Callback functor or lambda
+   * @tparam CallbackOp Unary callback functor or device lambda
    *
    * @param key The key to search for
    * @param callback Function to call on every element found
    */
-  template <class ProbeKey, class Callback>
-  __device__ void for_each(ProbeKey const& key, Callback&& callback) const noexcept
+  template <class ProbeKey, class CallbackOp>
+  __device__ void for_each(ProbeKey const& key, CallbackOp&& callback) const noexcept
   {
     static_assert(cg_size == 1, "Non-CG operation is incompatible with the current probing scheme");
     auto probing_iter = this->probing_scheme_(key, this->storage_ref_.window_extent());
@@ -1014,16 +1014,16 @@ class open_addressing_ref_impl {
    * each thread with a match will call the callback with its associated element.
    *
    * @tparam ProbeKey Input type which is convertible to 'key_type'
-   + @tparam Callback Callback functor or lambda
+   * @tparam CallbackOp Unary callback functor or device lambda
    *
    * @param group The Cooperative Group used to perform this operation
    * @param key The key to search for
    * @param callback Function to call on every element found
    */
-  template <class ProbeKey, class Callback>
+  template <class ProbeKey, class CallbackOp>
   __device__ void for_each(cooperative_groups::thread_block_tile<cg_size> const& group,
                            ProbeKey const& key,
-                           Callback&& callback) const noexcept
+                           CallbackOp&& callback) const noexcept
   {
     auto probing_iter = this->probing_scheme_(group, key, this->storage_ref_.window_extent());
     bool empty        = false;
--- a/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
@@ -963,6 +963,106 @@ class open_addressing_ref_impl {
   }
 
   /**
+   * @brief Executes a callback on every element in the container with key equivalent to the probe
+   key.
+   *
+   * @note Passes an un-incrementable input iterator to the element whose key is equivalent to
+   * `key` to the callback.
+   *
+   * @tparam ProbeKey Input type which is convertible to 'key_type'
+   + @tparam Callback Callback functor or lambda
+   *
+   * @param key The key to search for
+   * @param callback Function to call on every element found
+   */
+  template <class ProbeKey, class Callback>
+  __device__ void for_each(ProbeKey const& key, Callback callback) const noexcept
+  {
+    static_assert(cg_size == 1, "Non-CG operation is incompatible with the current probing scheme");
+    auto probing_iter = this->probing_scheme_(key, this->storage_ref_.window_extent());
+
+    while (true) {
+      // TODO atomic_ref::load if insert operator is present
+      auto const window_slots = this->storage_ref_[*probing_iter];
+
+      for (auto i = 0; i < window_size; ++i) {
+        switch (
+          this->predicate_.operator()<is_insert::NO>(key, this->extract_key(window_slots[i]))) {
+          case detail::equal_result::EMPTY: {
+            return;
+          }
+          case detail::equal_result::EQUAL: {
+            callback(const_iterator{&(*(this->storage_ref_.data() + *probing_iter))[i]});
+            if constexpr (allows_duplicates) {
+              continue;
+            } else {
+              return;
+            }
+          }
+          default: continue;
+        }
+      }
+      ++probing_iter;
+    }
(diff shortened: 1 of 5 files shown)


sleeepyjack:
whelp something is not right. let me clean up the git history.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -962,6 +962,106 @@ class open_addressing_ref_impl {
...
+   * `key` to the callback.
+   *
+   * @tparam ProbeKey Input type which is convertible to 'key_type'
+   + @tparam Callback Callback functor or lambda
+   *
+   * @param key The key to search for
+   * @param callback Function to call on every element found
+   */
+  template <class ProbeKey, class Callback>
+  __device__ void for_each(ProbeKey const& key, Callback callback) const noexcept
```
```suggestion
  __device__ void for_each(ProbeKey const& key, Callback&& callback) const noexcept
```

Unsure if this needs to be a mutable or even universal reference instead. Let's say we define a `count` functor as such:

```cpp
struct count_functor {
  std::size_t thread_count = 0;

  // counts the number of matching elements for this thread
  template <class InputIt>
  __device__ void operator()(InputIt) { thread_count++; }
};
```

And then call
```cpp
//...
auto thread_counter = count_functor{};
set.for_each(key, thread_counter);
auto const key_count = thread_counter.count;
//...
```

Then we want the functor to be taken as a mutable reference, right?

include/cuco/operator.hpp
```diff
@@ -62,6 +62,12 @@ struct count_tag {
 struct find_tag {
 } inline constexpr find;  ///< `cuco::find` operator
 
+/**
+ * @brief `for_each` operator tag
+ */
+struct for_each_tag {
+} inline constexpr for_each;  ///< `cuco::for_each` operator
```
PointKernel, line 69: I see `for_each` as an internal utility as opposed to an actual hash table operator. Need to think more on this.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -962,6 +962,106 @@ class open_addressing_ref_impl {
...
+   *
+   * @tparam ProbeKey Input type which is convertible to 'key_type'
+   + @tparam Callback Callback functor or lambda
+   *
+   * @param group The Cooperative Group used to perform this operation
+   * @param key The key to search for
+   * @param callback Function to call on every element found
+   */
+  template <class ProbeKey, class Callback>
+  __device__ void for_each(cooperative_groups::thread_block_tile<cg_size> const& group,
```
sleeepyjack, line 1026: Not sure why the unit test is failing. Seems like the logic in this function is flawed.

include/cuco/operator.hpp
sleeepyjack: From my standpoint I would treat it as an extension to the STL API that is more suitable for the GPU. Having a "cooperative iterator" instead, which would be closer to the spirit of modern C++ has its drawbacks. For example, how do we ensure users only increment the iterator with the same CG? `for_each` solves this problem by making the probing part internal. We should even be able to redefine any lookup function (`find`, `count`, `retrieve`) that relies on probing with `for_each`, giving us a proper abstraction layer for probing.

sleeepyjack: On a side note I personally find this funtional approach, i.e., "for each found key do X" very appealing. Historic evidence that it is indeed useful comes from warpcore, where many downstream applications (mostly genomics stuff) implemented their custom lookup operations through `for_each` functors.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
PointKernel: pass by value is preferred.

The above example is a good example of a bad callback, especially in a parallel context

sleeepyjack: Is it best practice to pass a callback by-value? I'd have to skim some stackoverflow/cppreference pages to get familiar with the topic. With pass-by-value we lose the ability of giving the callback an internal state that can hold the result of the operation. How would we solve the above example with a callback passed by-value? Pass a pointer to `thread_count` to the callback?

PointKernel: > Pass a pointer to thread_count to the callback?

Yes

I see a callable defining the operations to be performed on the output instead of being the output itself.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
sleeepyjack: I think I found the problem. #509 should fix the issue.

sleeepyjack:
/ok to test

WTH all of my commits are signed...

sleeepyjack:
effing bot

PointKernel (approved):
LGTM

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -962,6 +962,164 @@ class open_addressing_ref_impl {
...
+   * `key` to the callback.
+   *
+   * @tparam ProbeKey Input type which is convertible to 'key_type'
+   * @tparam CallbackOp Unary callback functor or device lambda
+   *
+   * @param key The key to search for
+   * @param callback_op Function to call on every element found
+   */
+  template <class ProbeKey, class CallbackOp>
+  __device__ void for_each(ProbeKey const& key, CallbackOp&& callback_op) const noexcept
```
PointKernel, line 979:
```suggestion
  __device__ void for_each(ProbeKey const& key, CallbackOp&& callback_op) const
```
we need extra attention to deal with `noexcept` in the future since this function may throw.

include/cuco/detail/static_multiset/static_multiset_ref.inl
```diff
@@ -446,6 +448,114 @@ class operator_impl<
...
+   * `key` to the callback.
+   *
+   * @tparam ProbeKey Input type which is convertible to 'key_type'
+   * @tparam CallbackOp Unary callback functor or device lambda
+   *
+   * @param key The key to search for
+   * @param callback_op Function to call on every element found
+   */
+  template <class ProbeKey, class CallbackOp>
+  __device__ void for_each(ProbeKey const& key, CallbackOp&& callback_op) const noexcept
```
PointKernel, line 481:
```suggestion
  __device__ void for_each(ProbeKey const& key, CallbackOp&& callback_op) const
```

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
PointKernel: nonsense here. sorry about the confusion.

PR #509, opened 2024-06-19
Add unit test for scalar vs. CGSize1 probing

sleeepyjack:
This PR adds a unit test to verify that the probing sequences for the scalar probing iterator vs. the iterator that uses a cooperative group of size 1 are actually equivalent.

### DIFF
--- a/include/cuco/detail/extent/extent.inl
+++ b/include/cuco/detail/extent/extent.inl
@@ -41,6 +41,13 @@ struct window_extent {
   friend auto constexpr make_window_extent(extent<SizeType_, N_> ext);
 
   template <typename Rhs>
+  friend __host__ __device__ constexpr value_type operator-(window_extent const& lhs,
+                                                            Rhs rhs) noexcept
+  {
+    return lhs.value() - rhs;
+  }
+
+  template <typename Rhs>
   friend __host__ __device__ constexpr value_type operator/(window_extent const& lhs,
                                                             Rhs rhs) noexcept
   {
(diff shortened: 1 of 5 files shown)


sleeepyjack:
Found the culprit with `double hashing`:
Apparently [this](https://github.com/NVIDIA/cuCollections/blob/057d3e05cf16d7df59e346b05246754bf0729802/include/cuco/detail/probing_scheme_impl.inl#L152-L154) expression in the scalar code path is different from [this](https://github.com/NVIDIA/cuCollections/blob/057d3e05cf16d7df59e346b05246754bf0729802/include/cuco/detail/probing_scheme_impl.inl#L168-L171) one in the CG code path. I'd have to wrap my head around that logic again to fix this. Alternatively we could internally dispatch to the CG code path in the CG probing iterator when `CGSize==1` and call it a day.

PointKernel:
With #501 fixing the sanitize hash logic, removing `max` from the scalar probing should fix the issue.

sleeepyjack:
that fix broke something else: https://github.com/NVIDIA/cuCollections/actions/runs/9584556889/job/26440581801?pr=509#step:7:758

PointKernel:
> that fix broke something else: https://github.com/NVIDIA/cuCollections/actions/runs/9584556889/job/26440581801?pr=509#step:7:758

Which fix are you referring to? removing `max` from double hashing?

sleeepyjack:
> Which fix are you referring to? removing max from double hashing?

Operator presedence: [f0ae693](https://github.com/NVIDIA/cuCollections/pull/509/commits/f0ae693323c1ea62cbfa11e4c4fa7f719f999c19)

Not sure where my logic is flawed. 😢

PointKernel:
> > Which fix are you referring to? removing max from double hashing?
> 
> 
> 
> Operator presedence: [f0ae693](https://github.com/NVIDIA/cuCollections/pull/509/commits/f0ae693323c1ea62cbfa11e4c4fa7f719f999c19)
> 
> 
> 
> Not sure where my logic is flawed. 😢 

No good way but printing out all probing steps. I will take a look tomorrow.

PointKernel (approved):
Should be ready to go.

sleeepyjack:
You sir are the MVP

PR #511, opened 2024-06-19
Fix inconsistent use of noexcept

### LINKED ISSUE #510: [BUG]: Many cuco functions incorrectly marked as `noexcept`
### Is this a duplicate?

### Type of Bug

Runtime Error

### Describe the bug

Many cuco functions, both ctors and methods, are incorrectly marked as `noexcept`, despite them making calls into the cuda runtime that might potentially fail.

Some (non-exhaustive) examples:

- `aow_storage::initialize`, calls `cuda_stream_ref::synchronize` which might throw
- `aow_storage` ctor, allocates memory which might throw
- `open_addressing_impl` ctor, allocates memory which might throw
- `open_addressing_impl::size` calls a kernel and syncs the stream which might throw
- `open_addressing_impl::initialize` calls `aow_storage::initialize` and therefore transitively might throw
- `distinct_count_estimator::add_async` is marked as `noexcept`, but the impl class's method (`hyperloglog::add_async`) is _not_ `noexcept`, and indeed might throw since it launches kernels and calls `CUCO_CUDA_TRY(cudaLaunchKernel(...))`.
- `static_map::clear` is marked noexcept, but `open_addressing_impl::clear` calls `aow_storage::initialize` and therefore transitively might throw

### How to Reproduce

Try and construct a static set requiring more storage than there is device memory (or perhaps using RMM with a limited memory resource). RMM will throw a subclass of `std::bad_alloc` but we will not be able to catch it with:
```c++
try {
   make_cuco_static_set(...);
} catch (...) {
   // expect to get here
}
```
and instead the program will abort.

### Expected behavior

It should be possible to catch exceptions thrown inside cuco functions.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

sleeepyjack:
Closes #510

### DIFF
--- a/include/cuco/aow_storage.cuh
+++ b/include/cuco/aow_storage.cuh
@@ -81,7 +81,7 @@ class aow_storage : public detail::aow_storage_base<T, WindowSize, Extent> {
    * @param size Number of windows to (de)allocate
    * @param allocator Allocator used for (de)allocating device storage
    */
-  explicit constexpr aow_storage(Extent size, Allocator const& allocator = {}) noexcept;
+  explicit constexpr aow_storage(Extent size, Allocator const& allocator = {});
 
   aow_storage(aow_storage&&) = default;  ///< Move constructor
   /**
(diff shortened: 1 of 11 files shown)


PointKernel:
Are you using `clang-tidy` or just manually going through all the functions?

sleeepyjack:
Manually stepping through the code, checking every function if it can be noexcept or not. Tedious work.

PointKernel (approved):
Thanks for the quick fix!

PR #514, opened 2024-06-24
Add identity hasher

### LINKED ISSUE #487: [FEATURE]: Add `perfect_hashing` probing scheme

[Perfect hash functions](https://en.wikipedia.org/wiki/Perfect_hash_function) describe an injective mapping from the input key domain into the hash map's slot index domain. In other words, Each distinct key hashes to a distinct slot in the map.

This setup allows for a set of optimizations:
- Perfect hash functions don't require any probing logic, since there can't be any colliding key in a slot that is not equal to the probe key
- The injectivity constraint guarantees that any index value produced by the hash function is smaller than the map's capacity. Thus, we can get rid of the remainder computation.
- Key equality comparison can be simplified, i.e., we only have to check if the key in the slot is equal to a sentinel or not.

Add a new class `cuco::perfect_hashing<class Hash>` to our probing scheme zoo which behaves as follows:

When the dereferencing operator of the probing iterator is called for the first time (at the initial probing position), `return slots + hash(key)`. After incrementing the iterator, always `return end()`, meaning that there is at most one probing step.

A user must ensure that the `Hash` function in combination with the input key set actually forms a perfect hash function, and the maximum hash values is smaller than the map's capacity. Otherwise behavior is undefined.

Notes on the implementation:
- Currently each of our probing schemes uses the same `probing_iterator` class. This new probing scheme doesn't fit into the logic of the existing iterator. Thus I propose to let each probing scheme define its own `probing_iterator` as a member class.
- Since perfect hashing only requires bitwise comparison against the sentinel, we ignore any user-specified `KeyEqual` operator.

There is one more optimization we could additionally apply, but I would vote against it due to technical reasons:

Perfect hashing guarantees that there are no collisions. Thus, we could `insert` keys using non-atomic `STG` instructions, which has proven to be significantly faster compared to atomic CAS operations.
This however leads to some undesireful side effects due to the relaxed memory ordering of the GPU, which ultimately leads to implausible return values from some of our APIs (`insert_and_find` and also bulk `insert`; see example in the bottom paragraph of https://github.com/NVIDIA/cuCollections/issues/475#issuecomment-2113437463).

If this optimization is desired, it can still be enabled by specifying `cuda::thread_scope_thread` when instantiating the map type. This is a bit hacky but I think it's better than breaking the existing logic, introducing spurious errors in the aforementioned return values.

### Additional context

See discussion #475
```[tasklist]
### Tasks
```

njones93531:
Contribute to #487 
Partially fixes #505 

- Added cuco/detail/hash_functions/identityhash.cuh
  - Identity hasher uses thrust::identity to apply the identity function to its key
- Added IdentityHash to hash_functions.cuh
- Added unit tests to validate that the identity hash function works properly. Tests passed successfully.


include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (c) 2023-2024, NVIDIA CORPORATION.
```
PointKernel:
```suggestion
 * Copyright (c) 2024, NVIDIA CORPORATION.
```
Let's use the correct copyright year.

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,56 @@
...
+ * @brief An Identity hash function to hash the given argument on host and device
+.*
+ * @note IdentityHash is perfect iff hash_table_capacity >= |input set|
+ *
+ * @note IdentityHash is only intended to be used perfectly.
+ *
+ * @note Perfect hashes are deterministic, and thus do not need seeds.
+ *
+ * -----------------------------------------------------------------------------
+ *
```
PointKernel:
```suggestion
```
nit: this won't render properly with doxyen so let's just remove it.

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,56 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cuco/detail/hash_functions/utils.cuh>
```
PointKernel:
```suggestion
#include <thrust/functional.h>
```
suggestion: include what's used only. FYI, cuco follows cudf development guidelines in most cases https://github.com/rapidsai/cudf/blob/branch-24.08/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#includes

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,56 @@
...
+   * @return The resulting hash value for `key`
+   */
+  constexpr result_type __host__ __device__ operator()(Key const& key) const noexcept
+  {
+    return idfunct_(key);
+  }
+
+ private:
+  thrust::identity<Key> idfunct_;
+};  // Identity hash
```
PointKernel:
```cuda
template<typename Key>
struct identity_hash : public thrust::identity<Key> {
  using argument_type = Key;  ///< The type of the values taken as argument
  using result_type   = Key;  ///< The type of the hash values produced
};
```
This probably also works (and simplifies the code a bit).


Murmurhash and XXhash use PascalCase since they are ported from existing libraries. Identity hashers don't have such copyright issues thus we could use [snake_case](https://github.com/rapidsai/cudf/blob/branch-24.08/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#code-and-documentation-style-and-formatting) directly for class naming.

tests/utility/hash_test.cu
```diff
@@ -48,6 +48,63 @@ __host__ __device__ bool check_hash_result(typename Hash::argument_type const& k
...
+
+template <typename OutputIter>
+__global__ void check_identity_hash_result_kernel(OutputIter result)
+{
+  int i = 0;
+
+  result[i++] = check_hash_result<cuco::identityhash<signed char>>(SCHAR_MIN, SCHAR_MIN);
+  result[i++] = check_hash_result<cuco::identityhash<signed char>>(SCHAR_MAX, SCHAR_MAX);
+
+  result[i++] = check_hash_result<cuco::identityhash<int32_t>>(INT32_MIN, INT32_MIN);
```
PointKernel: Let's use [std::numeric_limits](https://en.cppreference.com/w/cpp/types/numeric_limits) instead of C macros. In modern C++ code, we want to get rid of any C expressions when possible, e.g. using STL algorithms instead of raw for/while loops, and here `std::numeric_limits<int32_t>::min()` instead of INT32_MIN.

include/cuco/hash_functions.cuh
```diff
@@ -16,11 +16,20 @@
...
 
 namespace cuco {
 
+/**
+ * @brief An `Identity` hash function to hash the given argument on host and device.
+ *
+ * @tparam Key The type of the values to hash
+ */
+template <typename Key>
+using identityhash = detail::IdentityHash<Key>;
```
PointKernel:
```suggestion
using identity_hash = detail::IdentityHash<Key>;
```
snake_case naming

tests/utility/hash_test.cu
```diff
@@ -48,6 +48,63 @@ __host__ __device__ bool check_hash_result(typename Hash::argument_type const& k
...
+                                                          (__int128)INT64_MAX + 1));
+#endif
+  }
+  SECTION("Check if device-generated hash values match the identity function.")
+  {
+    thrust::device_vector<bool> result(7, true);
+
+    check_identity_hash_result_kernel<<<1, 1>>>(result.begin());
+
+    CHECK(cuco::test::all_of(result.begin(), result.end(), [] __device__(bool v) { return v; }));
```
PointKernel, line 100:
```suggestion
    CHECK(cuco::test::all_of(result.begin(), result.end(), thrust::identity<bool>{}));
```
nit: a perfect use case for identity.

PointKernel (changes requested):

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,56 @@
...
+#pragma once
+
+#include <cuco/detail/hash_functions/utils.cuh>
+
+namespace cuco::detail {
+
+/**
+ * @brief An Identity hash function to hash the given argument on host and device
+.*
+ * @note IdentityHash is perfect iff hash_table_capacity >= |input set|
```
sleeepyjack:
```suggestion
 * @note `identity_hash` is perfect iff `hash_table_capacity >= |input set|`
```

nit: Doxygen can render inline code which makes it a bit easier to parse.

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @note IdentityHash is only intended to be used perfectly.
```
sleeepyjack:
```suggestion
 * @note `identity_hash` is only intended to be used perfectly.
```

include/cuco/detail/hash_functions/identityhash.cuh
sleeepyjack: I agree, an alias typedef is the right way to go here since Thrust already provides exactly what we need.

tests/utility/hash_test.cu
sleeepyjack: Playing the devil's advocate for a moment: When switching to an alias typedef as proposed earlier in this PR discussion, do we even need a unit test for it?

I don't want to be the guy who proposes having fewer unit tests in our project (hell no!), but this seems like a test that just checks if `thrust::identity` works as advertised, which should already be tested upstream, i.e., in Thrust itself.

include/cuco/detail/hash_functions/identityhash.cuh
PointKernel: I vaguely recall @esoha-nvidia has a reason to not use the alias directly

include/cuco/detail/hash_functions/identityhash.cuh
njones93531: Agreed!

include/cuco/detail/hash_functions/identityhash.cuh
njones93531: Makes sense to me

include/cuco/detail/hash_functions/identityhash.cuh
njones93531: Good catch! Inline code is certainly easier to parse

include/cuco/detail/hash_functions/identityhash.cuh
njones93531: Sounds good

include/cuco/detail/hash_functions/identityhash.cuh
njones93531: The issue comes from a future addition-- we likely want additional methods in all perfect hashers to help protect users from improper usage. @esoha-nvidia has a good explanation [here](https://github.com/NVIDIA/cuCollections/pull/505#discussion_r1646605810): 

> What we'll do is put a method in the perfect hashers (such as `SigbitHash` and `IdentityHash`) called `is_perfect` that returns `true`

I wrote the `IdentityHash` class out with these future additions in mind, but you are correct that in the present state, this file could be 5 lines long. I'm happy to make this change if it is still desired.

njones93531: Happy to use snake case!

include/cuco/hash_functions.cuh
njones93531: Happy to use snake case!

tests/utility/hash_test.cu
njones93531: I agree that `thrust::identity` works perfectly here, however `[] __device__(bool v) { return v; }` is used by the existing tests in this file for `xxhash` and `mumurhash`. See lines 168 and 343. Should we leave this as-is to maintain consistency? Changing the tests for `xxhash` and `murmurhash` to `thrust::identity` seems out of scope.

As for @sleeepyjack 's comment; even if we change the current code to an alias typedef, in the future I don't think `IdentityHash` will remain that way. Does it make sense to keep the unit tests to protect the functionality of `IdentityHash` against future changes?

tests/utility/hash_test.cu
njones93531: Sounds good to me!

tests/utility/hash_test.cu
PointKernel: It's fine. Another PR moving all device lambda to identity is more than welcome

include/cuco/detail/hash_functions/identityhash.cuh
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include <thrust/functional.h>
+
+namespace cuco::detail {
+
+/**
+ * @brief An Identity hash function to hash the given argument on host and device
+ *
+ * @note `identity_hash` is perfect iff `hash_table_capacity >= |input set|`
```
PointKernel:
```suggestion
 * @note `identity_hash` is perfect if `hash_table_capacity >= |input set|`
```

include/cuco/detail/hash_functions/identityhash.cuh
PointKernel, line 1: Let's rename the file to `identity_hash.cuh` accordingly

include/cuco/detail/hash_functions/identityhash.cuh
PointKernel: > we likely want additional methods in all perfect hashers to help protect users from improper usage.

Such methods should be at the probing scheme level instead of the hasher level. On the other hand, some cuco details may depend on `argument_type`/`result_type` thus using alias directly could break.


Let's move forward with the inheritance way:
```cuda
template<typename Key>
struct identity_hash : public thrust::identity<Key> {
  using argument_type = Key;  ///< The type of the values taken as argument
  using result_type   = Key;  ///< The type of the hash values produced
};
```

PointKernel (changes requested):
Some final suggestions otherwise look good to me.

PointKernel (approved):
LGTM :fire: 

Thanks! @njones93531

PointKernel:
@njones93531 Just FYI, I've updated the PR title since it will become the commit message with squash merge. 

Using the github keyword `resolves` will close the #487 issue when merging this PR thus I've changed it as well.

include/cuco/detail/hash_functions/identity_hash.cuh
```diff
@@ -0,0 +1,40 @@
...
+ * @note `identity_hash` is only intended to be used perfectly.
+ *
+ * @note Perfect hashes are deterministic, and thus do not need seeds.
+ *
+ * @tparam Key The type of the values to hash
+ */
+template <typename Key>
+struct identity_hash : public thrust::identity<Key> {
+  using argument_type = Key;  ///< The type of the values taken as argument
+  using result_type   = Key;  ///< The type of the hash values produced
```
esoha-nvidia: It seems that `thrust:identity` already defines `argument_type` and `result_type`: https://nvidia.github.io/cccl/thrust/api/structthrust_1_1identity.html#_CPPv4N6thrust8identity13argument_typeE

Because you are doing `public` inheritance above, I think that you could delete these two lines and it would still work, right?  In which case, you could also delete the whole file for now.  (@PointKernel also argues for deleting that file and I explain pros/cons below.)

tests/utility/hash_test.cu
```diff
@@ -48,6 +48,77 @@ __host__ __device__ bool check_hash_result(typename Hash::argument_type const& k
...
 
+// Overload for hash functions without a seed
+template <typename Hash>
+__host__ __device__ bool check_hash_result(typename Hash::argument_type const& key,
+                                           typename Hash::result_type expected) noexcept
+{
+  Hash h;
+  return (h(key) == expected);
+}
+
```
esoha-nvidia, line 59: I see why you made this: Because we have hash functions that don't take a seed.  However, we'll eventually have the `SigBitHash` that takes not a seed but rather a list or a range or whatever.

I think that it would be better to not write a million of these functions.  The right way to do it is with a variadic template.  Also, you could make that a different PR.  The code would look like this:

```cuda
template <typename Hash, typename... HashConstructorArgs>
static __host__ __device__ bool check_hash_result(typename Hash::argument_type const& key,
                                                  typename Hash::result_type expected,
                                                  HashConstructorArgs&&... hash_constructor_args) noexcept
{
  Hash h(std::forward<HashConstructorArgs>(hash_constructor_args)...);
  return (h(key) == expected);
}
```

The `seed` is now the last argument so you need to find all occurrences of `check_hash_result` in this file and make the seed the last argument.

After doing this change, the function will now be compatible with all hash construction.

---

If you want to read more about `std::forward`, these are good:

* https://stackoverflow.com/questions/16447951/how-to-forward-variable-number-of-arguments-to-another-function
* https://stackoverflow.com/questions/7257144/when-to-use-stdforward-to-forward-arguments

The short of it is that `std::forward` written this way with a "universal reference" makes it so that this code works well when:
* you pass in something that has a name, like `check_hash_result(my_key, my_expected, my_seed)`. (`my_seed` has a "name", it's a left-value aka l-value aka something that can be on the left side of an assignment)
* you pass in something nameless, like check_hash_result(my_key, my_expected, `12345)`. (`12345` has no "name", it's a right-value aka r-value aka something that can *not* be on the left side of an assignment statement)

When used like this, the compiler will correctly optimize if the value of the `seed` or whatever is passed in there needs to be copied or moved.

tests/utility/hash_test.cu
```diff
@@ -48,6 +48,77 @@ __host__ __device__ bool check_hash_result(typename Hash::argument_type const& k
...
+  result[i++] = check_hash_result<cuco::identity_hash<int64_t>>(
+    std::numeric_limits<int64_t>::min(), std::numeric_limits<int64_t>::min());
+  result[i++] = check_hash_result<cuco::identity_hash<int64_t>>(
+    std::numeric_limits<int64_t>::max(), std::numeric_limits<int64_t>::max());
+
+#if defined(CUCO_HAS_INT128)
+  result[i++] = check_hash_result<cuco::identity_hash<__int128>>(
+    (__int128)std::numeric_limits<int64_t>::max() + 1,
+    (__int128)std::numeric_limits<int64_t>::max() + 1);
```
esoha-nvidia: I assume that there is no `numeric_limits` for `__int128` and that's why you did it like this?

I'm cool with that.

tests/utility/hash_test.cu
esoha-nvidia: To expand on the second half of [what @njones93531 said](https://github.com/NVIDIA/cuCollections/pull/514/files#r1652738321), as I described [here](https://github.com/NVIDIA/cuCollections/pull/505#discussion_r1646605810), it will greatly improve the safety and usability of cuco perfect hashing if we design the "perfect probing" scheme to require that the hash function used with it is perfect.  That is, we want a compilation failure if the user tried to use a non-perfect hash with perfect probing because it will definitely cause errors in the hash table's function.

The way that I propose to get that safety is like [how type traits work](https://www.internalpointers.com/post/quick-primer-type-traits-modern-cpp).  We'll define a method `bool is_perfect()` of hash functors that are perfect.  And the perfect probing scheme will fail at compile-time if you try to use it with a hash functor that doesn't have that function or the function returns `false`.

In order to do this, you'll have to have an `identity_hash` like above, that extends the `thrust::identity` and adds the `is_perfect` function into it.

So, while we *don't* need the file `identity_hash.cuh` at the moment, we *will* need it later.  We can add it now, like this PR does, or remove this and add it later along with the perfect hashing.

@njones93531 Let's not [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality) this one!  Let @PointKernel read this argument, decide if he wants the class now or later, and we'll just do it.  :smile:

esoha-nvidia (changes requested):
Thanks for writing this PR!  I like that it is small and easy to reason about!

sleeepyjack:
The tag type approach has been abandoned some time ago in many libraries and replaced with specialized type traits, e.g., specializing `std::iterator_traits<my_iterator_type>`.
Based on offline discussions (involving all of us IIRC) we settled on a different design to make using the `perfect_hashing` probing scheme safe:

- Make the ctor of the `perfect_hashing` class `private` and only allow it to be constructed via a `make_perfect_hashing` factory function
- The factory returns a `cuda::std::optional<perfect_hashing<my_hasher>>` object which will only hold an actual probing scheme object in case the factory could verify that it's actually perfect and thus safe to use

In case construction fails, the user has to provide an alternative code path, i.e., a different non-perfect probing scheme.

One thing is unclear to me: @esoha-nvidia you mentioned we want the `is_perfect` check to be available at compile time but in my understanding there can be scenarios where this information is only known after introspecting the data, i.e., at runtime.

include/cuco/detail/hash_functions/identity_hash.cuh
sleeepyjack: One thing that just popped up in my head is that we assume a hash value to be of unsigned type for all of the other hashers. In our probing internals we need to perform some arithmetic combining (potentially signed) size types with unsigned hash value types. The logic as implemented has become quite complex, ensuring no signed integer overflow occurs, etc.

I'm only mildly concerned this is actually a problem since the unit test passes just fine, yet I wanted to bring this up just to make sure.

One other thing is that since we're using the hash value to do some integer arithmetics, we need to ensure the result type is actually an integer. Maybe just `*reinterpret_cast<uint32_t* or uint64_t*>` the result of `thrust::identity` and call it a day?

njones93531: I think this points to a need for more user protection in `identity_hash` (and perfect hashing) in general. I believe we've discussed how to ensure `perfect_probing` always uses a perfect hash, but have we discussed whether perfect hashes can be used with other probing schemes? 
Here are my thoughts: 

# Option 1: `identity_hash` is ONLY intended for `perfect_probing`
It seems to me that `perfect_probing` would not require any arithmetic on the hash value for the following reasons:
- **No collisions occur.**
A hash value should indicate exactly one slot, and no others. No need to increment or otherwise alter the hash value to find a valid slot. 
- **Cooperative group logic is omitted.** 
Because we have no collisions, we shouldn't need cooperative groups either. We won't need to alter the hash value to enable cooperative probing. 
- **We already require `hash_table_size > max_possible_hash_value`, and `min_possible_hash_value >= 0`.** 
If this condition is not met, the hashing is not perfect. Presumably, we will have protections in place to ensure this requirement holds. 

If I'm correct about avoiding hash value arithmetic with `perfect_probing`, the only remaining way I can see `identity_hash` breaking in this mode is if the key type somehow cannot be converted into a valid slot. However, I would assume this means integer comparison won't work either, so it should fail the range test described above. 
In this mode, we would likely want protections to prevent users from using `identity_hash` with other probing schemes. 

# Option 2: We want `identity_hash` to work for non-perfect probing schemes
This mode seems closest to what a new user would assume, which may be a strong argument in its favor. In this mode, the points @sleeepyjack mentioned are extra important-- any output that isn't an integer will break the probing logic. But there are other problems as well: 
- **Double Hashing and perfect hashing don't cooperate** 
Any perfect hash that is supplied as both hashes to a `double_hashing` probing scheme will probably break (or at least perform terribly) on the first collision.  Because the prober should always return the same occupied slot, `insert` of the first collided value will either infinite loop or fail, depending on how that logic is implemented. Neither of these outcomes is acceptable-- therefore we likely need some protections to prevent users from using `double_hashing` with perfect hashers. 
- **Other probing can still be bad** 
Without the `is_perfect` guarantee, there are *zero* guarantees that `identity_hash` distributes values well. Note that every `key` that is a multiple of the `hash_table_size` hashes to zero due to the modulo operator. I imagine that most perfect hashes little-to-no performance guarantees for imperfect use.
- **Casting an `identity_hash` hash value could break perfect hashing** 
Certainly, if we allow types to pass through `identity_hash` that could break the prober, that's unacceptable. However, any of these problematic values would also break the `is_perfect` property if cast into to an acceptable range, because we now have two different keys that result in the same hash value. If go this route, I propose we limit the allowed `Key` types, or only cast if the probing scheme is not perfect. 


> I'm only mildly concerned this is actually a problem since the unit test passes just fine, yet I wanted to bring this up just to make sure.

I don't think my unit tests would catch this issue, as they only test the hasher and not any interaction with the probers. At the moment, the only protection in place is a comment stating `identity_hash` is only intended for use with `perfect probing`  🙂

tests/utility/hash_test.cu
njones93531: Actually, this is leftover logic-- there is no C macro for `__int128`, which was my first implementation. I believe any implementation with `__int128` is also required to implement `numeric_limits` for it. Good catch!

include/cuco/detail/hash_functions/identity_hash.cuh
njones93531: Good catch!

esoha-nvidia: @sleeepyjack When I saw the modulo operator in the probing scheme without any `static_assert` that the output of the hash function is unsigned and integral, I knew there was a potential for breakage.

Just put two `static_assert` right next to that modulo operator and it should take care of itself, right?  You might also want to add unit tests.  Is it possible to make a unit test that passes when compilation fails?  You could test that using an identity hash that returns non-integral or signed values causes a compilation failure.

esoha-nvidia:
>     * Make the ctor of the `perfect_hashing` class `private` and only allow it to be constructed via a `make_perfect_hashing` factory function
>     * The factory returns a `cuda::std::optional<perfect_hashing<my_hasher>>` object which will only hold an actual probing scheme object in case the factory could verify that it's actually perfect and thus safe to use

You can do it this way if you want but it has the downside that you can no longer default construct the `perfect_hashing`.  Though you weren't really able to in the first place.

If you make a private constructor that does all the work in the constructor then, as far as I can tell, your factory function is just:

```c++
std::optional<perfect_hashing> make_perfect_hashing(args) {
  try {
    return {perfect_hashing(args)};
  } catch {
    return std::nullopt;
  }
}
```

Right?  It's kind of a trivial factory fuction, just converting an exception to an optional.  Is it pointless?

Or, you could make a constructor that never throws and then put the failure logic into the factory.  The problem with this is that now you have split-brained code where the creation of the perfect hash is happening partly in the factory and party in the constructor.  If you get into a situation where you end up doing some math in the factory just to send it to the constructor and then having to examine that constructor to determine if a valid hash functor was created...

I guess that we'll see how it shakes out.  Just be prepared for maybe having to put everything into the constructor because you want to avoid split-brained code.  And then be prepare to ask yourself, "Why do we even have this factory?"

> One thing is unclear to me: @esoha-nvidia you mentioned we want the `is_perfect` check to be available at compile time but in my understanding there can be scenarios where this information is only known after introspecting the data, i.e., at runtime.

Hmm, let me think about this... Yes, you're right.  There will be some hashers for which `is_perfect` could be `constexpr` but some for which it cannot be.  For example, if the SigBits hasher computes more significant bits than there are bits in the result type then it will end up creating a hasher that is not perfect (though it might still be a pretty good hasher!).

Okay, so `is_perfect` is not always constexpr and, likewise, the check for `is_perfect` cannot be `if constexpr`.  (That's okay, we didn't really need the `if constexpr` because `if constexpr` doesn't actually make code run faster, it just makes code that previously did *not* compile now compile.)

tests/utility/hash_test.cu
njones93531: I like this solution, and the idea of doing it in a different PR since it affects so many non-identity_hash tests. Perhaps it makes sense to knock this and #518 out in the same PR

njones93531: Opened issue #521 to cover this

include/cuco/detail/hash_functions/identity_hash.cuh
njones93531: After looking at the other hashers, it makes sense why `probing_scheme.impl` can assume an unsigned integral type: the existing hashers can only return `uint32_t` or `uint64_t`. If this responsibility belongs to the existing hashers, it makes sense for `identity_hash` to guarantee its output types as well. 

I was wrong to be worried about disrupting the `is_perfect` property due to signed types, since the bit pattern should remain the same. However, I think key types larger than `uint64_t` will still break the `is_perfect` property. Since `identity_hash` is more of an example perfect hash anyway, I think it makes sense to limit its input to sizes smaller than `uint64_t`.

include/cuco/detail/hash_functions/identity_hash.cuh
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @tparam Key The type of the values to hash. Must have size <= uint64_t
+ */
+template <typename Key>
+struct identity_hash : private thrust::identity<Key> {
+  using argument_type = Key;       // The type of the values taken as argument
+  using result_type   = uint64_t;  // The type of the hash values produced
+
+  static_assert(sizeof(Key) <= sizeof(result_type),
+                "Key type must not be larger than result_type in identity_hash");
```
PointKernel:
```suggestion
  static_assert(cuda::std::is_convertible_v<Key, result_type>,
                "Key type must be convertible to result_type");
```
cuco supports heterogenous lookup without requiring C++20 thus the convertible trait probably fits better. In case you are interested, heterogenous lookup is a powerful add to the hash table: https://www.cppstories.com/2021/heterogeneous-access-cpp20/

include/cuco/detail/hash_functions/identity_hash.cuh
```diff
@@ -0,0 +1,48 @@
...
+
+  __host__ __device__ result_type operator()(const Key& x) const
```
PointKernel:
```suggestion
  __host__ __device__ result_type operator()(Key const& x) const
```
nit: cuco and RAPIDS projects prefer east const since there is only single rule to follow. 
For your reference:

- https://hackingcpp.com/cpp/design/east_vs_west_const.html
- https://github.com/rapidsai/cudf/blob/branch-24.08/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#c-guidelines

PointKernel (changes requested):
Looks great.

Two final suggestions otherwise the PR is ready to go.

As we discussed offline, discussions and evaluation of the perfect hashing design will be continued in a private repo until it matures.

include/cuco/detail/hash_functions/identity_hash.cuh
```diff
@@ -0,0 +1,48 @@
...
+ * @note `identity_hash` is only intended to be used perfectly.
+ *
+ * @note Perfect hashes are deterministic, and thus do not need seeds.
+ *
+ * @tparam Key The type of the values to hash. Must have size <= uint64_t
+ */
+template <typename Key>
+struct identity_hash : private thrust::identity<Key> {
+  using argument_type = Key;       // The type of the values taken as argument
+  using result_type   = uint64_t;  // The type of the hash values produced
```
sleeepyjack:
```suggestion
  using result_type   = std::conditional_t<sizeof(Key) <= 4, uint32_t, uint64_t>;  // The type of the hash values produced
```

Not sure if this is applicable in all cases, but I think if the input key type is already a 32bit int then there's no need in returning a 64bit hash value.

njones93531: Agreed!

include/cuco/detail/hash_functions/identity_hash.cuh
njones93531: SGTM, thanks for the reference!

PointKernel (approved):
I updated the docs accordingly to unblock CI and all looks good.

PR #523, opened 2024-06-27
[Review] Refactor hasher unit tests

### LINKED ISSUE #518: Let's move all device lambda to use thrust::identity
As discussed in https://github.com/NVIDIA/cuCollections/pull/514#discussion_r1653116516_

we find some obvious lambdas that could be using `thrust::identity` instead.  This should be an easy fix.

### LINKED ISSUE #521: Variadic Template to allow new hash functions in unit tests
I see why you made this: Because we have hash functions that don't take a seed.  However, we'll eventually have the `SigBitHash` that takes not a seed but rather a list or a range or whatever.

I think that it would be better to not write a million of these functions.  The right way to do it is with a variadic template.  Also, you could make that a different PR.  The code would look like this:

```cuda
template <typename Hash, typename... HashConstructorArgs>
static __host__ __device__ bool check_hash_result(typename Hash::argument_type const& key,
                                                  typename Hash::result_type expected,
                                                  HashConstructorArgs&&... hash_constructor_args) noexcept
{
  Hash h(std::forward<HashConstructorArgs>(hash_constructor_args)...);
  return (h(key) == expected);
}
```

The `seed` is now the last argument so you need to find all occurrences of `check_hash_result` in this file and make the seed the last argument.

After doing this change, the function will now be compatible with all hash construction.

---

If you want to read more about `std::forward`, these are good:

* https://stackoverflow.com/questions/16447951/how-to-forward-variable-number-of-arguments-to-another-function
* https://stackoverflow.com/questions/7257144/when-to-use-stdforward-to-forward-arguments

The short of it is that `std::forward` written this way with a "universal reference" makes it so that this code works well when:
* you pass in something that has a name, like `check_hash_result(my_key, my_expected, my_seed)`. (`my_seed` has a "name", it's a left-value aka l-value aka something that can be on the left side of an assignment)
* you pass in something nameless, like check_hash_result(my_key, my_expected, `12345)`. (`12345` has no "name", it's a right-value aka r-value aka something that can *not* be on the left side of an assignment statement)

When used like this, the compiler will correctly optimize if the value of the `seed` or whatever is passed in there needs to be copied or moved.

_Originally posted by @esoha-nvidia in https://github.com/NVIDIA/cuCollections/pull/514#discussion_r1653534402_

njones93531:
Based on unit tests for `identity_hash` implemented in #514, we saw that a refactor of `tests/utility/hash_test.cu` could enable much cleaner implementation of future hashers with different construction arguments. 

CHANGELIST: 
1. Added  variadic template to `check_hash_function` to account for different hasher constructor signatures. 
- Requires existing calls to `check_hash_function` to make `seed` the final argument
- resolves #521
2. Replaced device lambda with `thrust::identity`. 
- resolves #518


sleeepyjack:
Would it be possible to base this PR on the current `dev` branch? It's a small change that shouldn't break if not being on top of your identity hasher work. This way we can get it merged quicker.

tests/utility/hash_test.cu
```diff
@@ -39,12 +39,13 @@ struct large_key {
...
-                                           Seed seed,
-                                           typename Hash::result_type expected) noexcept
+template <typename Hash, typename... HashConstructorArgs>
+static __host__ __device__ bool check_hash_result(
+  typename Hash::argument_type const& key,
+  typename Hash::result_type expected,
+  HashConstructorArgs&&... hash_constructor_args) noexcept
 {
-  Hash h(seed);
+  Hash h(std::forward<HashConstructorArgs>(hash_constructor_args)...);
```
PointKernel:
```suggestion
  Hash h(cuda::std::forward<HashConstructorArgs>(hash_constructor_args)...);
```
With the support of CCCL, we probably want to use `cuda::std::forward` since this is a host-device function.

Rules of thumb for using STL utilities:

- `std::` in host function
- `cuda::` for device function
- `cuda::std::` for host-device function

njones93531: Thanks for the info!

PointKernel:
pre-commit.ci run

PointKernel (approved):
Looks great! Thanks @njones93531 !

sleeepyjack (approved):
LGTM too 🚀

Issue #477, opened 2024-05-07
[FEATURE]: Add host-bulk `insert_and_find`


PointKernel:
### Is your feature request related to a problem? Please describe.

Add host-bulk `insert_and_find` function for `static_set` and `static_map`

### Describe the solution you'd like

```cuda
/**
  * @brief Asynchronously inserts all elements in the range `[first, last)`.
  *
  * @note: For a given element `*(first + i)`, if the container doesn't already contain an element
  * with an equivalent key, inserts the element at a location pointed by `iter` and writes
  * `thrust::pair{iter, true}` to `output_begin + i`. Otherwise, finds the location of the
  * equivalent element, `iter` and writes `thrust::pair{iter, false}` to `output_begin + i`.
  * 
  * @tparam InputIt Device accessible random access input iterator
  * @tparam OutputIt Device accessible random access output iterator whose `value_type`
  * is constructible from `thrust::pair<map::iterator, bool>`
  *
  * @param first Beginning of the sequence of elements
  * @param last End of the sequence of elements
  * @param stream CUDA stream used for insert
  */
template<typename InputIt, typename OutputIt>
void insert_and_find_async (InputIt first, InputIt last, OutputIt output_begin, cuda_stream_ref stream = {});
```

### Describe alternatives you've considered

```cuda
/**
  * @brief Asynchronously inserts all elements in the range `[first, last)`.
  *
  * @note: For a given element `*(first + i)`, if the container doesn't already contain an element
  * with an equivalent key, inserts the element at a location pointed by `iter` and writes
  * `iter` to `location_begin + i` and writes `true` to `inserted_begin + i`. Otherwise, finds the location of the
  * equivalent element, `iter` and writes `iter` to `location_begin + i` and writes `false` to `inserted_begin + i`.
  * 
  * @tparam InputIt Device accessible random access input iterator
  * @tparam LocationIt Device accessible random access output iterator whose `value_type`
  * is constructible from `map::iterator` type
  * @tparam Boolt Device accessible random access output iterator whose `value_type`
  * is constructible from `bool`
  *
  * @param first Beginning of the sequence of elements
  * @param last End of the sequence of elements
  * @param stream CUDA stream used for insert
  */
template<typename InputIt, typename LocationIt, typename Boolt>
void insert_and_find_async (InputIt first, InputIt last, LocationIt location_begin, Boolt inserted_begin, cuda_stream_ref stream = {});
```
[...]

PR #534, opened 2024-07-09
Add host-bulk insert_and_find for static_map

srinivasyadav18:
This PR adds **host-bulk** `insert_and_find` API for `static_map`.

- [x] Updates insert_and_find unit tests.
- [x] Adds documentation. 

Contributes to #477

### DIFF
--- a/tests/static_map/insert_and_find_test.cu
+++ b/tests/static_map/insert_and_find_test.cu
@@ -23,57 +23,12 @@
 #include <thrust/device_vector.h>
 #include <thrust/execution_policy.h>
 #include <thrust/functional.h>
-#include <thrust/sequence.h>
+#include <thrust/iterator/counting_iterator.h>
+#include <thrust/iterator/transform_iterator.h>
 
 #include <catch2/catch_template_test_macros.hpp>
 
-static constexpr int Iters = 10'000;
-
-template <typename Ref>
-__global__ void parallel_sum(Ref v)
-{
-  for (int i = 0; i < Iters; i++) {
-#if __CUDA_ARCH__ < 700
-    if constexpr (cuco::detail::is_packable<Ref::value_type>())
-#endif
-    {
-      auto constexpr cg_size = Ref::cg_size;
-      if constexpr (cg_size == 1) {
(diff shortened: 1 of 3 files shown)


srinivasyadav18:
@PointKernel One of the test on **pre-volta arch** fails due to `static_assert`, as we don't support packed type (size > 8) writes on pre-volta architecture. I cannot disable those tests from hostside because `__CUDA_ARCH__` is not defined on hostside.  May I know how should be move further with this issue. Thanks!

PointKernel:
> @PointKernel One of the test on **pre-volta arch** fails due to `static_assert`, as we don't support packed type (size > 8) writes on pre-volta architecture. I cannot disable those tests from hostside because `__CUDA_ARCH__` is not defined on hostside. May I know how should be move further with this issue. Thanks!

voila https://github.com/NVIDIA/cuCollections/blob/63d313eb015ab756f3ee7665af3b4b5b274e17fc/tests/static_map/custom_type_test.cu#L104-L109

tests/static_map/insert_and_find_test.cu
```diff
@@ -23,57 +23,16 @@
 #include <thrust/device_vector.h>
 #include <thrust/execution_policy.h>
 #include <thrust/functional.h>
-#include <thrust/sequence.h>
+#include <thrust/iterator/counting_iterator.h>
+#include <thrust/iterator/transform_iterator.h>
+
+#include <cuda.h>
```
PointKernel: do we need this header?

tests/static_map/insert_and_find_test.cu
```diff
@@ -23,57 +23,16 @@
...
 
 #include <catch2/catch_template_test_macros.hpp>
 
-static constexpr int Iters = 10'000;
+#include <iostream>
```
PointKernel:
```suggestion
```
could be removed

tests/static_map/insert_and_find_test.cu
srinivasyadav18: cuda.h is not requried. I missed this during cleanup.

PointKernel (approved):

sleeepyjack (approved):
Nice work!

PR #535, opened 2024-07-09
Fix usage of `decltype` with `probing_scheme` in `.with_hash_function`

srinivasyadav18:
The current version of `with_hash_function` uses **decltype** directly on a **const** object which add's **const** qualifier to the template parameter. 
This PR adds missing `decay_t` when creating new device_ref of the corresponding containers.


PointKernel (approved):

sleeepyjack (approved):
🎯

PR #537, opened 2024-07-11
Implement OA `retrieve(_outer)` and its `multiset` API

### LINKED ISSUE #465: [FEATURE]: Add multiset host-bulk retrieve APIs

Add multiset host-bulk retrieve APIs 

The basic API to add:
```cuda
  /**
   * @brief Retrieves the matched key in the multiset corresponding to all probe keys in the range
   * `[first, last)`
   *
   * For key `k = *(first + i)` has matches in the multiset, copies `k` and each of its associated values to
   * unspecified locations in `output_probe + *` and `output_match + *`. Else, does nothing.
   *
   * `retrieve_async`.
   * @note Behavior is undefined if the size of the output range exceeds
   * `std::distance(output_*_begin, output_*_end)`.
   *
   * @tparam InputIt Device accessible input iterator
   * @tparam OutputIt1 Device accessible output iterator whose `value_type` can be constructed from
   * probe key type
   * @tparam OutputIt2 Device accessible output iterator whose `value_type` can be constructed from
   * `Key`
   *
   * @param first Beginning of the sequence of probe keys
   * @param last End of the sequence of probe keys
   * @param output_probe Beginning of the sequence of the probe keys that have a match
   * @param output_match Beginning of the sequence of the matched keys
   * @param stream CUDA stream used to retrieve
   *
   * @return Pair of the iterators indicating the last elements in the output
   */
  template <typename InputIt, typename OutputIt1, typename OutputIt2>
  std::pair<OutputIt1, OutputIt2> retrieve(InputIt first,
                                           InputIt last,
                                           OutputIt1 output_probe,
                                           OutputIt2 output_match,
                                           cuda_stream_ref stream = {}) const;
```
Two APIs desired by libcudf:
```cuda
  /**
   * @brief Asynchronously retrieves the matched key in the set corresponding to all probe keys in
   * the range `[first, last)`
   *
   * For key `k = *(first + i)` has matches in the multiset, copies `k` and each of its associated values to
   * unspecified locations in `output_probe + *` and `output_match + *`. Else, does nothing.
   *
   * @note Behavior is undefined if the size of the output range exceeds
   * `std::distance(output_*_begin, output_*_end)`.
   *
   * @tparam InputIt Device accessible input iterator
   * @tparam OutputIt1 Device accessible output iterator whose `value_type` can be constructed from
   * probe key type
   * @tparam OutputIt2 Device accessible output iterator whose `value_type` can be constructed from
   * `Key`
   * @tparam ProbeKeyEqual Binary callable equal type
   * @tparam ProbeHash Unary callable hasher type that can be constructed from
   * an integer value
   *
   * @param first Beginning of the sequence of probe keys
   * @param last End of the sequence of probe keys
   * @param probe_key_equal The binary function to compare set keys and probe keys for equality
   * @param probe_hash The unary function to hash probe keys
   * @param output_probe Beginning of the sequence of the probe keys that have a match
   * @param output_match Beginning of the sequence of the matched keys
   * @param stream CUDA stream used to retrieve
   *
   * @return Pair of the iterators indicating the last elements in the output
   */
  template <typename InputIt, typename OutputIt, typename ProbeKeyEqual, typename ProbeHash>
  std::pair<OutputIt1, OutputIt2>  retrieve(InputIt first,
                                            InputIt last,
                                            ProbeKeyEqual const& probe_key_equal,
                                            ProbeHash const& probe_hash,
                                            OutputIt1 output_probe,
                                            OutputIt2 output_match,
                                            cuda_stream_ref stream = {}) const;

  /**
   * @brief Asynchronously retrieves the matched key in the set corresponding to all probe keys in
   * the range `[first, last)`
   *
   * For key `k = *(first + i)` has matches in the multiset, copies `k` and each of its associated
   * values to unspecified locations in `output_probe + *` and `output_match + *`. Otherwise,
   * copies `key` and `empty_key_sentinel`.
   *
   * @note Behavior is undefined if the size of the output range exceeds
   * `std::distance(output_*_begin, output_*_end)`.
   *
   * @tparam InputIt Device accessible input iterator
   * @tparam OutputIt1 Device accessible output iterator whose `value_type` can be constructed from
   * probe key type
   * @tparam OutputIt2 Device accessible output iterator whose `value_type` can be constructed from
   * `Key`
   * @tparam ProbeKeyEqual Binary callable equal type
   * @tparam ProbeHash Unary callable hasher type that can be constructed from
   * an integer value
   *
   * @param first Beginning of the sequence of probe keys
   * @param last End of the sequence of probe keys
   * @param probe_key_equal The binary function to compare set keys and probe keys for equality
   * @param probe_hash The unary function to hash probe keys
   * @param output_probe Beginning of the sequence of the probe keys that have a match
   * @param output_match Beginning of the sequence of the matched keys
   * @param stream CUDA stream used to retrieve
   *
   * @return Pair of the iterators indicating the last elements in the output
   */
  template <typename InputIt, typename OutputIt, typename ProbeKeyEqual, typename ProbeHash>
  std::pair<OutputIt1, OutputIt2>  retrieve_outer(InputIt first,
                                                  InputIt last,
                                                  ProbeKeyEqual const& probe_key_equal,
                                                  ProbeHash const& probe_hash,
                                                  OutputIt1 output_probe,
                                                  OutputIt2 output_match,
                                                  cuda_stream_ref stream = {}) const;
```

- Can the `_outer` variant be represented in a more natural expression?

**Can we use cub::DeviceSelect in this case?**

### Additional context

- `retrieve` APIs are always synchronous
- The most challenging PR requires extensive profiling and benchmarking work: launching parameter (choosing optimal block size, shared memory sizes, etc), buffer size, probing/flushing CG sizes to tune

### LINKED ISSUE #489: [FEATURE]: Add `static_multiset::retrieve_outer`

`static_multiset` requires an equivalent function to `static_multimap::retrieve_outer`.

After #465 is finished, add a new API `retrieve_outer` and also add a template parameter to the kernel to switch between `retrieve` and `retrieve_outer` behavior.

Still an open discussion if this is the best approach to tackle this problem but we couldn't come up with a better solution so far.

### Additional context

_No response_

sleeepyjack:
Closes #465
Closes #489


include/cuco/detail/open_addressing/kernels.cuh
```diff
@@ -359,6 +359,49 @@ CUCO_KERNEL __launch_bounds__(BlockSize) void find(InputIt first,
...
+                                                       Ref ref)
+{
+  namespace cg = cooperative_groups;
+
+  auto const block           = cg::this_thread_block();
+  auto const blocks_in_grid  = gridDim.x;
+  auto const elems_per_block = cuco::detail::int_div_ceil(n, blocks_in_grid);
+
+  auto const block_begin_offset = block.group_index().x * elems_per_block;
+  auto const block_end_offset   = min(n, block_begin_offset + elems_per_block);
```
PointKernel: This is based on the assumption that one thread will be used per input. Therefore, the active block is doing more work than expected while the idle block is doing nothing but wait

tests/static_multiset/retrieve_test.cu
```diff
@@ -0,0 +1,189 @@
...
+  using probe = std::conditional_t<Probe == cuco::test::probe_sequence::linear_probing,
+                                   cuco::linear_probing<CGSize, cuco::default_hash_function<Key>>,
+                                   cuco::double_hashing<CGSize, cuco::default_hash_function<Key>>>;
+
+  auto set = cuco::static_multiset{
+    num_keys, desired_load_factor, cuco::empty_key<Key>{empty_key_sentinel}, {}, probe{}};
+
+  test_unique_sequence(set, num_keys);
+  // test_multiplicity(set, num_keys, decltype(set)::cg_size + 1); // TODO deadlock or infinite loop
+  // :/ test_outer(set, num_keys); // TODO also deadlocks -.-
```
sleeepyjack: Something is not right. Daang.

include/cuco/detail/open_addressing/kernels.cuh
sleeepyjack: Not sure I understand. The idea here is that we launch a grid with a fixed number of blocks. We partition the input range evenly among all blocks and pass the sub-ranges to each block. The actual assignment from input elements to CGs happens inside the `retrieve` device function itself by using a (CG) block stride loop over the sub-range.

sleeepyjack: I'm 90% sure I fixed it

sleeepyjack:
The `outer` test is still failing and the speeddown compared to the previous implementation is still 1.5x. Apart from that, the other unit tests look good. So the natural next steps would be to fix the bug in the `retrieve_outer` (shouldn't be a big deal) and dive into optimizations. For the latter I could use a second pair of eyes since this kernel is notoriously complex.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -907,6 +994,78 @@ class open_addressing_impl {
...
+    auto const n = detail::distance(first, last);
+    if (n == 0) { return {output_probe, output_match}; }
+
+    using counter_type = detail::counter_storage<size_type, thread_scope, allocator_type>;
+    auto counter       = counter_type{this->allocator()};
+    counter.reset(stream.get());
+
+    int32_t constexpr block_size = cuco::detail::default_block_size();
+    // int32_t grid_size =
+    //   detail::max_occupancy_grid_size(block_size,
```
sleeepyjack: The `max_occupancy` version is always a bit slower. I still don't understand why that's the case.

PointKernel: I noticed it's a better choice only for several cases of the legacy `static_multimap::pair_count`.

PointKernel:
> For the latter I could use a second pair of eyes since this kernel is notoriously complex.

Commenting out the code part by part to find the largest bottleneck is probably the most efficient way.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -996,6 +1081,78 @@ class open_addressing_impl {
...
+    //                                   detail::retrieve<IsOuter,
+    //                                                    block_size,
+    //                                                    InputProbeIt,
+    //                                                    OutputProbeIt,
+    //                                                    OutputMatchIt,
+    //                                                    typename counter_type::value_type,
+    //                                                    Ref>);
+    // grid_size *= 64;  // oversubscription factor
+    // // TODO shrink grid if n is very small
```
```suggestion
```

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -977,6 +978,325 @@ class open_addressing_ref_impl {
...
+    auto const flushing_tile_id = flushing_tile.meta_group_rank();
+    auto idx                    = probing_tile.meta_group_rank();
+    auto const stride           = probing_tile.meta_group_size();
+
+    // TODO align to 16B?
+    __shared__ probe_type probe_buffers[num_flushing_tiles][buffer_size];
+    __shared__ value_type match_buffers[num_flushing_tiles][buffer_size];
+    size_type num_matches = 0;
+
+    auto flush_buffers = [&](cg::coalesced_group const& tile) {
```
PointKernel, line 1151: I probably need to go over the CUDA doc again but good to know that we can pass any thread block tiles as a `coalesced_group`.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -977,6 +978,325 @@ class open_addressing_ref_impl {
...
+            //               });
+            // #else
+            //               if (probing_tile.thread_rank() == 0) {
+            //                 probe_buffers[flushing_tile_id][num_matches] = probe;
+            //                 probe_buffers[flushing_tile_id][num_matches] =
+            //                 this->empty_slot_sentinel();
+            //               }
+            // #endif
+            //               num_matches++;  // not really a match but a sentinel in the buffer
+            //             }
```
PointKernel:
```suggestion
```

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -977,6 +978,325 @@ class open_addressing_ref_impl {
...
+            }
+
+            if constexpr (IsOuter) {
+              if (not found_any_match /*yet*/ and probing_tile.any(match_found) /*now*/) {
+                found_any_match = true;
+              }
+            }
+
+            // reset flag for next iteration
+            match_found = false;
```
PointKernel, line 1232: question: reset for next key?

include/cuco/detail/static_multiset/static_multiset.inl
```diff
@@ -277,6 +277,48 @@ void static_multiset<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Sto
   impl_->find_async(first, last, output_begin, ref(op::find), stream);
 }
 
+// TODO docs
```
PointKernel:
```suggestion
```

include/cuco/detail/static_multiset/static_multiset.inl
```diff
@@ -277,6 +277,48 @@ void static_multiset<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Sto
...
+  InputProbeIt last,
+  OutputProbeIt output_probe,
+  OutputMatchIt output_match,
+  cuda::stream_ref stream) const
+{
+  return this->impl_->retrieve(
+    first, last, output_probe, output_match, this->ref(op::retrieve), stream);
+}
+
+// TODO docs
```
PointKernel:
```suggestion
```

include/cuco/detail/static_multiset/static_multiset_ref.inl
```diff
@@ -479,6 +479,123 @@ class operator_impl<
...
+
+  /**
+   * @brief Retrieves all the slots corresponding to all keys in the range `[input_probe_begin,
+   * input_probe_end)`.
+   *
+   * If key `k = *(first + i)` exists in the container, copies `k` to `output_probe` and associated
+   * slot contents to `output_match`, respectively. The output order is unspecified.
+   *
+   * Behavior is undefined if the size of the output range exceeds the number of retrieved slots.
+   * Use `count()` to determine the size of the output range.
```
PointKernel:
```suggestion
   * Use `count_outer()` to determine the size of the output range.
```

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -977,6 +978,325 @@ class open_addressing_ref_impl {
...
+            //                 this->empty_slot_sentinel();
+            //               }
+            // #endif
+            //               num_matches++;  // not really a match but a sentinel in the buffer
+            //             }
+          }
+
+          // if the buffer has not enough empty slots for the next iteration
+          if (num_matches > (buffer_size - max_matches_per_step)) {
+            flush_buffers(active_flushing_tile);
```
PointKernel, line 1260: There might be an opportunity to use the entire flushing tile to clear the buffer, rather than relying on active threads. I'll verify this in a separate PR.

PointKernel (approved):
Looks great to me. @sleeepyjack Thanks!

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
sleeepyjack: Same key but reset for the next probing slot. We use this flag to determine which threads need to write an output value and also also create a `cg::binary_partition` group from it at after each probing step.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
sleeepyjack: Threads that are in `flushing_tile` but are not in `active_flushing_tile` will never reach that line. Maybe we could rearrange the code a bit to make this possible.

PR #552, opened 2024-07-18
Remove unified memory from test util

ZelboK:
WSL and Windows do not support unified memory, so they will not be able to run the tests. This refactors the use of unified memory to just use the `counter_storage` in its place.


tests/test_utils.hpp
```diff
@@ -42,20 +44,18 @@ int count_if(Iterator begin, Iterator end, Predicate p, cudaStream_t stream = 0)
...
-  int device_id;
-  CUCO_CUDA_TRY(cudaGetDevice(&device_id));
-  CUCO_CUDA_TRY(cudaMemPrefetchAsync(count, sizeof(int), device_id, stream));
+  cuco::detail::counter_storage<int, cuda::thread_scope_device, Allocator> counter_storage(
+    Allocator{});
 
-  detail::count_if<<<grid_size, block_size, 0, stream>>>(begin, end, count, p);
-  CUCO_CUDA_TRY(cudaStreamSynchronize(stream));
+  counter_storage.reset(cuda::stream_ref(stream));
+  int* ptr = reinterpret_cast<int*>(counter_storage.data());
```
sleeepyjack: `counter_storage` allocates a `cuda::atomic` object. Reinterpreting it as the underlying type is undefined behavior, although it works in this case since they are layout-compatible for `int`.
The downstream `thrust` call expects a plain integer rather than an atomic.
Thus, can we instead of using the `counter_storage` allocate a single `int` "by-hand" using the `Allocator`?

ZelboK: Can you clarify which thrust call you're referring to? 

So are you saying to just do `int* ptr = allocator.allocate(1)` and send that pointer to the `count_if` kernel itself? I'm perfectly fine iwth that, however I was under the impression there was a reason `counter_storage` was needed for this case. @PointKernel Could you clarify please? Does it need to be atomic?

sleeepyjack: `thrust::count_if` expects a pointer to an integer rather than an atomic. That's why you had to `reinterpret_cast` the pointer in order to make it work.

ZelboK: Is it not calling the `count_if` kernels defined in `test_utils.cuh`?

sleeepyjack: Ah, sorry, I misread that. Yes, the call goes to the `count_if` kernel rather than `thrust`. However, the kernel is expecting an `int*` rather than a C++-style atomic so the problem still exists. We could change the kernel's signature to 
```cpp
CUCO_KERNEL void count_if(Iterator begin, Iterator end, cuda::atomic<int, cuda::thread_scope_device>* count, Predicate p)
```

and then use `.fetch_add` rather than `atomicAdd` in the body.

ZelboK: Yeah that makes sense to me. I wasn't sure if I was allowed to touch the kernel itself so I decided to do it this way. I'll do this instead, thanks!

PointKernel (approved):
LGTM. Thanks for the fix!

sleeepyjack (approved):

realhaik:
@ZelboK You are asking about WSL and Windows and how to run the tests, but cuCollections does not build/run on windows. I don't understand what am I missing, and what you are trying to do.

PointKernel:
> cuCollections does not build/run on windows

It does.

sleeepyjack:
@realhaik this is about WSL support which does run on Windows. UVM isn't supported on Windows platforms including WSL. This PR effectively fixed WSL support by removing all uses of UVM in cuco.

PR #556, opened 2024-07-22
Fix while loop predicate

PointKernel:
The existing while pred works properly only if `CGSize == 1`. This PR fixes the issue and slightly improves the performance by 1 or 2 percent in certain scenarios.


PointKernel:
cc @srinivasyadav18

sleeepyjack:
CI seems to be stuck

edit: long queue after a GPU runner outage

sleeepyjack (approved):

PR #559, opened 2024-07-24
Fix double hashing `with_hash_function`

PointKernel:
This PR includes several fixes and cleanups:

- Move probing scheme details to its own folder
- Add new overloads for double hashing so it can be (re)constructed from a tuple-like hasher
- Remove `noexcept` as needed since those functions may throw
- Improve bitwise compare traits to use CCCL utilities


PointKernel:
cc @srinivasyadav18

Let me know if this can help your potential tuning with double hashing in mixed join refactoring.

include/cuco/detail/probing_scheme/probing_scheme_impl.inl
```diff
@@ -141,6 +149,20 @@ __host__ __device__ constexpr auto double_hashing<CGSize, Hash1, Hash2>::with_ha
...
 
+template <int32_t CGSize, typename Hash1, typename Hash2>
+template <typename NewHash, typename Enable>
+__host__ __device__ constexpr auto double_hashing<CGSize, Hash1, Hash2>::with_hash_function(
+  NewHash const& hash) const
+{
+  static_assert(cuco::is_tuple_like<NewHash>::value,
+                "The given hasher must be a tuple-like object");
+
+  auto const [hash1, hash2] = cuco::pair{hash};
```
sleeepyjack, line 160: Why do we need to cast `hash` to a cuco pair first if we already established that it's a tuple-like object that supports structured bindings?

include/cuco/pair.cuh
```diff
@@ -142,6 +142,18 @@ template <class T1, class T2, class U1, class U2>
...
+struct is_cuco_pair : cuda::std::false_type {};
+
+template <typename F, typename S>
+struct is_cuco_pair<cuco::pair<F, S>> : cuda::std::true_type {};
+
+/// Trait determing whether a given type is tuple-like or not
+template <typename T>
+struct is_tuple_like : cuda::std::disjunction<is_cuco_pair<T>,
+                                              detail::is_std_pair_like<T>,
+                                              detail::is_cuda_std_pair_like<T>> {};
```
sleeepyjack, line 155: The other traits are located in `detail/pair/traits.hpp`. Shall we move them?

sleeepyjack (approved):
nice cleanups!

include/cuco/detail/probing_scheme/probing_scheme_impl.inl
PointKernel: Otherwise, we would do some dispatching like:
```cuda
auto const [hash1, hash2] = [&](){
  if constexpr (is_cuda_std_pair_like<NewHash>) {
    return {cuda::std::get<0>(hash), cuda::std::get<1>(hash)};
  }
  if constexpr (is_std_pair_like<NewHash>::value) {
    return {std::get<0>(hash), std::get<1>(hash)};
  }
}();
```

include/cuco/pair.cuh
PointKernel: Yeah, that's also the first place I wanted to put this trait but it requires `detail/pair/traits.hpp` to include `cuco/pair.cuh` header otherwise it won't work. The same for `detail/utility/traits.hpp`.

sleeepyjack: Cyclic include. I see. Thanks for the clarification!

PR #564, opened 2024-07-29
fix non-async overloads early return in `static_map`

srinivasyadav18:
Current implementation of `insert_or_assign` and `insert_or_apply` sync overloads return early and do not synchronize on stream.
This PR removes the early return from the async dispatch.


sleeepyjack (approved):
Ha, nice catch!

PointKernel (approved):

PR #565, opened 2024-07-29
Add host-bulk `for_each` for static_map

srinivasyadav18:
This PR adds two overloads of host-bulk `for_each` API for **static_map**.

1.  `for_each(Callback callback_op)` 
    executes **callback_op** on every filled element in the map.

2.  `for_each(InputIt first, IntputIt last, CallbackOp callback_op)` 
    executes **callback_op** on every matched slot with key equals to a key from the input sequence.


include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -681,6 +681,73 @@ class open_addressing_impl {
...
+   */
+  template <typename CallbackOp>
+  void for_each_async(CallbackOp&& callback_op, cuda::stream_ref stream)
+  {
+    using const_iterator = typename storage_ref_type::const_iterator;
+
+    auto const is_filled = open_addressing_ns::detail::slot_is_filled<has_payload, key_type>{
+      this->empty_key_sentinel(), this->erased_key_sentinel()};
+
+    thrust::for_each(
```
sleeepyjack: Ah, good use of `thrust::for_each` here. IIRC this function as is is still synchronous w.r.t. the host. You'd have to use `thrust::async::for_each` and/or `thrust::cuda::par_nosync` execution policy.

srinivasyadav18: Thanks! Resolved in https://github.com/NVIDIA/cuCollections/pull/565/commits/ded6180e14564ff1e88a2a9022d5e0dbda85b7b8.

PointKernel: Let's switch to `cub::DeviceFor`. `thrust::for_each` will not use the map allocator to deal with temporary memory allocation/deallocation.

sleeepyjack: Does `for_each` need any temp allocations?

PointKernel: TBH, I don't know. If the underlying algo is [cub::DeviceFor](https://nvidia.github.io/cccl/cub/api/structcub_1_1DeviceFor.html#_CPPv4N3cub9DeviceForE), then probably yes. If not, it might be switched to `cub::DeviceFor` in the future and the only way to make sure that we are using the given allocator is to invoke APIs with explicit memory control. 

With that, cuco details probably always want to use CUB instead of thrust.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -681,6 +682,89 @@ class open_addressing_impl {
...
+
+    // Allocate temporary storage
+    auto d_temp_storage = temp_allocator.allocate(temp_storage_bytes);
+
+    CUCO_CUDA_TRY(cub::DeviceFor::ForEachN(d_temp_storage,
+                                           temp_storage_bytes,
+                                           thrust::make_counting_iterator(static_cast<size_t>(0)),
+                                           this->capacity(),
+                                           op,
+                                           stream.get()));
```
```suggestion
                                           stream.get()));
    temp_allocator.deallocate(d_temp_storage, temp_storage_bytes);
```

sleeepyjack (approved):
LGTM apart from the missing `deallocate(tmp)`

PointKernel:
> This PR adds two overloads of host-bulk `for_each` API for **static_map**.
> 
> 1. `for_each(Callback callback_op)`
>    executes **callback_op** on every filled element in the map.
> 2. `for_each(InputIt first, IntputIt last, CallbackOp callback_op)`
>    executes **callback_op** on every matched slot with key equals to a key from the input sequence.

With the first overload no longer desired by libcudf, shall we remove it from the current PR?

sleeepyjack:
@PointKernel the first overload was requested (or better proposed by me) in this thread (NV internal): https://nvidia.enterprise.slack.com/archives/CRJDZ0SKG/p1721975163922999

I think this overload is useful in general as it enables functionality similar to the `begin/end` iterators in STL containers.

include/cuco/detail/open_addressing/kernels.cuh
```diff
@@ -182,6 +182,47 @@ CUCO_KERNEL __launch_bounds__(BlockSize) void erase(InputIt first,
...
+ * @tparam CallbackOp Unary callback functor or device lambda
+ * @tparam Ref Type of non-owning device ref allowing access to storage
+ *
+ * @param first Beginning of the sequence of input elements
+ * @param n Number of input elements
+ * @param callback_op Function to call on every element found in the container
+ * @param ref Non-owning container device ref used to access the slot storage
+ */
+template <int32_t CGSize, int32_t BlockSize, typename InputIt, typename CallbackOp, typename Ref>
+CUCO_KERNEL __launch_bounds__(BlockSize) void for_each(InputIt first,
```
PointKernel:
```suggestion
CUCO_KERNEL __launch_bounds__(BlockSize) void for_each_n(InputIt first,
```

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -681,6 +682,91 @@ class open_addressing_impl {
...
+
+    temp_allocator.deallocate(d_temp_storage, temp_storage_bytes);
+  }
+
+  /**
+   * @brief Asynchronously executes a callback on every element in the container whose key matches
+   * with a key from the input key sequence.
+   *
+   * @note Passes a copy of the element whose `key` matches with a key from the input key sequence
+   * to the callback.
```
```suggestion
   * @brief Asynchronously applies the given callable to all matches for each key within the specified range.
   *
   * @note Iterates over each key in the range `[first, last)` and applies the provided `callback_op` functor to all matches of the key in the data structure.
```

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -681,6 +682,91 @@ class open_addressing_impl {
...
+
+    CUCO_CUDA_TRY(cub::DeviceFor::ForEachN(nullptr,
+                                           temp_storage_bytes,
+                                           thrust::make_counting_iterator(static_cast<size_t>(0)),
+                                           this->capacity(),
+                                           op,
+                                           stream.get()));
+
+    // Allocate temporary storage
+    auto d_temp_storage = temp_allocator.allocate(temp_storage_bytes);
```
sleeepyjack: This allocation (unless it's stream-ordered/async allocator) prevents this function from being async w.r.t. the host. Does `ForEachN` even need a temp allocation?

PointKernel: > Does ForEachN even need a temp allocation?

For my education, if we know in advance that the algorithm won't use temporary data, can we bypass the "allocate" step and just pass `nullptr` to `d_temp_storage`?

sleeepyjack: If we pass `nullptr` wouldn't that trigger the dummy call again, i.e. only computing the temp memory size and not doing any actual work? We could pass an arbitrary pointer value but this only works if we are guaranteed that the implementation will acutually never dereference that pointer.

sleeepyjack: @srinivasyadav18 It got to my attention that there is [another overload](https://nvidia.github.io/cccl/cub/api/structcub_1_1DeviceFor.html?highlight=devicefor#_CPPv4I000EN3cub9DeviceFor12ForEachCopyNE11cudaError_t21RandomAccessIteratorT9NumItemsT3OpT12cudaStream_t) of `ForEachCopyN` that doesn't require the temp storage parameter. If we switch to that one we should be golden.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -705,7 +705,8 @@ class open_addressing_impl {
       }
     };
 
-    cub::DeviceFor::ForEachCopyN(storage_ref.data(), storage_ref.num_windows(), op, stream.get());
+    CUCO_CUDA_TRY(cub::DeviceFor::ForEachCopyN(
```
sleeepyjack, line 709: `CUCO_CUDA_TRY` will fetch the last CUDA error and thus synchronize the device with the host. We can't do that in a function marked `_async`. The error will pop up anyway if we call the non-`_async` variant after we call `stream.wait()`. For the `_async` variant, the user is in charge of proper CUDA error handling.

PointKernel: > fetch the last CUDA error and thus synchronize the device with the host. 

IIUC, it only synchronizes when the return value is not `cudaSuccess`. And in the case it throws, whether it synchronizes or not no longer matters.
https://github.com/NVIDIA/cuCollections/blob/356ce59bb9a403437bf7b1be668bc914ad2bf26f/include/cuco/detail/error.hpp#L54-L58

include/cuco/detail/open_addressing/kernels.cuh
```diff
@@ -182,6 +182,44 @@ CUCO_KERNEL __launch_bounds__(BlockSize) void erase(InputIt first,
...
+ * @tparam CGSize Number of threads in each CG
+ * @tparam BlockSize Number of threads in each block
+ * @tparam InputIt Device accessible input iterator whose `value_type` is
+ * convertible to the `key_type` of the data structure
+ * @tparam CallbackOp Type of unary callback function object
+ * @tparam Ref Type of non-owning device ref allowing access to storage
+ *
+ * @param first Beginning of the sequence of input elements
+ * @param n Number of input elements
+ * @param callback_op Function to call on every element found in the container
```
bdice: We need to clarify that this is acting on pairs of keys/values. Something like this, in all the docstrings?
```suggestion
 * @param callback_op Function to call on every key-value pair found in the container.
 * The function is called with a cuco::pair<Key, Value> and returns void.
```

include/cuco/detail/open_addressing/open_addressing_impl.cuh
sleeepyjack: Hmm, good point. Just to make sure I got this right: The cub/kernel call will immediately return and will not wait for the kernel to finish? Then we should probably wrap all kernel launches in our `_async` APIs with this macro, yes?

include/cuco/detail/open_addressing/kernels.cuh
sleeepyjack: This `open_addressing` impl class is the unified backend for all of the hashing-based containers, i.e., `static_map`, `static_set`, `static_multimap`, etc. thus we have to choose a generic docstring that works for all of them. It gives us headaches too sometimes I have to admit :D

In the actual `static_map` header we could be more precise but we often just copy&paste the docstring from the impl class to make our lives easier keeping the docs up to date.

PointKernel: > In the actual static_map header we could be more precise 

that's my plan for this PR as well. On the other hand, replacing `element` with `slot` is probably sufficient.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
PointKernel: > The cub/kernel call will immediately return and will not wait for the kernel to finish?

Yes

> Then we should probably wrap all kernel launches in our _async APIs with this macro, yes?

All cub and CUDA API invocations are guarded with `CUCO_CUDA_TRY`. If not, we need to add it. Kernels don't return CUDA error so no need to wrap them.

sleeepyjack: Thanks for the clarification. Thread resolved :)

PointKernel (approved):

PR #567, opened 2024-08-01
Fix a dangling reference warning with GCC-13

PointKernel:
Related to https://github.com/rapidsai/cudf/issues/16395

This PR fixes a dangling reference to a temporary warning with gcc13.


sleeepyjack (approved):
I wonder how that worked in the first place.

LGTM

PR #570, opened 2024-08-06
Add multimap conditional contains

PointKernel:
This PR adds `contains_if` and `contains_if_async` host-bulk APIs for the new multimap.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -111,7 +111,7 @@ ConfigureTest(STATIC_MULTIMAP_TEST
     static_multimap/custom_pair_retrieve_test.cu
     static_multimap/custom_type_test.cu
     static_multimap/heterogeneous_lookup_test.cu
-    static_multimap/insert_test.cu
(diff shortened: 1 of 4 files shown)


tests/CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ ConfigureTest(STATIC_MULTIMAP_TEST
     static_multimap/custom_pair_retrieve_test.cu
     static_multimap/custom_type_test.cu
     static_multimap/heterogeneous_lookup_test.cu
-    static_multimap/insert_test.cu
+    static_multimap/insert_contains_test.cu
```
sleeepyjack, line 114: I was a bit confused at first when reading the test name but I guess it makes sense in order to reduce code duplication.

sleeepyjack (approved):

tests/CMakeLists.txt
PointKernel: I cannot find an easy way to test `insert` without using `contains` and vice versa. :smiling_face_with_tear:

PR #571, opened 2024-08-07
Add multimap count and conditional insert

PointKernel:
This PR adds `count`, `insert_if` and `insert_if_async` APIs to the new multimap.

### DIFF
--- a/include/cuco/detail/static_multimap/static_multimap.inl
+++ b/include/cuco/detail/static_multimap/static_multimap.inl
@@ -152,11 +152,11 @@ template <class Key,
           class Allocator,
           class Storage>
 template <typename InputIt>
-static_multimap<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::size_type
-static_multimap<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::insert(
+void static_multimap<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::insert(
   InputIt first, InputIt last, cuda::stream_ref stream)
 {
-  return impl_->insert(first, last, ref(op::insert), stream);
+  this->insert_async(first, last, stream);
+  stream.wait();
 }
 
 template <class Key,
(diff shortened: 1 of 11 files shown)


include/cuco/detail/static_multimap/static_multimap.inl
```diff
@@ -249,6 +284,22 @@ void static_multimap<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator,
...
+          class Extent,
+          cuda::thread_scope Scope,
+          class KeyEqual,
+          class ProbingScheme,
+          class Allocator,
+          class Storage>
+template <typename InputIt>
+static_multimap<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::size_type
+static_multimap<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::count(
+  InputIt first, InputIt last, cuda::stream_ref stream) const noexcept
```
sleeepyjack: `count` needs to copy the result back to the host so there is a CUDA API call which should be wrapped with `CUCO_CUDA_TRY` which could potentially throw. So this function cannot be marked as `noexcept`.

include/cuco/static_multimap.cuh
```diff
@@ -297,11 +296,9 @@ class static_multimap {
    * @param first Beginning of the sequence of keys
    * @param last End of the sequence of keys
    * @param stream CUDA stream used for insert
-   *
-   * @return Number of successful insertions
    */
   template <typename InputIt>
-  size_type insert(InputIt first, InputIt last, cuda::stream_ref stream = {});
+  void insert(InputIt first, InputIt last, cuda::stream_ref stream = {});
```
sleeepyjack, line 301: So we remove the reduction from `insert` but introduce it again in `insert_if`?

include/cuco/detail/static_multimap/static_multimap.inl
PointKernel: Good catch. Addressed.

include/cuco/static_multimap.cuh
PointKernel: Yes, the cleanup I was aiming:

- multiset/multimap `insert` returning `void` since all inputs will be inserted so the reduction is redundant and more expensive than needed
- multiset/multimap `insert_if` returning `size_type` in case we want to track the number of conditional insertions
- multiset/multimap `insert_if_async` returning `void` for the cases that the #insertions is not desired

sleeepyjack: Totally makes sense. Thanks for the explanation.

sleeepyjack (approved):

PR #551, opened 2024-07-17
Add host-bulk `insert_or_apply` using shared_memory

srinivasyadav18:
This PR add's host-bulk `insert_or_apply` using **shared_memory** which could improve performance in low cardinality and very high mulitiplicty case.

PR #579, opened 2024-08-15
Fix `insert_or_apply`

srinivasyadav18:
This PR cleans up some of the issues occured during merge of #551. 

1. propagate the **key_eq** and **probing_scheme** from **global** `ref` to constructor of `shared_memory_ref` in **insert_or_apply_shmem** kernel. 
2. Disable **init** overload of `insert_or_apply` using **sfinae**, because `cuda::stream_ref` is default constructed, this can invoke the **init** overload even though the user calls **no-init** overload.

### DIFF
--- a/include/cuco/detail/static_map/kernels.cuh
+++ b/include/cuco/detail/static_map/kernels.cuh
@@ -202,8 +202,8 @@ CUCO_KERNEL __launch_bounds__(BlockSize) void insert_or_apply_shmem(
 
   auto shared_map     = SharedMapRefType{cuco::empty_key<Key>{ref.empty_key_sentinel()},
                                      cuco::empty_value<Value>{ref.empty_value_sentinel()},
-                                         {},
-                                         {},
+                                         ref.key_eq(),
+                                         ref.probing_scheme(),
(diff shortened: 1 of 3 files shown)


include/cuco/detail/static_map/kernels.cuh
```diff
@@ -202,8 +202,8 @@ CUCO_KERNEL __launch_bounds__(BlockSize) void insert_or_apply_shmem(
 
   auto shared_map     = SharedMapRefType{cuco::empty_key<Key>{ref.empty_key_sentinel()},
                                      cuco::empty_value<Value>{ref.empty_value_sentinel()},
-                                         {},
-                                         {},
+                                     ref.key_eq(),
+                                     ref.probing_scheme(),
                                          {},
```
PointKernel, line 207:
```suggestion
                                         cuco::thread_scope_block,
```
should this be thread scope block?

PointKernel: Thread scope block is defined in the helper function so this is correct

PointKernel (approved):

PR #580, opened 2024-08-15
Fix retrieve_all for containers with large capacity

### LINKED ISSUE #576: [BUG]: retrieve_all fails with over 1B items
### Is this a duplicate?

### Type of Bug

Silent Failure

### Describe the bug

We observed a hang in cuDF for hash-based groupby aggregations with over 1B items. I traced it to a hang in the `static_map` `retrieve_all` algorithm. The same hang can be observed in `static_set` benchmarks:
```
build/latest/benchmarks/STATIC_SET_BENCH -b 10 -d 0 -a Occupancy=0.5 -a NumInputs=1200000000 -a Key=I32 --run-once
```
This passes at Occupancy 0.9, so it must be a problem with the total size.

### How to Reproduce

Construct a static_map or static_set with a size greater than 1B elements (1.2B will hang). Call `retrieve_all`.

### Expected behavior

Results are returned without hanging.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

PointKernel:
Fix #576

This PR fixes the large input retrieve_all bug with a method similar to the streaming approach mentioned in https://github.com/NVIDIA/cccl/issues/1422#issuecomment-2285307255.

To be reverted once the CCCL fix is in place.

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -649,6 +649,7 @@ class open_addressing_impl {
     auto d_num_out      = reinterpret_cast<size_type*>(
       std::allocator_traits<temp_allocator_type>::allocate(temp_allocator, sizeof(size_type)));
 
+    // TODO: PR #580 to be reverted once https://github.com/NVIDIA/cccl/issues/1422 is resolved
     for (cuco::detail::index_type offset = 0;
          offset < static_cast<cuco::detail::index_type>(this->capacity());
          offset += stride) {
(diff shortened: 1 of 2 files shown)


sleeepyjack (approved):

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -639,45 +639,60 @@ class open_addressing_impl {
   template <typename OutputIt>
   [[nodiscard]] OutputIt retrieve_all(OutputIt output_begin, cuda::stream_ref stream) const
   {
-    std::size_t temp_storage_bytes = 0;
     using temp_allocator_type =
       typename std::allocator_traits<allocator_type>::template rebind_alloc<char>;
+
+    cuco::detail::index_type constexpr stride = std::numeric_limits<int32_t>::max();
```
sleeepyjack, line 645: We should add a note that this is a workaround and can be removed once the underlying cub issue has been solved

PointKernel: [b6c06ac](https://github.com/NVIDIA/cuCollections/pull/580/commits/b6c06ac38944501d21ee89c491d3f4f173bd9936) done

PR #585, opened 2024-08-23
Work around gcc bug in pair_alignment

miscco:
There seems to be a gcc bug where gcc versions prior to gcc13 get confused by passing in the output of cuda::std::bit_ceil into a constant expression.

This only manifests when trying to use the constant expression in an alignas direction.

However, we can convince gcc that it is indeed a constant expression by storing it in a constexpr variable before passing it along.

This ensures that cuDF can build against current CCCL ToT

### DIFF
--- a/include/cuco/detail/bitwise_compare.cuh
+++ b/include/cuco/detail/bitwise_compare.cuh
@@ -18,6 +18,7 @@
 
 #include <cuco/utility/traits.hpp>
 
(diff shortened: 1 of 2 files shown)


bernhardmgruber (approved):

sleeepyjack (approved):
Holy moly, tracing down this bug must've been no fun at all. Kudos!

PR #591, opened 2024-08-27
Add diagnostic to detect unsupported or missing CCCL

### LINKED ISSUE #590: [ENHANCEMENT]: Emit diagnostic when detected CCCL version is not supported

cuco's only dependency is CCCL and we make heavy use of new libcu++/CUB/Thrust features to improve our QoL as library devs.
Thus we require a fairly new CCCL version (IIRC currently it's v2.5).

Building cuco without any auxiliary dependency management like cmake/CPM can result in compilation errors, e.g., when cuco picks up the CCCLs version that shipped with an outdated CTK.

Check if the detected CCCL library version is sufficient and emit a compile-time diagnostic in case if it's not supported.
Rather than having to dive through some (sometimes obscure) compiler errors, the user is presented with a one-line error message.

_No response_

### Additional context

Related issue: #588

sleeepyjack:
Closes #590

### DIFF
--- a/include/cuco/detail/__config
+++ b/include/cuco/detail/__config
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <nv/target>
+#include <cuda/std/version>
 
 #if !defined(__CUDACC_VER_MAJOR__) || !defined(__CUDACC_VER_MINOR__)
 #error "NVCC version not found"
@@ -30,6 +31,10 @@
 
 #if !defined(__CUDACC_EXTENDED_LAMBDA__)
 #error "Support for extended device lambdas is required (nvcc flag --expt-extended-lambda)"
+#endif
+
+#if !defined(CCCL_VERSION) || (CCCL_VERSION < 2005000)
+#error "CCCL version 2.5.0 or later is required"
 #endif
 
 // WAR for libcudacxx/296


sleeepyjack:
For the `__config` header, should we include it at the top of every public-facing header? My understanding is that this header should be parsed first in order to define some crucial symbols and check if the configuration is correct. Currently only a subset of our headers actually includes `__config`.

PointKernel:
> For the `__config` header, should we include it at the top of every public-facing header? My understanding is that this header should be parsed first in order to define some crucial symbols and check if the configuration is correct. Currently only a subset of our headers actually includes `__config`.

Not necessary IMO. The build will fail anyway.

sleeepyjack:
> The build will fail anyway.

True. But IIUC if `__config` is the first header that's parsed then we will see the diagnostic being triggered instead of an obscure compilation error somewhere else. Not 100% I understand this right.

PointKernel:
> > The build will fail anyway.
> 
> True. But IIUC if `__config` is the first header that's parsed then we will see the diagnostic being triggered instead of an obscure compilation error somewhere else. Not 100% I understand this right.

Not worth the effort IMO. cuco users will inevitably encounter template build errors, so prioritizing the clarity of config errors isn't a major concern.

include/cuco/detail/__config
```diff
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <nv/target>
+#include <cuda/std/version>
```
PointKernel, line 20: question: is it possible that this header doesn't exist?

PointKernel (approved):
A small question but I don't need to review it again.

include/cuco/detail/__config
sleeepyjack: Yeah, that was also my concern. Kind of an hen-egg problem so I tested it in godbold. It seems like the diagnostic is always triggered (first), even if the header doesn't exist. I still need to wrap my head around what's exactly going on but I think this PR is good to merge as of now.

PR #573, opened 2024-08-08
Add cuco::bloom_filter

sleeepyjack:
Superseeds #101

Implementation of a GPU "Blocked Bloom Filter".

This PR is an updated/optimized version of #101 and features the following improvements:

- Incorporate the new library design
- Improve performance by computing the key's bit pattern based on a single hash value instead of using a double hashing derivative

PR #597, opened 2024-08-30
Use `cuda::std::byte` as the default device storage type

PointKernel:
Address https://github.com/NVIDIA/cuCollections/pull/573#discussion_r1737574327

This PR updates the existing code to use `cuda::std::byte` as the default device data type instead of `std::byte`. This change addresses potential issues where `cuda::std::` utilities cannot be applied to `std::byte` when relaxed constexprs are disabled.

### DIFF
--- a/include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
+++ b/include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
@@ -18,12 +18,12 @@
 #pragma once
 
 #include <cuda/std/array>
+#include <cuda/std/cstddef>
 #include <cuda/stream_ref>
 #include <thrust/device_malloc_allocator.h>
 #include <thrust/device_vector.h>
 
 #include <climits>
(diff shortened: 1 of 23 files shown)


include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.cuh
```diff
@@ -18,12 +18,12 @@
 #pragma once
 
 #include <cuda/std/array>
+#include <cuda/std/cstddef>
 #include <cuda/stream_ref>
 #include <thrust/device_malloc_allocator.h>
 #include <thrust/device_vector.h>
 
 #include <climits>
-#include <cstddef>
```
sleeepyjack: We need to keep `<cstddef>` since this header uses `std::size_t` in line 84. We could also switch to `cuda::std` types for anything in `<cstddef> and <cstdint>`.

sleeepyjack (approved):

PR #598, opened 2024-09-01
Improve rebind utilities for cuco hash tables

PointKernel:
This PR renames all `with_*` member functions to `rebind_*` for improved clarity.

The legacy `with_operators` will be removed once libcudf is migrated to use the new `rebind_operators`.

### DIFF
--- a/include/cuco/detail/static_multimap/static_multimap_ref.inl
+++ b/include/cuco/detail/static_multimap/static_multimap_ref.inl
@@ -334,6 +334,33 @@ template <typename Key,
           typename ProbingScheme,
           typename StorageRef,
           typename... Operators>
+template <typename NewHash>
+__host__ __device__ constexpr auto
+static_multimap_ref<Key, T, Scope, KeyEqual, ProbingScheme, StorageRef, Operators...>::
+  rebind_hash_function(NewHash const& hash) const
+{
+  auto const probing_scheme = this->probing_scheme().rebind_hash_function(hash);
+  return static_multimap_ref<Key,
+                             T,
+                             Scope,
+                             KeyEqual,
+                             cuda::std::decay_t<decltype(probing_scheme)>,
+                             StorageRef,
+                             Operators...>{cuco::empty_key<Key>{this->empty_key_sentinel()},
+                                           cuco::empty_value<T>{this->empty_value_sentinel()},
+                                           this->key_eq(),
+                                           probing_scheme,
+                                           {},
+                                           this->storage_ref()};
+}
+
+template <typename Key,
+          typename T,
+          cuda::thread_scope Scope,
+          typename KeyEqual,
+          typename ProbingScheme,
+          typename StorageRef,
+          typename... Operators>
 template <typename CG, cuda::thread_scope NewScope>
 __device__ constexpr auto
 static_multimap_ref<Key, T, Scope, KeyEqual, ProbingScheme, StorageRef, Operators...>::make_copy(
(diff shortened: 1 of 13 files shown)


include/cuco/detail/static_multimap/static_multimap_ref.inl
```diff
@@ -294,11 +294,22 @@ template <typename Key,
...
+  ProbingScheme,
+  StorageRef,
+  Operators...>::with_operators(NewOperators...) const noexcept
 {
   return static_multimap_ref<Key, T, Scope, KeyEqual, ProbingScheme, StorageRef, NewOperators...>{
-    std::move(*this)};
+    cuco::empty_key<Key>{this->empty_key_sentinel()},
+    cuco::empty_value<T>{this->empty_value_sentinel()},
+    this->key_eq(),
+    impl_.probing_scheme(),
```
sleeepyjack:
```suggestion
    this->probing_scheme(),
```

include/cuco/detail/static_multimap/static_multimap_ref.inl
```diff
@@ -327,6 +338,54 @@ __host__ __device__ auto constexpr static_multimap_ref<
     impl_.storage_ref()};
```
sleeepyjack:
```suggestion
    this->storage_ref()};
```

Same with `impl_.probing_scheme()` two lines above

include/cuco/static_map_ref.cuh
```diff
@@ -227,11 +227,7 @@ class static_map_ref
   [[nodiscard]] __host__ __device__ constexpr auto probing_scheme() const noexcept;
 
   /**
-   * @brief Creates a reference with new operators from the current object.
-   *
-   * @deprecated This function is deprecated. Use the new `with_operators` instead.
```
sleeepyjack, line 232: Are we dropping the deprecation warning?

tests/static_multiset/custom_count_test.cu
```diff
@@ -61,21 +61,27 @@ void test_custom_count(Set& set, size_type num_keys)
 {
   using Key = typename Set::key_type;
 
+  auto const hash = []() {
+    if constexpr (cuco::is_double_hashing<typename Set::probing_scheme_type>::value) {
+      return cuco::pair{custom_hash{}, custom_hash{}};
```
sleeepyjack: Shall we use `cuda::std::tuple` here instead?

include/cuco/static_map_ref.cuh
PointKernel: I removed this API directly since it's marked as deprecated for a while and no exiting RAPIDS code uses it.

tests/static_multiset/custom_count_test.cu
PointKernel: I’ve been thinking about this: `cuco::pair` can be constructed from tuple-like inputs, so it works with types like `cuda::std::pair/tuple` and `std::pair/tuple`. Initially, we opted for `cuco::pair` for convenience, but I’m now questioning if that’s a convenience we still need to maintain. What’s your take on this?

sleeepyjack: I would say `cuco::pair` is only needed when interacting with table slots as it guarantees proper alignment. For any other case `cuda::std::tuple` is the more generic choice.

PointKernel: > cuco::pair is only needed when interacting with table slots as it guarantees proper alignment.

Good point.

sleeepyjack (approved):
Thanks! Nice work.

PR #608, opened 2024-10-02
Add `compute_hash` overloads for `std::byte*` to maintain backward compatibility

PointKernel:
Unblock https://github.com/rapidsai/cugraph/pull/4686

The `compute_hash` function is a host-device function, so the appropriate argument type is `cuda::std::byte`. However, to ensure backward compatibility and avoid breaking existing use cases, this PR adds overloads that accept `std::byte`.

### DIFF
--- a/include/cuco/detail/hash_functions/murmurhash3.cuh
+++ b/include/cuco/detail/hash_functions/murmurhash3.cuh
@@ -203,6 +203,26 @@ struct MurmurHash3_32 {
     return h1;
   }
 
+  /**
+   * @brief Returns a hash value for its argument, as a value of type `result_type`.
+   *
+   * @note This API is to ensure backward compatibility with existing use cases using `std::byte`.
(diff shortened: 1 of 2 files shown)


sleeepyjack (approved):
The PR itself LGTM and can be merged to unblock cugraph.

However, I find this workaround a bit nasty given that `std::byte` and `cuda::std::byte` are virtually the same thing. In my book, there should exist at least an implicit conversion between the two.

Also, I think cugraph should use `cuda::std::byte` here in the first place.

PointKernel:
> Also, I think cugraph should use cuda::std::byte here in the first place.

Agreed. I will open a cugraph PR for this once the current work is merged

PointKernel:
> Also, I think cugraph should use cuda::std::byte here in the first place.

Opened https://github.com/rapidsai/cugraph/pull/4707 to fix this

PR #609, opened 2024-10-02
CUPTI metrics for bloom_filter benchmarks

sleeepyjack:
This PR introduces collecting CUPTI metrics (DRAM throughput, cache hit rates, etc.) during the `cuco::bloom_filter` benchmark.

Additionally, the benchmark now uses the default `pattern_bits` value for a given policy instead of hardcoding it to 8.


PointKernel:
@sleeepyjack, could you please share the new benchmark results in the comments for reference?

sleeepyjack:
Here they are: https://github.com/NVIDIA/cuCollections/pull/573#issuecomment-2389992664

PointKernel (approved):

PR #613, opened 2024-10-09
Add Godbolt precommit hook

PointKernel:
TBD


PointKernel:
As discussed offline, this should be a Github Action instead of a pre-commit hook.

PointKernel:
Replaced by #620

PR #620, opened 2024-10-15
Add pre-commit hook checking example links

PointKernel:
Replace #613

This PR introduces a pre-commit hook that automatically verifies and updates the Godbolt links in the `README` when there are discrepancies between the example files and their associated links.


PointKernel:
Corresponding CI will be added in a separate PR.

PointKernel:
I have observed network traffic issues when invoking the script too frequently:
```bash
Error: Request failed with status code 502
Response: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
<HTML><HEAD><META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=iso-8859-1">
<TITLE>ERROR: The request could not be satisfied</TITLE>
</HEAD><BODY>
<H1>502 ERROR</H1>
<H2>The request could not be satisfied.</H2>
<HR noshade size="1px">
The origin closed the connection.
We can't connect to the server for this app or website at this time. There might be too much traffic or a configuration error. Try again later, or contact the app or website owner.
<BR clear="all">
If you provide content to customers through CloudFront, you can find steps to troubleshoot and help prevent this error by reviewing the CloudFront documentation.
<BR clear="all">
<HR noshade size="1px">
<PRE>
Generated by cloudfront (CloudFront)
Request ID: lLJehNBQT-6cWYEDUaVC5Eyi0cKe7Eh28W9krnuSgweX4-ht0G7lyw==
</PRE>
<ADDRESS>
</ADDRESS>
</BODY></HTML>
```

sleeepyjack:
It seems we're getting rate limited at the Compiler Explorer’s /api/shortener endpoint.

I did some digging and found that we can just append the zlib-compressed config JSON to the URL. 🤯 

This means we can generate a link without having to query the API (although it's not shortened):

```python
import json
import zlib
import base64
import urllib.parse

def create_compiler_explorer_url():
    # Prepare the configuration
    config = {
        "sessions": [
            {
                "id": 1,
                "language": "c++",
                "source": '''#include <iostream>

int main() {
    std::cout << "Hello, World!" << std::endl;
    return 0;
}
''',
                "compilers": [
                    {
                        "id": "g122",  # GCC 12.2
                        "options": "-O2",
                        "libs": []
                    }
                ]
            }
        ]
    }
    
    # Convert the config to JSON
    config_json = json.dumps(config, separators=(',', ':'))
    
    # Compress the JSON using zlib (deflate)
    compressed = zlib.compress(config_json.encode('utf-8'))
    
    # Base64 encode the compressed data
    encoded = base64.urlsafe_b64encode(compressed).decode('utf-8')
    
    # Replace problematic unicode characters
    # As per the documentation, map characters in the range \u007F-\uFFFF
    encoded = ''.join(
        c if '\u0000' <= c <= '\u007F' else '\\u{:04x}'.format(ord(c))
        for c in encoded
    )
    
    # Construct the final URL
    url = f'https://godbolt.org/clientstate/{encoded}'
    
    return url

# Example usage
url = create_compiler_explorer_url()
print('Compiler Explorer URL:', url)
```

PointKernel:
@sleeepyjack works great IMO. What do you think?

sleeepyjack:
Nice.

I have a few things:

- IIRC some browsers have a 2k character limit for URLs. I don't think we're currently hitting that limit but it would be good to add a check.
- The new approach no longer requires an API access or internet connection. This means we can re-introduce the pre-commit hook again for verifying if the links are valid. We can do so by encoding each example source file and then check if the base64 string matches with the one in the current godbolt link.
- This workflow is similar to what we do with the devcontainer.json: We have two scripts. [make_devcontainer.sh](https://github.com/NVIDIA/cuCollections/blob/dev/.devcontainer/make_devcontainers.sh) which creates the config, and [verify_devcontainer.py](https://github.com/NVIDIA/cuCollections/blob/dev/.devcontainer/verify_devcontainer.sh) to check if the configs are valid. I would suggest we use a similar naming scheme, i.e., `ci/make\update_example_links.py` and `ci/verify_example_links.py`

PointKernel:
> I don't think we're currently hitting that limit but it would be good to add a check.

We are hitting that limit, and 8 of the 13 links have exceeded 2000 characters. I've tested those links with Chrome, Firefox and Edge and they work just fine.


> This means we can re-introduce the pre-commit hook again for verifying if the links are valid.

The pre-commit hook images don't have the extra python modules installed so if we want to proceed with pre-commit, we need to convert this script to bash. I'm inclined to have the current PR supporting auto-godbolt, fixing existing mismatches, and addressing CI in a separate PR.


> I would suggest we use a similar naming scheme, i.e., ci/make\update_example_links.py and ci/verify_example_links.py

Sure thing. Note this is only needed if we want to use github actions instead of pre-commit hook. No need to have both at the same time.

sleeepyjack:
> We are hitting that limit, and 8 of the 13 links have exceeded 2000 characters. I've tested those links with Chrome, Firefox and Edge and they work just fine.

Ok, my knowledge there might be outdated. Thanks for verifying!

> The pre-commit hook images don't have the extra python modules installed so if we want to proceed with pre-commit, we need to convert this script to bash.

Which modules are missing? I was under the impression that `json, zlib, urllib, base64` all come with the base python3 install. Porting this to bash might be a bit of a pain.

> I'm inclined to have the current PR supporting auto-godbolt, fixing existing mismatches, and addressing CI in a separate PR.

Yep, we don't want autofixing at the moment. The CI check should just fail in case the base64 strings don't match. It's the contributors responsibility to update the links accordingly. The devcontainer check does the same thing (but with a GH action rather than pre-commit).

> Note this is only needed if we want to use github actions instead of pre-commit hook. No need to have both at the same time.

Dealer's choice. If we can make pre-commit work without any additional module dependencies then that's probably the path of least resistance.

ci/pre-commit/example_links.py
```diff
@@ -0,0 +1,164 @@
...
+    readme_content = open_file(readme_path)
+    lines = readme_content.splitlines()
+
+    updated_lines = []
+    mismatched_files = []
+    # Iterate through each line of the README
+    for line in lines:
+        updated_line = line
+        # Check each file in the url_table
+        for file_name, url in url_table.items():
```
sleeepyjack, line 85: As discussed offline, we should throw/print an error in case an example source file has no matching line in the readme

sleeepyjack (approved):
Nice QoL improvement. Thank you!

PR #629, opened 2024-10-29
Add count APIs for set and map

PointKernel:
This PR adds host and device `count` APIs for `static_set` and `static_map`

### DIFF
--- a/include/cuco/detail/static_map/static_map_ref.inl
+++ b/include/cuco/detail/static_map/static_map_ref.inl
@@ -1367,5 +1367,59 @@ class operator_impl<
   }
 };
 
+template <typename Key,
+          typename T,
+          cuda::thread_scope Scope,
+          typename KeyEqual,
+          typename ProbingScheme,
+          typename StorageRef,
+          typename... Operators>
+class operator_impl<
+  op::count_tag,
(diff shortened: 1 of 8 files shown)


include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -1367,5 +1367,59 @@ class operator_impl<
...
+   * @brief Counts the occurrence of a given key contained in map
+   *
+   * @tparam ProbeKey Input type
+   *
+   * @param key The key to count for
+   *
+   * @return Number of occurrences found by the current thread
+   */
+  template <typename ProbeKey>
+  __device__ size_type count(ProbeKey const& key) const noexcept
```
sleeepyjack, line 1400: Question for my understanding: can `__device__ static_set::count(key)` return a value that is larger than 1 due to heterogeneous lookup?

If so, does that mean we're never using the early exit strategy for this operation, i.e., return from probing once we found the first match?

sleeepyjack (approved):

include/cuco/detail/static_map/static_map_ref.inl
PointKernel: I don't think so. It uses the return value of `contains` which exits whenever a match is found. The probing sequence won't continue after the first match.

PR #636, opened 2024-11-06
Migrate Doxygen setup to Conda for consistent version management

PointKernel:
This PR moves the Doxygen installation from the default Ubuntu version via `apt-get` to Conda, preventing potential version conflicts during system updates.


.github/workflows/pr.yml
```diff
@@ -37,12 +37,13 @@ jobs:
...
       - name: Checkout repo
         uses: actions/checkout@v3
-      - name: Install Doxygen
-        run: |
-          sudo apt-get update -q
-          sudo apt-get install -y doxygen
-      - name: Check Doxygen docs
+      - name: Install and Check Doxygen
         run: |
+          conda init bash
```
bdice: Let’s move all of this to a standalone bash script.

bdice: Model it after https://github.com/rapidsai/cudf/blob/branch-24.12/ci/check_style.sh.

sleeepyjack: Do we need conda after all? We could just wget one of the binary releases from the doxygen GH repo. I'm not saying I'm against conda, I'm just wondering if it comes with any additional overhead compared to downloading the raw binary and placing it in an executable path.

ci/doxygen_check.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
PointKernel:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

sleeepyjack (approved):

PR #643, opened 2024-11-20
Add retrieve for new map/multimap

PointKernel:
This PR introduces `retrieve` APIs for map and multimap, while also removing the outdated and odd `pair_count/retrieve` tests.


include/cuco/detail/static_map/static_map.inl
```diff
@@ -612,6 +612,26 @@ static_map<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::
...
+          class ProbingScheme,
+          class Allocator,
+          class Storage>
+template <typename InputIt, typename OutputIt1, typename OutputIt2>
+std::pair<OutputIt1, OutputIt2>
+static_map<Key, T, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::retrieve(
+  InputIt first,
+  InputIt last,
+  OutputIt1 output_probe,
+  OutputIt2 output_match,
```
```suggestion
  OutputProbeIt output_probe,
  OutputMatchIt output_match,
```

Just a suggestion. Probably not worth the refactoring effort.

include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -1420,5 +1420,73 @@ class operator_impl<
...
+            class InputProbeIt,
+            class OutputProbeIt,
+            class OutputMatchIt,
+            class AtomicCounter>
+  __device__ void retrieve(cooperative_groups::thread_block const& block,
+                           InputProbeIt input_probe_begin,
+                           InputProbeIt input_probe_end,
+                           OutputProbeIt output_probe,
+                           OutputMatchIt output_match,
+                           AtomicCounter* atomic_counter) const
```
sleeepyjack, line 1484:
```suggestion
                           AtomicCounter& atomic_counter) const
```

I still think passing by mutable reference is the better way than passing by pointer. Anyway, we can move this to a cleanup PR.

include/cuco/static_multimap.cuh
```diff
@@ -666,6 +666,40 @@ class static_multimap {
...
+   * the range
+   * `[first, last)`
+   *
+   * If key `k = *(first + i)` has a match `m` in the multimap, copies a `cuco::pair{k, m}` to
+   * unspecified locations in `[output_begin, output_end)`. Else, does nothing.
+   *
+   * @note This function synchronizes the given stream.
+   * @note Behavior is undefined if the size of the output range exceeds
+   * `std::distance(output_begin, output_end)`.
+   * @note Behavior is undefined if the given key has multiple matches in the set.
```
sleeepyjack, line 680: Undefined or rather "will only retrieve the first occurrence found"?

sleeepyjack (approved):
Some minor suggestions otherwise LGTM. Thanks!

include/cuco/detail/static_map/static_map.inl
PointKernel: Fixed

include/cuco/static_multimap.cuh
PointKernel: The undefined part is that if multiple matches exist in the map, the "first" occurrence is not deterministic.

PR #645, opened 2024-11-25
Add host find APIs taking custom key equality and hasher

PointKernel:
This PR introduces host find APIs for the hash set, enabling queries using a different key_eq and hash function. It can be used to improve cudf distinct join performance.

### DIFF
--- a/include/cuco/detail/static_set/static_set.inl
+++ b/include/cuco/detail/static_set/static_set.inl
@@ -344,6 +344,29 @@ template <class Key,
           class ProbingScheme,
           class Allocator,
           class Storage>
+template <typename InputIt, typename ProbeEqual, typename ProbeHash, typename OutputIt>
+void static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::find_async(
+  InputIt first,
+  InputIt last,
+  ProbeEqual const& probe_equal,
+  ProbeHash const& probe_hash,
+  OutputIt output_begin,
+  cuda::stream_ref stream) const
+{
+  impl_->find_async(first,
+                    last,
+                    output_begin,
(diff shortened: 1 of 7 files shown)


PointKernel:
To be merged after #648

include/cuco/detail/static_set/static_set.inl
```diff
@@ -337,6 +337,29 @@ void static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>
...
+          class Allocator,
+          class Storage>
+template <typename InputIt, typename ProbeEqual, typename ProbeHash, typename OutputIt>
+void static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::find_async(
+  InputIt first,
+  InputIt last,
+  ProbeEqual const& probe_equal,
+  ProbeHash const& probe_hash,
+  OutputIt output_begin,
+  cuda::stream_ref stream) const
```
sleeepyjack, line 354:
```suggestion
void static_set<Key, Extent, Scope, KeyEqual, ProbingScheme, Allocator, Storage>::find_async(
  InputIt first,
  InputIt last,
  OutputIt output_begin,
  cuda::stream_ref stream,
  ProbeEqual const& probe_equal,
  ProbeHash const& probe_hash) const
```

Would this reordering make sense? We could default `probe_equal` and `probe_hash` to the ones the container currently uses. This would save us one more overload.

sleeepyjack (approved):
One suggestion/question otherwise LGTM.

include/cuco/detail/static_set/static_set.inl
PointKernel: We arrange arguments in the order from input to output, with the stream parameter always placed last and given a default value.

PR #648, opened 2024-11-28
Move kernels to their own namespaces to avoid build conflicts

PointKernel:
Replace #644.

The original PR was accidentally closed since I force-pushed to my `dev` branch.

Including the Bloom filter header alongside any other hash table header leads to build-time failures due to naming conflicts (see https://godbolt.org/z/P66ah8hah). This PR resolves the issue by relocating the open-addressing and Bloom filter kernels into separate namespaces to prevent such conflicts.

### DIFF
--- a/include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
+++ b/include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
@@ -198,7 +198,7 @@ class bloom_filter_impl {
     auto const grid_size =
       cuco::detail::grid_size(num_keys, cg_size, cuco::detail::default_stride(), block_size);
 
-    detail::add_if_n<cg_size, block_size>
+    bloom_filter_ns::detail::add_if_n<cg_size, block_size>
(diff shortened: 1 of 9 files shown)


include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -198,7 +198,7 @@ class bloom_filter_impl {
     auto const grid_size =
       cuco::detail::grid_size(num_keys, cg_size, cuco::detail::default_stride(), block_size);
 
-    detail::add_if_n<cg_size, block_size>
+    bloom_filter_ns::detail::add_if_n<cg_size, block_size>
```
sleeepyjack: `cuco::bloom_filter_ns::detail` or `cuco::detail::bloom_filter_ns`? I think the latter makes it more clear that these `_ns` namespaces are not meant to be publicly available.

PointKernel: Makes sense. Updated.

sleeepyjack (approved):

Issue #653, opened 2024-12-10
[ENHANCEMENT]: Consider removing the use of `argument_type` and `result_type` from `hasher` in cuco `default_filter_policy`


mhaseeb123:
### Is your feature request related to a problem? Please describe.

The member type std::hash<T>::argument_type is deprecated in C++17 and removed in C++20, so we should consider syncing with STL and remove them from our hashers as well as their use in bloom filter policies.

Ref: https://en.cppreference.com/w/cpp/utility/hash

_Originally posted by @bdice in https://github.com/rapidsai/cudf/pull/17289#discussion_r1876965059_

### Describe the solution you'd like

Consider refactoring to not use hasher::argument_type and hasher::result_type across cuCollections.

### Describe alternatives you've considered

Keep using the deprecated hasher type aliases until C++20

PointKernel:
I thought about mentioning this during the code review but didn’t want to come across as nitpicking. Instead of https://github.com/NVIDIA/cuCollections/blob/b8429d44314a02da9b5ebc17fa0a9284543aa825/include/cuco/detail/bloom_filter/arrow_filter_policy.cuh#L84-L85

we could make the hasher a template type:
```cpp
template<class Key, template <typename> class Hash>
class arrow_filter_policy {
  ...
};
```

To use it, users can simply pass the hasher type without repeating the `Key` type again, e.g.
```cpp
auto p = arrow_filter_policy<Key, cuco::xxhash_32>{...};
```

mhaseeb123:
> I thought about mentioning this during the code review but didn’t want to come across as nitpicking.

No worries about that. I am always for pursuing excellence and attention to the detail is how we all learn. 😄 

I think this is a great idea. Let's do this. But we still would be using the member types (`argument_type` and `result_type`) from the hasher which Bradley mentioned

PointKernel:
>  we still would be using the member types (argument_type and result_type) from the hasher

Half-half. The `argument_type` will simply be `key_type (using key_type = Key)`. However, retaining `result_type` (should be `hash_value_type` of the policy) is still desirable. The STL simplifies hashers by always using `std::size_t` as the return type, but this approach is suboptimal, especially in terms of performance and for GPUs. Such oversimplification does not fit our goals and IMO should be avoided.

mhaseeb123:
> Such oversimplification does not fit our goals and IMO should be avoided.

[...]

PR #654, opened 2024-12-10
Minor improvements for arrow_filter_policy

mhaseeb123:
This PR make the hasher to `arrow_filter_policy` a template type for users to simply pass the hasher without specifying the `Key` type twice. See this comment for discussion: https://github.com/NVIDIA/cuCollections/issues/653#issuecomment-2530111406. Also update member types `argument_type` and `hash_result_type` to directly use `Key` and `std::uint64_t` for this policy.

### DIFF
--- a/include/cuco/detail/bloom_filter/arrow_filter_policy.cuh
+++ b/include/cuco/detail/bloom_filter/arrow_filter_policy.cuh
@@ -84,11 +84,10 @@ namespace cuco::detail {
 template <class Key, template <typename> class XXHash64>
 class arrow_filter_policy {
  public:
-  using hasher             = XXHash64<Key>;  ///< 64-bit XXHash hasher for Arrow bloom filter policy
-  using word_type          = std::uint32_t;  ///< uint32_t for Arrow bloom filter policy
-  using key_type = Key;  ///< Hash function input type
-  using hash_value_type   = uint64_t;
-    std::declval<hash_argument_type>()));  ///< hash function output type
+  using hasher          = XXHash64<Key>;  ///< 64-bit XXHash hasher for Arrow bloom filter policy
+  using word_type       = std::uint32_t;  ///< uint32_t for Arrow bloom filter policy
+  using key_type        = Key;            ///< Hash function input type
+  using hash_value_type = std::uint64_t;  ///< hash function output type
 
   static constexpr uint32_t bits_set_per_block = 8;  ///< hardcoded bits set per Arrow filter block
   static constexpr uint32_t words_per_block    = 8;  ///< hardcoded words per Arrow filter block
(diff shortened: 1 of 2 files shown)


include/cuco/detail/bloom_filter/arrow_filter_policy.cuh
```diff
@@ -81,10 +81,10 @@ namespace cuco::detail {
...
  */
-template <class Key, class XXHash64>
+template <class Key, template <typename> class XXHash64>
 class arrow_filter_policy {
  public:
-  using hasher             = XXHash64;       ///< 64-bit XXHash hasher for Arrow bloom filter policy
+  using hasher             = XXHash64<Key>;  ///< 64-bit XXHash hasher for Arrow bloom filter policy
   using word_type          = std::uint32_t;  ///< uint32_t for Arrow bloom filter policy
   using hash_argument_type = typename hasher::argument_type;  ///< Hash function input type
   using hash_result_type   = decltype(std::declval<hasher>()(
```
PointKernel:
```suggestion
  using key_type = Key;  ///< Hash function input type
  using hash_value_type   = uint64_t;
```
We could still keep the current definition as it is and adding a build time check verifying it's the same as `uint64_t`.

mhaseeb123: With this, we can just revert the change in libcudf's `XXHash64` and I will update the `arrow_fitler_policy.cuh` that I copied over to libcudf as well (which will be deleted with the cuco bump anyway)

include/cuco/detail/bloom_filter/arrow_filter_policy.cuh
```diff
@@ -81,14 +80,13 @@ namespace cuco::detail {
...
+template <class Key, template <typename> class XXHash64>
 class arrow_filter_policy {
  public:
-  using hasher             = XXHash64;       ///< 64-bit XXHash hasher for Arrow bloom filter policy
+  using hasher             = XXHash64<Key>;  ///< 64-bit XXHash hasher for Arrow bloom filter policy
   using word_type          = std::uint32_t;  ///< uint32_t for Arrow bloom filter policy
-  using hash_argument_type = typename hasher::argument_type;  ///< Hash function input type
-  using hash_result_type   = decltype(std::declval<hasher>()(
-    std::declval<hash_argument_type>()));  ///< hash function output type
+  using hash_argument_type = Key;            ///< Hash function input type
```
mhaseeb123: ~Kept the names of these member types consistent with the rest of the file as well as the members of `default_filter_policy`~ Nvm, let's update here at least

PointKernel (approved):
Thanks for the quick fix @mhaseeb123

PR #658, opened 2025-01-16
Add docs for mixin operators

### LINKED ISSUE #371: [ENHANCEMENT]: Mixin operators should be documented in the public API

Currently, the API docs for the mixin operators are hidden in the detail namespace [here](https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/open_addressing_ref_impl.cuh), which isn't ideal.

We should either move or copy these docs to the class bodies of our `*_ref` classes since this is where the user expects them to be.

_No response_

### Additional context

_No response_

sleeepyjack:
This PR adds publicly available inline docs for device operators.

Closes #371

### DIFF
--- a/include/cuco/operator.hpp
+++ b/include/cuco/operator.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2024, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


sleeepyjack:
The current solution with documenting parameters (`@see`) might not be ideal and I haven't checked how the current approach renders to HTML. However, this is something we can fix once we are ready to build the docs using sphinx. At least the doxygen check doesn't complain for now.

include/cuco/operator.hpp
```diff
@@ -17,65 +17,360 @@
...
+ *
+ * API Signature:
+ * ```cpp
+ * template <typename Value>
+ * __device__ bool insert(Value const& value) noexcept
+ *
+ * template <typename Value>
+ * __device__ bool insert(cooperative_groups::thread_block_tile<cg_size> const& group,
+ *                        Value const& value) noexcept
+ * ```
```
PointKernel, line 62: Based on the offline discussion. An alternative to avoid `@see` is to move parameter docs into the doc cpp code samples.

PointKernel (commented):
One idea to test otherwise looks good to me

PointKernel (approved):
Approving as the alternative didn't work out

PR #669, opened 2025-02-12
Bloom filter optimizations (1/5): Less noisy benchmarks

sleeepyjack:
This PR is part 1/5 of the Bloom filter optimization project and must be merged in the correct order.

This PR introduces two changes that reduce the noise level of the benchmark measurements drastically:

- Generate input data on-the-fly rather than loading the keys from global memory.
- Increase the number of input keys to make the kernels run longer (former runtimes were in the single-digit ms range which was too noisy).


benchmarks/bloom_filter/defaults.hpp
```diff
@@ -30,7 +30,7 @@ using BF_KEY  = nvbench::int64_t;
 using BF_HASH = cuco::xxhash_64<char>;
 using BF_WORD = nvbench::uint32_t;
 
-static constexpr auto BF_N               = 400'000'000;
+static constexpr auto BF_N               = 1'000'000'000;
```
PointKernel, line 33: copyright years to be updated otherwise looks good

PointKernel (approved):

mhaseeb123 (approved):

PR #672, opened 2025-02-12
Bloom filter optimizations (4/5): Implement device bulk add

sleeepyjack:
This PR is part 4/5 of the Bloom filter optimization project and must be merged in the correct order.

This PR introduces the following API:
```cpp
template <class CG, class InputIt>
__device__ void add(CG const& group, InputIt first, InputIt last)
```
, i.e., a device-bulk operation for adding multiple items into the filter using a CG.

Using this approach in the (host) bulk `add` kernel improves performance due to the following reason:
The current kernel uses one CG per input element. Each thread in the CG loads the same key and computes a hash value and target block index. This computation is redundant and leads to increased compute pipeline pressure, which becomes a bottleneck in case the filter is small, i.e., fits into L2$.
In the new kernel, each thread loads one key at a time, computes the hash value and the index, and then shuffles these values to the other threads in the group, which then perform the cooperative `add` operation.

PR #675, opened 2025-02-19
[REVIEW] fixed an index issue in bloom_filter_impl

eleanorTurintech:
There seem to be some small issues with the for-loop indices in bloom filter. This pr aims to fix them.

In particular, there are two issues that this pr fixes:

1. in ```cuco::detail::bloom_filter_impl::add(CG const& group, ProbeKey const& key)```, variable ```i``` in the for-loop doesn't seem to be used for indexing ```word``` and ```atom_word```. This pr updates the index from ```rank``` to ```i```.
2. in ```cuco::detail::bloom_filter_impl::clear(CG const& group)```, the for-loop's terminating condition was not complete. This pr adds ```i < ``` to the condition to complete it.

### DIFF
--- a/include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
+++ b/include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
@@ -97,7 +97,7 @@ class bloom_filter_impl {
   template <class CG>
   __device__ constexpr void clear(CG const& group)
   {
-    for (int i = group.thread_rank(); num_blocks_ * words_per_block; i += group.size()) {
+    for (int i = group.thread_rank(); i < num_blocks_ * words_per_block; i += group.size()) {
       words_[i] = 0;
     }
   }
@@ -152,10 +152,10 @@ class bloom_filter_impl {
 
 #pragma unroll
       for (uint32_t i = rank; i < optimal_num_threads; i += num_threads) {
-        auto const word = policy_.word_pattern(hash_value, rank);
+        auto const word = policy_.word_pattern(hash_value, i);
 
         auto atom_word =
-          cuda::atomic_ref<word_type, thread_scope>{*(words_ + (idx * words_per_block + rank))};
+          cuda::atomic_ref<word_type, thread_scope>{*(words_ + (idx * words_per_block + i))};
         atom_word.fetch_or(word, cuda::memory_order_relaxed);
       }
     }


include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -97,7 +97,7 @@ class bloom_filter_impl {
   template <class CG>
   __device__ constexpr void clear(CG const& group)
   {
-    for (int i = group.thread_rank(); num_blocks_ * words_per_block; i += group.size()) {
+    for (int i = group.thread_rank(); i < num_blocks_ * words_per_block; i += group.size()) {
```
PointKernel, line 102: @sleeepyjack Do we want to keep this CG clear? It doesn't seem to be used.

PointKernel (commented):
We probably need stricter unit tests for the CG add.  

@eleanorTurintech Thanks a lot for catching and fixing those issues!

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
sleeepyjack: Yes, we want to keep this as this is the only way of initializing a shmem filter.

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -152,10 +152,10 @@ class bloom_filter_impl {
 
 #pragma unroll
       for (uint32_t i = rank; i < optimal_num_threads; i += num_threads) {
-        auto const word = policy_.word_pattern(hash_value, rank);
+        auto const word = policy_.word_pattern(hash_value, i);
 
         auto atom_word =
-          cuda::atomic_ref<word_type, thread_scope>{*(words_ + (idx * words_per_block + rank))};
+          cuda::atomic_ref<word_type, thread_scope>{*(words_ + (idx * words_per_block + i))};
```
sleeepyjack: Ha, very good catch! 🎣 The bug would have surfaced if we use a CG that is smaller than the optimal CG size.

sleeepyjack (commented):
Thank you @eleanorTurintech for this contribution!

I reworked the CG `add` in #672 a couple of days ago but haven't merged the changes yet. Let me get that merged first and then we can evaluate which of your changes are still relevant.

sleeepyjack:
@eleanorTurintech #672 has just been merged. The fix for `clear` is definitely still needed but the fix for `add` could be obsolete. I would kindly request your review on the new implementation.

eleanorTurintech:
> @eleanorTurintech #672 has just been merged. The fix for `clear` is definitely still needed but the fix for `add` could be obsolete. I would kindly request your review on the new implementation.

You are right. The ```add``` change is obsolete now. I have synced with the main branch, and only kept the ```clear``` change in this pr. Thanks for your comments!

sleeepyjack (approved):

PR #677, opened 2025-02-20
Update examples to use tranform iterator for insert instead of zip

PointKernel:
The current `static_map` host API example demonstrates the use of a zip iterator to combine two separate iterators for convenience. However, this approach is not ideal, as using a zip iterator for insertion can potentially lead to misaligned memory access, causing insertion hangs. To avoid promoting this practice, this PR updates the example to use a transform iterator instead, eliminating this issue.


examples/static_map/host_bulk_example.cu
```diff
@@ -57,16 +58,21 @@ int main(void)
...
   thrust::sequence(insert_values.begin(), insert_values.end(), 0);
-  auto zipped =
-    thrust::make_zip_iterator(thrust::make_tuple(insert_keys.begin(), insert_values.begin()));
+  // Combine keys and values into pairs {{0,0}, {1,1}, ... {i,i}}
+  auto pairs = thrust::make_transform_iterator(
+    thrust::counting_iterator<std::size_t>{0},
+    cuda::proclaim_return_type<cuco::pair<Key, Value>>(
+      [keys = insert_keys.begin(), values = insert_values.begin()] __device__(auto i) {
+        return cuco::pair<Key, Value>{keys[i], values[i]};
+      }));
```
PointKernel, line 72: We could simplify the logic by using a transform iterator along with a counting iterator, eliminating the need to materialize both keys and values. However, I prefer to retain the materialization since this code also demonstrates how users can combine keys and values on the fly without needing a dedicated pair array.

sleeepyjack: Yep, I totally agree with that.

sleeepyjack (approved):

PR #685, opened 2025-02-27
Add accuracy test for Bloom filter

sleeepyjack:
This PR adds two things:

- A utility function `expected_false_positive_rate` which returns the theoretical false-positive rate of the filter given its parameters
- A unit test that compares the empirical false-positive rate of the Bloom filter with the theoretical value


sleeepyjack:
Superseeded by #808

PR #686, opened 2025-02-27
Fix incorrect use of buffer_size instead of bucket_size in retrieve

PointKernel:
This PR fixes a minor bug in the retrieve function where `buffer_size` was incorrectly used instead of `bucket_size`.


sleeepyjack (approved):
Thanks for fixing my mess :)

PR #690, opened 2025-03-03
Update map insert bench to avoid fully occupied map

PointKernel:
This PR modifies the single map insert benchmark to construct the map on the fly in each loop iteration, preventing insertions into a fully occupied map.

### DIFF
--- a/benchmarks/static_map/insert_bench.cu
+++ b/benchmarks/static_map/insert_bench.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021-2024, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


benchmarks/static_map/insert_bench.cu
```diff
@@ -54,10 +54,19 @@ std::enable_if_t<(sizeof(Key) == sizeof(Value)), void> static_map_insert(
...
+                                cuco::empty_value<Value>{-1},
+                                {},
+                                {},
+                                {},
+                                {},
+                                {},
+                                {launch.get_stream()}};
+    timer.start();
     map.insert_async(pairs.begin(), pairs.end(), {launch.get_stream()});
+    timer.stop();
```
sleeepyjack:
```suggestion
  state.exec(nvbench::exec_tag::timer, [&](nvbench::launch& launch, auto& timer) {
    timer.start();
    map.insert_async(pairs.begin(), pairs.end(), {launch.get_stream()});
    timer.stop();
    map.clear_async({launch.get_stream()});
```

This moves the memory allocation (which is also not stream-ordered, so you'd have to use `exec_tag::sync`; not ideal for accuracy) out of the hot loop.

sleeepyjack (changes requested):

sleeepyjack (approved):

PR #692, opened 2025-03-09
Optimize extent calculation for linear probing by eliminating prime number computation

### LINKED ISSUE #581: [ENHANCEMENT]: Improve `make_window_extent` in the case of linear probing

The current `make_window_extent` function always calculates prime capacity, regardless of the probing scheme used. This approach is suboptimal for linear probing, where the capacity does not need to be a prime number.

If it's linear probing, return the window extent based on CG size and window size without the prime table lookup.

```cpp
template <typename T>
struct is_double_hashing : cuda::std::false_type {};

template <int32_t CGSize, typename Hash1, typename Hash2>
struct is_double_hashing<cuco::double_hashing<CGSize, Hash1, Hash2>> : cuda::std::true_type {};
```
A trait similar to the above idea is needed to dispatch between linear probing and double hashing

_No response_

### Additional context

_No response_

PointKernel:
Closes #581

This PR enhances the extent calculation for linear probing by eliminating the need to compute the prime number for the final extent. It also resolves an embarrassing bug where two prime numbers were missing in the existing prime number array.


include/cuco/detail/prime.hpp
PointKernel, line 1: 151 and 11 were missing in the existing prime array. shame on me :sob:

sleeepyjack: Could also be my fault. IIRC I updated the prime file a while ago. The diff view for this file is buggy because it's so larger. Did you double check that the specified array size mathes the number of items?

sleeepyjack (approved):

include/cuco/detail/prime.hpp
PointKernel: Yes, they match!

Issue #695, opened 2025-04-04
[ENHANCEMENT]: Make cuco code NVRTC-compatible


PointKernel:
### Is your feature request related to a problem? Please describe.

Related to https://github.com/NVIDIA/cccl/issues/4305

Make sure cuco is JIT or NVRTC compatible.

### Describe the solution you'd like

To enable cuco code to be used in JIT kernels, we need to ensure that no thrust algorithms are used in cuco implementations. When available, we will replace the two existing uses of thrust algorithms with their `cuda::std::` equivalents.
https://github.com/NVIDIA/cuCollections/blob/e3aec271627a5d482cf2c720c6e2ec6e15cae06c/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh#L1269
https://github.com/NVIDIA/cuCollections/blob/e3aec271627a5d482cf2c720c6e2ec6e15cae06c/include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh#L1257

As a simpler initial task, we can start by replacing thrust utilities such as `pair`, `tuple`, and `equal_to` with their corresponding `cuda::std::` equivalents, which are already available for use.

sleeepyjack:
How does rapids test for nvrtc compatibility? Can we set up a CI job for this?

lamarrr:
Rapids CUDF has a couple of NVRTC-compatible kernels, we run these kernels as part of our tests via JITIFY.
Yes, you can set up a CI job for it; you'd need to use JITIFY2 or NVRTC directly instead of NVCC.
As long as the tests compile and run successfully, you'd be good.

gengar-maker:
Hi! This ticket is still open. Is there any known problems with using NVRTC for cuco code? Or it's already working fine?

lamarrr:
I haven't tested yet. Perhaps @PointKernel has?

0z5a:
Hi @PointKerne ,l 'd be interested in taking the remaining work here. It looks like some of the original dependencies have already disappeared, but `open_addressing_ref_impl.cuh` still has the device-side `thrust::reduce(thrust::seq, ...)` dependency.

I can first add a minimal NVRTC/JIT compile test for the relevant cuco ref path, audit the remaining Thrust dependencies exposed by it, and then remove or replace the blockers without changing the existing `cuco::pair` alignment semantics.

PointKernel:
Great. @0z5a, assigning this issue to you.

PR #700, opened 2025-04-16
Replace deprecated thrust features

miscco:
We are failing to build cuDF as it breaks due to deprecation warnings


include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -105,7 +105,7 @@ class bloom_filter_impl {
   __host__ constexpr void clear(cuda::stream_ref stream)
   {
     this->clear_async(stream);
-    stream.wait();
+    stream.sync();
```
PointKernel: @miscco Thanks for working on this! This aligns with part of my plan in #695. It looks like the CI is failing due to the use of `stream.sync`. Has `wait` been deprecated in CCCL?

miscco: Yeah it has been deprecated in 3.0 but sync was only added when it was deprecated :(

PointKernel: Got it. We can revisit and adjust this breaking change once `rapids-cmake` updates its CCCL to 3.0. I will keep an eye on this.

miscco: I dropped the sync changes so we can get this merged

PointKernel (approved):
Appreciate your help cleaning up cuco, @miscco!

PR #702, opened 2025-04-27
Use relaxed memory-order in dynamic_map::find

caaatch22:
This PR updates the dynamic kernel code to use relaxed memory ordering for retrieving the matched payload, instead of strict sequential ordering, in order to improve runtime performance.


PointKernel (approved):
Looks good. Thank you @caaatch22, for your contribution to the cuCollections!

PR #706, opened 2025-04-29
Add atomic storage test

PointKernel:
This PR adds a test that mimics a Spark runtime failure scenario not covered by the existing unit tests.


tests/static_set/atomic_storage_test.cu
```diff
@@ -0,0 +1,73 @@
...
+
+class build_fn {
+ public:
+  __device__ __forceinline__ auto operator()(T i) const noexcept { return cuco::pair{_hash(i), i}; }
+
+ private:
+  cuco::default_hash_function<T> _hash{};
+};
+
+// This test exercise is designed to replicate a Spark runtime failure scenario
```
sleeepyjack, line 50: Can we add a link to the specific GH issue for future reference

sleeepyjack (approved):

PR #705, opened 2025-04-28
Upgrade rapids-cmake version to fetch CCCL-2.8

PointKernel:
This PR upgrades the rapids-cmake version to 25.06 to fetch CCCL-2.8.

PR #707, opened 2025-04-30
Update RAPIDS, CUDA versions

bdice:
Depends on https://github.com/NVIDIA/cuCollections/pull/705.

This PR adds a script `ci/update_rapids_version.sh` that can be used to update all references to RAPIDS versions in this repository. Call it like `./ci/update_rapids_version.sh 25.06.00` to update devcontainers, rapids-cmake, and other locations using RAPIDS versioning. (A similar script exists for CCCL, which was used as a basis for this script.)

I also updated the maximum CUDA version to 12.8, to align better with the latest RAPIDS support matrix.


PointKernel (approved):
@bdice Thanks for the help!

PR #708, opened 2025-05-01
Update cuco implementations to use cuda::std utilities when appropriate

PointKernel:
This PR updates cuco implementations to use `cuda::std` utilities in place of `std` in device code and to replace `thrust` utilities wherever possible, as they are being deprecated in CCCL.


include/cuco/detail/open_addressing/functors.cuh
```diff
@@ -18,6 +18,8 @@
 #include <cuco/detail/bitwise_compare.cuh>
 #include <cuco/detail/pair/traits.hpp>
 
+#include <thrust/tuple.h>
```
bdice, line 21: I think `cuda::std::tuple` should be preferred here? Please try it. It's also okay to defer this to another PR.

bdice (approved):
Generally LGTM -- ideally we would use `cuda::std::tuple` in the places where `thrust::tuple` is being used.

include/cuco/detail/open_addressing/functors.cuh
PointKernel: Right, the issue is from the outdated CCCL version, where `zip_iterator` is incompatible with `cuda::std::tuple`: https://godbolt.org/z/jnKafGPf4. Currently, cuco is still pulling in `rapids-cmake` branch-25.02, which brings in CCCL 2.7. Once we update to a newer version of `rapids-cmake`, CCCL will be upgraded accordingly, and we’ll be able to fully replace `thrust::tuple` with `cuda::std::tuple` in cuco.

sleeepyjack (approved):

PR #709, opened 2025-05-03
Add retrieve_all for dynamic_map

caaatch22:
This PR adds host-bulk `retrieve_all` for `dynamic_map`.

### DIFF
--- a/include/cuco/detail/dynamic_map.inl
+++ b/include/cuco/detail/dynamic_map.inl
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -235,6 +235,52 @@ void dynamic_map<Key, Value, Scope, Allocator>::find(InputIt first,
   detail::find<block_size, tile_size, Value><<<grid_size, block_size, 0, stream>>>(
     first, last, output_begin, submap_views_.data().get(), submaps_.size(), hash, key_equal);
   CUCO_CUDA_TRY(cudaDeviceSynchronize());
+}
+
+template <typename Key, typename Value, cuda::thread_scope Scope, typename Allocator>
+template <typename KeyOut, typename ValueOut>
+std::pair<KeyOut, ValueOut> dynamic_map<Key, Value, Scope, Allocator>::retrieve_all(
+  KeyOut keys_out, ValueOut values_out, cudaStream_t stream) const
+{
+  auto constexpr block_size = 128;
+  auto constexpr stride     = 1;
+
+  auto const capacity       = get_capacity();
+  auto grid_size            = (capacity + stride * block_size - 1) / (stride * block_size);
+
+  std::vector<size_t> submap_cap_prefix(submaps_.size());
+  std::inclusive_scan(
+    submaps_.begin(),
+    submaps_.end(),
+    submap_cap_prefix.begin(),
+    [](auto const& sum, auto const& submap) { return sum + submap->get_capacity(); },
+    (size_t)0);
+  thrust::device_vector<size_t> submap_cap_prefix_d(submap_cap_prefix);
+
+  using temp_allocator_type =
+    typename std::allocator_traits<Allocator>::template rebind_alloc<char>;
+  auto temp_allocator = temp_allocator_type{alloc_};
+  auto d_num_out =
+    reinterpret_cast<unsigned long long*>(std::allocator_traits<temp_allocator_type>::allocate(
+      temp_allocator, sizeof(unsigned long long)));
+  CUCO_CUDA_TRY(cudaMemsetAsync(d_num_out, 0, sizeof(unsigned long long), stream));
+
+  detail::retrieve_all<block_size>
+    <<<grid_size, block_size, 0, stream>>>(keys_out,
+                                           values_out,
+                                           submap_views_.data().get(),
+                                           submaps_.size(),
+                                           capacity,
+                                           d_num_out,
+                                           submap_cap_prefix_d.data().get());
+
+  size_t h_num_out;
+  CUCO_CUDA_TRY(
+    cudaMemcpyAsync(&h_num_out, d_num_out, sizeof(std::size_t), cudaMemcpyDeviceToHost, stream));
+  CUCO_CUDA_TRY(cudaStreamSynchronize(stream));
+  std::allocator_traits<temp_allocator_type>::deallocate(
+    temp_allocator, reinterpret_cast<char*>(d_num_out), sizeof(unsigned long long));
+  return {keys_out + h_num_out, values_out + h_num_out};
 }
 
 template <typename Key, typename Value, cuda::thread_scope Scope, typename Allocator>
--- a/include/cuco/detail/dynamic_map.inl
+++ b/include/cuco/detail/dynamic_map.inl
@@ -253,7 +253,8 @@ std::pair<KeyOut, ValueOut> dynamic_map<Key, Value, Scope, Allocator>::retrieve_
     submaps_.begin(),
     submaps_.end(),
     submap_cap_prefix.begin(),
-    [](auto const& sum, auto const& submap) { return sum + submap->get_capacity(); },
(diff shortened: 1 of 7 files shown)


include/cuco/detail/dynamic_map.inl
```diff
@@ -237,6 +237,52 @@ void dynamic_map<Key, Value, Scope, Allocator>::find(InputIt first,
...
+  auto const capacity = get_capacity();
+  auto grid_size      = (capacity + stride * block_size - 1) / (stride * block_size);
+
+  std::vector<size_t> submap_cap_prefix(submaps_.size());
+  std::inclusive_scan(
+    submaps_.begin(),
+    submaps_.end(),
+    submap_cap_prefix.begin(),
+    [](auto const& sum, auto const& submap) { return sum + submap->get_capacity(); },
+    (size_t)0);
```
PointKernel:
```suggestion
    size_t{0};
```
nit: use ctor instead of c-style casting in modern c++

include/cuco/detail/dynamic_map.inl
```diff
@@ -237,6 +237,52 @@ void dynamic_map<Key, Value, Scope, Allocator>::find(InputIt first,
...
+  thrust::device_vector<size_t> submap_cap_prefix_d(submap_cap_prefix);
+
+  using temp_allocator_type =
+    typename std::allocator_traits<Allocator>::template rebind_alloc<char>;
+  auto temp_allocator = temp_allocator_type{alloc_};
+  auto d_num_out =
+    reinterpret_cast<unsigned long long*>(std::allocator_traits<temp_allocator_type>::allocate(
+      temp_allocator, sizeof(unsigned long long)));
+  CUCO_CUDA_TRY(cudaMemsetAsync(d_num_out, 0, sizeof(unsigned long long), stream));
```
PointKernel: we have a [counter_storage](https://github.com/NVIDIA/cuCollections/blob/3e04706debe7ce8ef9f42f0ded26b62d6dfd1be1/include/cuco/detail/storage/counter_storage.cuh) utility dedicated to help in this situation, e.g. https://github.com/NVIDIA/cuCollections/blob/3e04706debe7ce8ef9f42f0ded26b62d6dfd1be1/include/cuco/detail/open_addressing/open_addressing_impl.cuh#L340-L342

include/cuco/detail/dynamic_map_kernels.cuh
```diff
@@ -487,6 +487,85 @@ CUCO_KERNEL void find(InputIt first,
...
+  using BlockScan = cub::BlockScan<unsigned int, block_size>;
+
+  __shared__ typename BlockScan::TempStorage scan_temp_storage;
+  __shared__ unsigned int block_base;
+
+  auto tid                       = blockDim.x * blockIdx.x + threadIdx.x;
+  auto const empty_key_sentinel  = submap_views[0].get_empty_key_sentinel();
+  auto const erased_key_sentinel = submap_views[0].get_erased_key_sentinel();
+
+  while (tid < capacity) {
```
PointKernel:
```suggestion
  while ((tid - threadIdx.x) < capacity) {
```
We need to ensure that the entire thread block follows the same execution path; otherwise, the `__syncthreads()` at line 558 could fail if some threads don't enter the while loop. The handling around `tid` also needs to be updated accordingly to reflect this change.

tests/dynamic_map/retrieve_all_test.cu
```diff
@@ -0,0 +1,123 @@
...
+  SECTION("retrieve_all after insert")
+  {
+    thrust::device_vector<Key> d_keys(num_keys);
+    thrust::device_vector<Value> d_values(num_keys);
+
+    thrust::sequence(thrust::device, d_keys.begin(), d_keys.end(), 0);
+    thrust::sequence(thrust::device, d_values.begin(), d_values.end(), 0);
+
+    auto pairs_begin =
+      thrust::make_zip_iterator(thrust::make_tuple(d_keys.begin(), d_values.begin()));
```
PointKernel: Using a zip iterator to generate keys has known data alignment issues and can potentially lead to runtime hangs, as noted in https://github.com/NVIDIA/cuCollections/pull/677. Ideally, we should always use a transform iterator instead.

PointKernel (commented):
@caaatch22 Thanks for the PR! The changes look great overall, just a few minor issues need to be addressed.

As a side note, we haven’t been actively maintaining the `dynamic_map` data structure, as we initially thought there were no active users. The current setup is somewhat outdated, and we’ve only recently started transitioning from `cuco::dynamic_map` to `cuco::experimental::dynamic_map`. However, this process has been progressing slowly due to the lack of active usage.

That said, we're happy to prioritize the migration to `cuco::experimental::dynamic_map`, aiming for more organized and unified interfaces. If you don’t mind, could you share more details about your use case? For example, are there any features available in the `static_map` that you find missing in `dynamic_map`, or are there specific performance concerns you’d like to address? Please let us know.

caaatch22:
> @caaatch22 Thanks for the PR! The changes look great overall, just a few minor issues need to be addressed.
> 
> As a side note, we haven’t been actively maintaining the `dynamic_map` data structure, as we initially thought there were no active users. The current setup is somewhat outdated, and we’ve only recently started transitioning from `cuco::dynamic_map` to `cuco::experimental::dynamic_map`. However, this process has been progressing slowly due to the lack of active usage.
> 
> That said, we're happy to prioritize the migration to `cuco::experimental::dynamic_map`, aiming for more organized and unified interfaces. If you don’t mind, could you share more details about your use case? For example, are there any features available in the `static_map` that you find missing in `dynamic_map`, or are there specific performance concerns you’d like to address? Please let us know.

Thank you very much for your thorough review! I have already made some adjustments based on your suggestions.

It's really great to hear that you're open to supporting this use case and the migration to `cuco::experimental::dynamic_map`. We are currently working on a recommendation training framework where we need a scalable GPU-based embedding solution, similar to hugeCTR. And the `retrieve_all` method is used for saving checkpoints.

For our use case, device-side APIs for dynamic_map would be nice to have. We are trying to implement a function like `find_or_insert`. With static_map, we can implement that by using device-side  `insert_and_find` API. With `dynamic_map`, achieving similar functionality is not as trivial. My solution now is to launch multiple kernels to do that. I’d love to hear if you have a better way or any insights to handle this.

We also observed that the performance of insert, find, and contain operations drops quite a lot when the number of submaps is greater than 1,  improving this issue may require non-trivial work.

Thank you again for your incredible work on this library. I’m really excited about the potential of cuco::experimental::dynamic_map, and I’d be happy to contribute to its development.

include/cuco/detail/dynamic_map.inl
caaatch22: Yeah, using detail::counter_storage would be much more elegant, but I found it may not be consistent with current interface. The stream is a `cuda::stream_ref` in counter_storage, and `cudaStream_t` in cuco::dynamic_map.
Perhaps we can use `counter_storage` in `cuco::experimental::dynamic_map`  (if we add a `retrieve_all` method there)?

PointKernel: Good point. Let me resolve this.

PointKernel (approved):
@caaatch22 Thanks again for your contribution.

I made a few small changes directly and also updated the PR title and description accordingly, since we do squash merges and they will be included in the commit message.

Another idea I wanted to try is to use `cub::DeviceSelect::If` directly similar to what we do in the new OA base class: https://github.com/NVIDIA/cuCollections/blob/3e04706debe7ce8ef9f42f0ded26b62d6dfd1be1/include/cuco/detail/open_addressing/open_addressing_impl.cuh#L772-L832

This can help avoid writing kernels on our own and may bring some performance improvement.

PointKernel:
>  My solution now is to launch multiple kernels to do that. I’d love to hear if you have a better way or any insights to handle this.

Having multiple passes may not be a bad idea, as we often run into register issues in such situations. Using multiple kernels can actually reduce register pressure and improve runtime performance.

> We also observed that the performance of insert, find, and contain operations drops quite a lot when the number of submaps is greater than 1, improving this issue may require non-trivial work.

Do you happen to know whether the slowdown is caused by data allocation for submaps or by the insert/query process? If it's due to extra memory allocation, a device memory pool like https://github.com/rapidsai/rmm could help significantly. If it's the latter, exposing a `rehash` API may help improve the situation.

Feel free to open GitHub issues or start a discussion if needed. We're happy to assist and improve cuco based on your feedback.

PR #710, opened 2025-05-05
Use `cuda::std::tuple` to replace `thrust::tuple`

PointKernel:
This PR replaces `thrust::tuple` with `cuda::std::tuple`, as the former is being phased out in CCCL.


sleeepyjack (approved):
LGTM
Just to make sure: Did you ran a few benchmarks to verify we don't introduce any unexpected regression?

PR #719, opened 2025-05-15
Add proper handling for invalid desired load factor

PointKernel:
This PR fixes an error handling bug when users provide an invalid load factor. It helps unblock the disabled unit tests in cudf hash join and distinct join.


sleeepyjack (approved):
Copyright year in `tests/test_utils.hpp` and `include/cuco/extent.cuh` need to be updated.

Apart from that LGTM

PR #733, opened 2025-06-23
Several fixes to make cuCo compile with LLVM:

vitor1001:
- add `template` keyword to help parsing.
- do not mark as `constexpr` functions that cannot be evaluated at compile-time.


PointKernel (approved):
@vitor1001 Looks good. Thanks for the contribution!

PR #736, opened 2025-06-27
Replace deprecated call to wait with sync

vyasr:
See https://github.com/NVIDIA/cccl/blob/68583867aa879f9a4e0c955f466289585bbaa541/libcudacxx/include/cuda/__stream/stream_ref.h#L106.


PointKernel:
Thanks, I'm aware of this deprecation. 

`stream_ref::sync` will get in with CCCL 3.1 but `rapids-cmake` is still fetching CCCL 3.0.

vyasr:
rapids-cmake is currently fetching commit dc1db8aefcfe52c66c26c6038fef58da37a46ae9, which is tag [v3.0.0-rc5](https://github.com/NVIDIA/cccl/releases/tag/v3.0.0-rc5). It looks like you're right that this change isn't making it in to the 3.0.0* line though since I also don't see it on [the latest rc](https://github.com/NVIDIA/cccl/blob/v3.0.0-rc9/cudax/include/cuda/experimental/__stream/stream_ref.cuh#L105). Good to know, thanks.

include/cuco/detail/storage/counter_storage.cuh
```diff
@@ -96,7 +96,7 @@ class counter_storage : public storage_base<cuco::extent<SizeType, 1>> {
     CUCO_CUDA_TRY(cudaMallocHost(&h_count, sizeof(size_type)));
     CUCO_CUDA_TRY(cudaMemcpyAsync(
       h_count, this->data(), sizeof(size_type), cudaMemcpyDeviceToHost, stream.get()));
-    stream.wait();
+    stream.sync();
```
bdice, line 99: We need a way to make cuCollections compatible with CCCL 3.0 and 3.1 until we can migrate all of RAPIDS to CCCL 3.1.

For now, I need this change to unblock cuDF builds (which fail with deprecation warnings as errors).

```suggestion
#if CCCL_MAJOR_VERSION > 3 || (CCCL_MAJOR_VERSION == 3 && CCCL_MINOR_VERSION >= 1)
    stream.sync();
#else   // ^^^ CCCL >= 3.1 ^^^ / vvv CCCL < 3.1 vvv
    stream.wait();
#endif  // CCCL < 3.1
```

bdice (approved):
I verified that libcudf builds with CCCL 3.1 and the deprecation warnings go away when these changes are included. Thank you!

PointKernel (approved):

PR #737, opened 2025-06-28
Fix primitive type namespace

PointKernel:
This PR updates the relevant code to use `std::` for primitive types in host-only code, `cuda::std::` for host-device or device-only code, and raw integers or booleans for non-type template parameters. These changes also improve compatibility with NVRTC.


include/cuco/detail/dynamic_map.inl
```diff
@@ -77,42 +79,52 @@ dynamic_map<Key, Value, Scope, Allocator>::dynamic_map(std::size_t initial_capac
...
 {
-  int64_t num_elements_remaining = n;
-  uint32_t submap_idx            = 0;
-  while (num_elements_remaining > 0) {
-    std::size_t submap_capacity;
-
-    // if the submap already exists
-    if (submap_idx < submaps_.size()) {
-      submap_capacity = submaps_[submap_idx]->get_capacity();
+  // Calculate current total available capacity across all submaps
```
PointKernel, line 82: The legacy implementation is poorly designed, relying on `num_elements_remaining` being a signed integer to terminate the loop when it becomes negative. This PR updates the logic to remove that dependency, ensuring it works regardless of the integer type.

include/cuco/operator.hpp
```diff
@@ -290,7 +290,7 @@ struct find_tag {
  *
  * API Signature:
  * ```cpp
- * template <int32_t BlockSize,
+ * template <::int BlockSize,
```
sleeepyjack: Why is the `::` needed?

include/cuco/detail/utility/cuda.hpp
```diff
@@ -58,18 +57,18 @@ constexpr auto grid_size(index_type num,
...
+  std::int32_t device = 0;
   CUCO_CUDA_TRY(cudaGetDevice(&device));
 
-  int32_t num_multiprocessors = -1;
+  std::int32_t num_multiprocessors = -1;
   CUCO_CUDA_TRY(
     cudaDeviceGetAttribute(&num_multiprocessors, cudaDevAttrMultiProcessorCount, device));
 
-  int32_t max_active_blocks_per_multiprocessor{};
+  std::int32_t max_active_blocks_per_multiprocessor{};
```
sleeepyjack: The CUDA runtime API always return builtins aka `int`.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -834,7 +835,7 @@ class open_addressing_impl {
     using temp_allocator_type =
       typename std::allocator_traits<allocator_type>::template rebind_alloc<char>;
 
-    cuco::detail::index_type constexpr stride = std::numeric_limits<int32_t>::max();
+    cuco::detail::index_type constexpr stride = std::numeric_limits<std::int32_t>::max();
```
sleeepyjack, line 838: Stride is also used on the GPU, right? Should it be `cuda::std::` instead, then?

include/cuco/detail/utility/cuda.hpp
PointKernel: Yeah, that's the part I'm not sure either. The same for all the aliases of template parameters, e.g. 
```cpp
template<int BlockSize>
struct toto {
  static constexpr int block_size = BlockSize;
};
```
Should `block_size` be `int` or `cuda::std::int32_t`? On one hand, `int` keeps things consistent, but on the other, `cuda::std::int32_t` offers better namespace control.

include/cuco/operator.hpp
PointKernel: me just quoting @srinivasyadav18 "it's fully qualified". Ideally, this would be the right approach, but I’m not sure we should enforce it at this point.

include/cuco/detail/open_addressing/open_addressing_impl.cuh
PointKernel: `stride` is host-only

include/cuco/detail/utility/cuda.hpp
sleeepyjack: Can we just use `auto` in these cases so we generically adapt whatever type is given to us by the CUDA API? It should also work for your `toto` example.

PointKernel: You are right, let me move to raw `int` types in both cases for consistency

sleeepyjack (approved):

PR #704, opened 2025-04-28
cmake option to build.sh

esoha-nvidia:
In case the default `cmake` isn't appropriate for building, let the user specify a `cmake` binary.

PR #741, opened 2025-07-05
Support for 32- and 64-bit cuco::experimental::roaring_bitmap lookups

### LINKED ISSUE #725: Roaring bitmaps discussion and scoping
This issue is intended to discuss the feasability and scope of a potential `cuco::roaring_bitmap` implementation.

## Background
Roaring Bitmaps \[[paper](https://arxiv.org/pdf/1603.06549)\] are a compressed bitmap format designed to efficiently represent sets of 32- or 64-bit unsigned integers. They aim to outperform traditional methods such as hash- or tree-based set data structures, bitset, and run-length encoding (RLE) methods like WAH, EWAH, and Concise by offering superior speed and compression, especially on unsorted or sparse data.

For simplicity reasons I will focus on the 32-bit implementation in the following paragraphs.

The concept is drivenby the following observations comparing the abovementioned set implementations:
- Bitsets are very efficient in case the set has a high density given its domain. In other words, if the set of integers covers a significant portion of the 32-bit integer domain \[0, 2^32-1\]. Operations such as membership tests, set unions, or set intersections can be implemented using efficient bitwise operations.
- In case the set density is low, the bitset will be mostly empty which makes it storage inefficient. Instead, sparse sets can be better represented using dense data structures such as hash sets, trees, or sorted arrays. Roaring chooses the latter for this situation. Set operations over sorted arrays require more irregular algorithms such as binary search or merge steps.

A roaring bitmap mixes both the bitset and sorted array concept to get the "best of both worlds" as follows:
For a given set of input integers, we first partition it into 65k buckets/containers based on the 16 MSB bits of the elements.
(img1_placeholder, GH has some issues with image upload right now. I will try again later)
Each container represents either a sorted array or a bitset that stores the 16 LSB bits of each assigned element. 
(img2_placeholder)
Combining the container index with the 16bit elements stored in it reconstructs the original set members.
A sorted array container is a dynamically resizeable data structure such as a `std::vector` and can hold up-to 4096 elements (8KiB). When this limit is exceeded, the array gets converted into a bitmap (also 8KiB to cover the 16-bit domain). It becomes directly clear that in case the container holds less than 4096 elements, storing the subset as a compact array has a better storage efficiency than the bitset equivalent.
As a final optimization, roaring offers an opt-in step called `runOptimize()` which converts containers that contain large runs of consecutive elements into an RLE representation for even better storage efficiency and subsequently makes the entire data structure immutable.

## Discussion
The concept of a roaring bitmap fits perfectly into the scope of cuco. However, I see some challenges regarding implementation details.
### Dynamic allocation issue
Roaring bitmaps heavily rely on fine-grained dynamic memory allocation which is unfortunately not easily portable to the GPU.
In fact, this feature is what makes roaring so powerful by keeping the memory footprint small during construction and transformation.
If we want to keep this feature, we would need a memory pool such as RMM to back the allocations, as well as extensive synchronization mechanisms, e.g., we need to block concurrent threads accessing one container while it is being resized.
One could argue that relying on a memory pool that pre-allocates large chunks of memory does not help with storage efficiency.
We do not plan to add RMM or any other memory pool implementation to cuco in the foreseeable future.

### Proposed Workaround
Instead of starting each container as a dynamically growing array, we could instead use bitsets during construction and then convert them into sorted arrays/RLE containers on-demand, similar to the aforementioned `runOptimize()` function, making the data structure immutable.

I have the following workflow in mind:
1. Sort the input elements, e.g., `cub::DeviceRadixSort`.
2. Count the number of elements assigned to each container. (can also be done through some device-level `cub` or `thrust` primitives)
3. For each non-zero count check if a bitmap for this container has already been allocated. If not allocate it. (`cudaMallocAsync` or similar)
4. Merge the input elements into their assigned bitset using, e.g., `atomicOr`. (custom kernel)

After all elements have been added into the roaring bitmap, we can then compress the data structure by calling `runOptimize()`or whatever we want to name it.

There is one problem with this approach. The domain of 32-bit integers can be represented using a 512MiB bitset. So even in the worst case (only one element per container), we can still fit the uncompressed data structure in-memory. If we switch to 64-bit integers, however, then this worst-case memory footprint explodes to 2 exabyte if my napkin math is correct.
## Questions
- What's the target application for this data structure in RAPIDS/Spark? Are we primarily interested in compression ratio, e.g., for better network utilization or is it about matching existing CPU implementations in Spark?
- Does it have to be a roaring bitmap or can we emulate it with say a decently sized `cuco::bloom_filter` which will likely give you an even better compression ratio at a moderate false-positive rate?

## Resources
- [arxiv paper](https://arxiv.org/pdf/1603.06549)
- screeshots taken from [this video](https://www.youtube.com/watch?v=KmL6Z8iFves)

sleeepyjack:
Closes #725


benchmarks/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ function(ConfigureBench BENCH_NAME)
     target_include_directories(${BENCH_NAME} PRIVATE
                                              "${CMAKE_CURRENT_SOURCE_DIR}")
     target_compile_options(${BENCH_NAME} PRIVATE --compiler-options=-Wall --compiler-options=-Wextra
-      --compiler-options=-Werror -Wno-deprecated-gpu-targets --expt-extended-lambda)
+      --compiler-options=-Werror -Wno-deprecated-gpu-targets --expt-extended-lambda -lineinfo)
```
PointKernel, line 31: #it's good to have but probably not to be added on our end

ci/build.sh
```diff
@@ -126,6 +131,7 @@ while [ "${#args[@]}" -ne 0 ]; do
     --arch) CUDA_ARCHS="${args[1]}";    args=("${args[@]:2}");;
     --std)  CXX_STANDARD="${args[1]}";  args=("${args[@]:2}");;
     -v | -verbose | --verbose) VERBOSE=1; args=("${args[@]:1}");;
+    --) EXTRA_CMAKE_OPTIONS+=("${args[@]:1}"); break;;
```
PointKernel: why do we need this change?

include/cuco/roaring_bitmap.cuh
```diff
@@ -0,0 +1,161 @@
...
+  ~roaring_bitmap() = default;  ///< Destructor
+
+  /**
+   * @brief Bulk membership query for keys in `[first, last)`.
+   *
+   * @note This function synchronizes the given stream. For asynchronous execution use
+   *       `contains_async`.
+   *
+   * @tparam InputIt  Device-accessible random access input iterator of keys convertible to `T`
+   * @tparam OutputIt Device-accessible random access output iterator to `bool`
```
PointKernel:
```suggestion
   * @tparam OutputIt Device-accessible random access output iterator whose `value_type` is constructible from `bool`
```

include/cuco/roaring_bitmap.cuh
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * @param first Beginning of the sequence of keys
+   * @param last  End of the sequence of keys
+   * @param contained Output iterator where results are written; `true` iff the corresponding key
```
PointKernel, line 82:
```suggestion
   * @param contained Output iterator where results are written; `true` if the corresponding key
```
probably worth asking cursor to do a thorough typo check

include/cuco/roaring_bitmap.cuh
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * @brief Asynchronously performs a bulk membership query for keys in `[first, last)`.
+   *
+   * @tparam InputIt  Device-accessible random access input iterator of keys convertible to `T`
+   * @tparam OutputIt Device-accessible random access output iterator to `bool`
+   *
+   * @param first Beginning of the sequence of keys
+   * @param last  End of the sequence of keys
+   * @param contained Output iterator where results are written; `true` iff the corresponding key
```
PointKernel, line 100:
```suggestion
   * @param contained Output iterator where results are written; `true` if the corresponding key
```

include/cuco/roaring_bitmap.cuh
```diff
@@ -0,0 +1,161 @@
...
+   * @brief Number of keys stored in the bitmap.
+   *
+   * @return Count of keys in the bitmap
+   */
+  [[nodiscard]] cuda::std::size_t size() const noexcept;
+
+  /**
+   * @brief Checks whether the bitmap contains no keys.
+   *
+   * @return `true` iff `size() == 0`
```
PointKernel, line 120:
```suggestion
   * @return `true` if `size() == 0`
```

include/cuco/roaring_bitmap_ref.cuh
```diff
@@ -0,0 +1,150 @@
...
+   *
+   * @note This function synchronizes the given stream. For asynchronous execution use
+   *       `contains_async`.
+   *
+   * @tparam InputIt  Device-accessible random access input iterator of keys convertible to `T`
+   * @tparam OutputIt Device-accessible random access output iterator to `bool`
+   *
+   * @param first Beginning of the sequence of keys
+   * @param last  End of the sequence of keys
+   * @param contained Output iterator where results are written; `true` iff the corresponding key
```
PointKernel, line 79:
```suggestion
   * @param contained Output iterator where results are written; `true` if the corresponding key
```

include/cuco/roaring_bitmap_ref.cuh
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * @brief Asynchronously performs a bulk membership query for keys in `[first, last)`.
+   *
+   * @tparam InputIt  Device-accessible random access input iterator of keys convertible to `T`
+   * @tparam OutputIt Device-accessible random access output iterator to `bool`
+   *
+   * @param first Beginning of the sequence of keys
+   * @param last  End of the sequence of keys
+   * @param contained Output iterator where results are written; `true` iff the corresponding key
```
PointKernel, line 97:
```suggestion
   * @param contained Output iterator where results are written; `true` if the corresponding key
```

include/cuco/roaring_bitmap_ref.cuh
```diff
@@ -0,0 +1,150 @@
...
+                               InputIt last,
+                               OutputIt contained,
+                               cuda::stream_ref stream = {}) const noexcept;
+
+  /**
+   * @brief Device-side membership query for a single key.
+   *
+   * @param value Key to test for membership
+   *
+   * @return `true` iff `value` is contained in the bitmap
```
PointKernel, line 112:
```suggestion
   * @return `true` if `value` is contained in the bitmap
```

include/cuco/roaring_bitmap_ref.cuh
```diff
@@ -0,0 +1,150 @@
...
+   * @brief Number of keys stored in the bitmap.
+   *
+   * @return Count of keys in the bitmap
+   */
+  [[nodiscard]] __host__ __device__ cuda::std::size_t size() const noexcept;
+
+  /**
+   * @brief Checks whether the bitmap contains no keys.
+   *
+   * @return `true` iff `size() == 0`
```
PointKernel, line 126:
```suggestion
   * @return `true` if `size() == 0`
```

include/cuco/roaring_bitmap.cuh
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * The returned reference type provides device functions such as `contains(T)` for per-thread
+   * membership testing.
+   *
+   * @return Non-owning reference to the underlying bitmap
+   */
+  [[nodiscard]] ref_type ref() const noexcept;
+
+ private:
+  storage_type storage_;
```
PointKernel:
```suggestion
  storage_type storage_; ///< Storage type
```
suprisingly this didn't fail the doc check

include/cuco/detail/roaring_bitmap/util.cuh
```diff
@@ -0,0 +1,226 @@
...
+
+template <class T>
+struct roaring_bitmap_metadata {
+  static_assert(cuco::dependent_false<T>, "T must be either uint32_t or uint64_t");
+};
+
+template <>
+struct roaring_bitmap_metadata<cuda::std::uint32_t> {
+  cuda::std::size_t size_bytes             = 0;
+  cuda::std::uint32_t num_keys             = 0;
```
PointKernel: is the `uint32_t` choice from the spec?

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
```diff
@@ -0,0 +1,232 @@
...
+    CUCO_CUDA_TRY(cudaMemcpyAsync(
+      buckets_.get(),
+      buckets_h_.data(),
+      metadata_.num_buckets * sizeof(cuda::std::pair<cuda::std::uint32_t, bucket_ref_type>),
+      cudaMemcpyHostToDevice,
+      stream.get()));
+    // stream.wait();
+    // clear intermediate data
+    // bucket_metadata.clear();
+    // buckets_h.clear();
```
PointKernel:
```suggestion
```

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
```diff
@@ -0,0 +1,232 @@
...
+  {
+    CUCO_CUDA_TRY(cudaMemcpyAsync(
+      data_.get(), bitmap, metadata_.size_bytes, cudaMemcpyHostToDevice, stream.get()));
+  }
+
+  ref_type ref() const noexcept { return ref_; }
+
+ private:
+  allocator_type allocator_;
+  typename ref_type::metadata_type metadata_;
```
PointKernel, line 155: nit: we should have a `metadata_type` alias in the current class

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
```diff
@@ -0,0 +1,232 @@
...
+
+  __host__ __device__ cuda::std::byte const* container_offsets() const noexcept
+  {
+    return container_offsets_;
+  }
+
+ private:
+  metadata_type metadata_;
+  cuda::std::byte const* data_;
+  cuda::std::uint8_t const* run_container_bitmap_;
```
PointKernel: what does `run_container_bitmap` do

include/cuco/detail/roaring_bitmap/roaring_bitmap_ref.inl
```diff
@@ -0,0 +1,90 @@
...
+
+template <class T>
+__host__ __device__ roaring_bitmap_ref<T>::roaring_bitmap_ref(storage_ref_type const& storage_ref)
+  : impl_{storage_ref}
+{
+}
+
+template <class T>
+template <class U /* = T */,
+          class /* = cuda::std::enable_if_t<cuda::std::is_same_v<U, cuda::std::uint32_t>> */>
```
PointKernel, line 35: is this some cleanup leftover?

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
```diff
@@ -0,0 +1,372 @@
...
+      thrust::fill(
+        nosync_exec_policy, contained, contained + cuda::std::distance(first, last), false);
+    } else {
+      thrust::transform(nosync_exec_policy,
+                        first,
+                        last,
+                        contained,
+                        cuda::proclaim_return_type<bool>(
+                          [*this] __device__(auto key) { return this->contains(key); }));
+    }
```
PointKernel: As discussed offline, we should use cub algorithms here as thrust doesn't provide control for potential temporary data allocation/deallocation

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
```diff
@@ -0,0 +1,372 @@
...
+    } else {
+      cuda::std::uint32_t card;
+      if constexpr (Aligned) {
+        card = 1u + aligned_load<cuda::std::uint16_t>(
+                      storage_ref_.key_cards() + (index * 2 + 1) * sizeof(cuda::std::uint16_t));
+      } else {
+        card = 1u + misaligned_load<cuda::std::uint16_t>(
+                      storage_ref_.key_cards() + (index * 2 + 1) * sizeof(cuda::std::uint16_t));
+      }
+      if (card <= 4096) {
```
PointKernel: let's get rid of magic number and give it a name

include/cuco/detail/roaring_bitmap/util.cuh
```diff
@@ -0,0 +1,226 @@
...
+    constexpr cuda::std::uint32_t serial_cookie                 = 12347;
+    // constexpr cuda::std::uint32_t frozen_cookie                 = 13766; // not implemented
+    constexpr cuda::std::int32_t no_offset_threshold = 4;
+
+    cuda::std::byte const* buf = bitmap;
+
+    cuda::std::uint32_t cookie;
+    cuda::std::memcpy(&cookie, buf, sizeof(cuda::std::uint32_t));
+    buf += sizeof(cuda::std::uint32_t);
+    if ((cookie & 0xFFFF) != serial_cookie && cookie != serial_cookie_no_runcontainer) {
```
PointKernel: maybe name `0xFFFF` as `mask`

PointKernel (commented):
First pass

The logic is straightforward, so there’s not much to review from a technical standpoint.

ci/build.sh
sleeepyjack: This allows for passing arbitrary cmake build options through our `ci/build.sh` script. For instance, this PR introduces the `-DCUCO_DOWNLOAD_ROARING_DATA` option, Which I can now pass as `ci/build.sh — -DCUCO_DOWNLOAD_ROARING_DATA=OFF`.

include/cuco/detail/roaring_bitmap/roaring_bitmap_ref.inl
sleeepyjack: It's a hint mirroring what is there in the definition of the function to help understand why these tparams are there. In this case it means this function is only enabled for the 32-bit roaring bitmap format.

ci/build.sh
PointKernel: So this can replace https://github.com/NVIDIA/cuCollections/pull/704?

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
sleeepyjack: There are three types of containers in the bitmap. (1) array containers, (2) bitset containers, and (3) "run containers" aka array that encode consecutive sequences of elements in RLE format. To distinguish the first two types we can look at the cardinality (if it's >4096 then it is stored as a bitset, otherwise it is an array container). We cannot apply this trick for run containers, thus we have to store a bitset with one bit per container, indicating if it's a run container or not.

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
sleeepyjack: Does this imply we have to move it to public namespace?

include/cuco/detail/roaring_bitmap/util.cuh
sleeepyjack: Thinking about this a bit, a 32-bit roaring bitmap can only store unique `uint32_t` items. Thus, the upper-bound of the stored keys is `ÙINT32_MAX+1` (including storing `0`). So I might have to change that to `size_t`...

include/cuco/detail/roaring_bitmap/util.cuh
sleeepyjack: Yep, `cookie_mask`

include/cuco/roaring_bitmap.cuh
sleeepyjack: Discussed offline

benchmarks/CMakeLists.txt
sleeepyjack: I would vote for keeping it. It's super helpful during development, and adds only minor binary size overhead, which isn't included in production code as cuco benchmarks aren't built when using cuco in any downstream projects.

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
sleeepyjack: My last commit should have switched to cub but somehow the diff is not showing it. I'll verify this.

ci/build.sh
sleeepyjack: #704 allows for specifying the `cmake` binary while this change here allows for passing build arguments directly to `cmake` as a pass-through.

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
sleeepyjack: See commit diff: https://github.com/NVIDIA/cuCollections/pull/741/commits/aa56fd6398a340e78ce0a58dc60d778b9ff16b4a

benchmarks/roaring_bitmap/contains_bench.cu
```diff
@@ -0,0 +1,97 @@
...
+  auto const num_items   = state.get_int64("NumInputs");
+  auto const bitmap_file = state.get_string_or_default("BitmapFile", {});
+
+  std::ifstream file(bitmap_file, std::ios::binary);
+  if (!file.is_open()) { state.skip("Bitmap file not found"); }
+
+  // Get file size
+  file.seekg(0, std::ios::end);
+  std::streamsize file_size = file.tellg();
+  file.seekg(0, std::ios::beg);
```
mhaseeb123:
```suggestion
  auto file_size = std::filesystem::file_size(bitmap_file);
```

benchmarks/roaring_bitmap/contains_bench.cu
```diff
@@ -0,0 +1,97 @@
...
+  thrust::universal_host_pinned_vector<cuda::std::byte> buffer(file_size);
+
+  file.read(reinterpret_cast<char*>(thrust::raw_pointer_cast(buffer.data())), file_size);
+  file.close();
+
+  cuco::roaring_bitmap<T> roaring_bitmap(thrust::raw_pointer_cast(buffer.data()));
+
+  thrust::device_vector<T> items(num_items);
+
+  key_generator gen{};
```
mhaseeb123, line 58: Just caffeinating the engine ;)

```suggestion
  key_generator gen{0xc0ffee};
```

ci/build.sh
mhaseeb123: Very optional but do we instead want to rename this as `CMAKE_ARGS` which is what libcudf uses to add extra CMAKE_ARGS to the build

examples/roaring_bitmap/host_bulk_example.cu
```diff
@@ -0,0 +1,142 @@
...
+      }
+      for (int k = 700000; k < 800000; ++k) {
+        keys.push_back(k);
+      }
+      return thrust::device_vector<cuda::std::uint32_t>(keys.begin(), keys.end());
+    } else if constexpr (cuda::std::is_same_v<KeyType, cuda::std::uint64_t>) {
+      // Create query keys for the portable_bitmap64.bin file:
+      // https://github.com/RoaringBitmap/RoaringFormatSpec/blob/5177ad9/testdata64/README.md#portable_bitmap64bin
+      std::vector<cuda::std::uint64_t> keys;
+      for (cuda::std::uint64_t k = 0x00000ull; k < 0x09000ull; ++k) {
```
mhaseeb123, line 73: `std::iota` at a few places here

examples/roaring_bitmap/host_bulk_example.cu
```diff
@@ -0,0 +1,142 @@
...
+  std::ifstream file(bitmap_file_path, std::ios::binary);
+  if (!file.is_open()) {
+    std::cerr << "Failed to open " << bitmap_file_path << std::endl;
+    return false;
+  }
+
+  // Get file size
+  file.seekg(0, std::ios::end);
+  std::streamsize file_size = file.tellg();
+  file.seekg(0, std::ios::beg);
```
mhaseeb123:
```suggestion
  auto file_size = std::filesystem::file_size(bitmap_file);
```

examples/roaring_bitmap/host_bulk_example.cu
```diff
@@ -0,0 +1,142 @@
...
+
+int main()
+{
+#ifdef CUCO_ROARING_DATA_DIR
+  std::string const data_dir = CUCO_ROARING_DATA_DIR;
+  bool success               = check<cuda::std::uint32_t>(data_dir + "/bitmapwithoutruns.bin");
+  success &= check<cuda::std::uint32_t>(data_dir + "/bitmapwithruns.bin");
+  success &= check<cuda::std::uint64_t>(data_dir + "/portable_bitmap64.bin");
+
+  std::cout << "success: " << (success ? "true" : "false") << std::endl;
```
mhaseeb123:
```suggestion
  std::cout << "success: " << success << std::boolalpha << std::endl;
```

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
```diff
@@ -0,0 +1,372 @@
...
+#include <thrust/execution_policy.h>
+#include <thrust/fill.h>
+
+namespace cuco::detail {
+
+// primary template
+template <class T>
+class roaring_bitmap_impl {
+  static_assert(cuco::dependent_false<T>, "T must be either uint32_t or uint64_t");
+};
```
mhaseeb123, line 38: Optional: I am not sure if we are using C++20 in cuco yet. If so, we can use concepts here like `concept RoaringBitmapTypes = ...`

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
```diff
@@ -0,0 +1,372 @@
...
+  __device__ bool contains_array_container(cuda::std::byte const* container,
+                                           cuda::std::uint16_t lower,
+                                           cuda::std::uint32_t card) const
+  {
+    cuda::std::uint16_t elem;
+    // Use linear search for small arrays, binary search for larger ones
+    if (card < binary_search_threshold) {
+      for (cuda::std::uint32_t i = 0; i < card; i++) {
+        if constexpr (Aligned) {
+          elem = aligned_load<cuda::std::uint16_t>(container + i * sizeof(cuda::std::uint16_t));
```
mhaseeb123: Minor and optional: We could just have a constexpr function(x) to return `x * sizeof(cuda::std::uint16_t)` for ease.

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
```diff
@@ -0,0 +1,372 @@
...
+      return false;
+    }
+  }
+
+  __device__ bool contains_bitset_container(cuda::std::byte const* container,
+                                            cuda::std::uint16_t lower,
+                                            cuda::std::uint32_t card) const
+  {
+    return static_cast<cuda::std::uint8_t>(container[lower / 8]) &
+           (cuda::std::uint8_t(1) << (lower % 8));
```
mhaseeb123: May a comment about the logic and any ref.

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
```diff
@@ -0,0 +1,372 @@
...
+#include <cuda/std/functional>
+#include <cuda/std/iterator>
+#include <cuda/stream_ref>
+#include <thrust/execution_policy.h>
+#include <thrust/fill.h>
+
+namespace cuco::detail {
+
+// primary template
+template <class T>
```
mhaseeb123, line 35: Looks like we are re-implementing binary search quite a few times in this file. Perhaps we could add a couple of functions in anonymous namespace and reuse. Future PR to do that is also ok.

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
mhaseeb123: Can we document this somewhere for posterity?

include/cuco/detail/roaring_bitmap/util.cuh
```diff
@@ -0,0 +1,226 @@
...
+                                                                            // handling
+      return;
+    }
+
+    if ((cookie & 0xFFFF) == serial_cookie)
+      num_containers = (cookie >> 16) + 1;
+    else {
+      cuda::std::memcpy(&num_containers, buf, sizeof(cuda::std::uint32_t));
+      buf += sizeof(cuda::std::uint32_t);
+    }
```
mhaseeb123, line 104: Don't have a better suggestion but quite a few magic numbers in here 🪄

tests/roaring_bitmap/contains_test.cu
```diff
@@ -0,0 +1,134 @@
...
+#include <cuda/std/type_traits>
+#include <thrust/device_vector.h>
+#include <thrust/logical.h>
+#include <thrust/universal_vector.h>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <fstream>
+#include <string>
+#include <vector>
```
mhaseeb123, line 32: Same comments from `host_bulk_example.cu` about Use of std::iota and std::filesystem::file_size

ci/build.sh
sleeepyjack: I've outsourced the changes to the build script in #749

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
sleeepyjack: This is taken directly from the [RoaringBitmapFormatSpec](https://github.com/RoaringBitmap/RoaringFormatSpec) which I reference a couple of times in the docs. I you read throught the cookie header definition you'll find the following explanation:

> Let size be the number of containers. Then we store (size + 7) / 8 bytes, following the initial 32 bits, as a bitset to indicate whether each of the containers is a run container (bit set to 1) or not (bit set to 0). The first (least significant) bit of the first byte corresponds to the first stored container and so forth. In this scenario, the cookie header uses 32 bits followed by (size + 7) / 8 bytes.

examples/roaring_bitmap/host_bulk_example.cu
sleeepyjack: I would lean towards keeping the original logic as it is a 1:1 copy of how the bitmap was serialize (see link in line 70). Just makes it clearer for the reader where these numbers come from.

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
sleeepyjack: Yeah, that's definitely a candidate for a refactoring. However, there a some details that are slightly different between the implementations which I'd have to abstract from somehow. I would say this is something we can do in a future PR.

include/cuco/detail/roaring_bitmap/util.cuh
sleeepyjack: I cleaned up and simplified some of the logic and added some variable names for these constants. Let me know if I missed something.

benchmarks/roaring_bitmap/contains_bench.cu
sleeepyjack: Ultimately, I would like to use the same default seed for all cuco benchmarks and add an option to adjust it via nvbench. We currently use `std::time(nullptr)` as the default seed which is another problem in itself. So yeah, this will be a separate PR to fix this for the entire library.

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
sleeepyjack: Added some comments and also added a helper function `check_bit()` to reduce code duplication.

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
```diff
@@ -0,0 +1,375 @@
...
+                               InputIt last,
+                               OutputIt contained,
+                               cuda::stream_ref stream = {}) const noexcept
+  {
+    if (this->empty()) {
+      cub::DeviceTransform::Transform(
+        thrust::constant_iterator<bool>(false),
+        contained,
+        cuda::std::distance(first, last),
+        cuda::proclaim_return_type<bool>([] __device__(auto /* dummy */) { return false; }),
```
sleeepyjack, line 81: Can anyone think of a better way of doing a `fill` operation using `cub` that works with arbitrary input iterators? We want to avoid `thrust::fill` as it might introduce some unexpected host syncs.

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
PointKernel: Are there cases where users need to access or manage the metadata when working with a roaring bitmap? If so, yes.

PointKernel (commented):
Looks good to me. One last question needs some discussions. Do we want to get this merged in cudf or cuco?

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
sleeepyjack: From my perspective the user should have no direct interaction with both the `storage` and `metadata` class, as their construction is handled internally. One could argue that we should expose an opaque handle so that users can serialize a `storage` object in memory and create a `ref` from it but I couldn't come up with a good usecase yet.

sleeepyjack:
> Looks good to me. One last question needs some discussions. Do we want to get this merged in cudf or cuco?

@PointKernel The container as is does pass the cuco vibecheck I'd say. The current implementation is limited to lookups but could be extended to support mutable operations as well (although it would require some significant effort).

That said, I'm not at all opposed to moving it into cudf instead - just saying that it could fit into cuco from my standpoint. If you have strong oppinions against including it in cuco I'm open to discuss them. If we move it into cudf I would need someone to lead the effort who is more versed with libcudf details than me.

CC @mhaseeb123 what do you think?

mhaseeb123:
>The container as is does pass the cuco vibecheck I'd say

Agreed with this statement. My vote would be to keep this container in cuco especially considering we may want to expand the set of features in the future and then would need to move it back to cuco. That said, if y'all think it's better to move this to libcudf, I am happy to lead the effort.

PointKernel (approved):
I’m concerned that this is mainly addressing Spark’s requirements rather than providing a general-purpose bitmap solution on GPUs. Once we decide to support insertion, the existing lookups may also need to change. That said, it’s fine for now—let’s merge the PR as is and revisit the full integration later once we have a clearer picture of how cudf/Spark will use this data structure.

mhaseeb123 (approved):

mhaseeb123:
> I’m concerned that this is mainly addressing Spark’s requirements rather than providing a general-purpose bitmap solution on GPUs.

Agreed. I don't see any requirement from Spark or libcudf, in the near term at least, to add insertions, deletions etc.

>Once we decide to support insertion, the existing lookups may also need to change.

Yup and we may need to move it back to cuco. 

> That said, it’s fine for now—let’s merge the PR as is and revisit the full integration later once we have a clearer picture of how cudf/Spark will use this data structure.

Agreed with this. We can always move it to libcudf if needed.

include/cuco/detail/roaring_bitmap/roaring_bitmap_impl.cuh
sleeepyjack: I wish we had C++20 concepts 😢 
For now we're stuck with C++17 unfortunately

PointKernel (approved):
Looks good. @sleeepyjack Thanks!

PR #743, opened 2025-07-17
Add clang CI test

### LINKED ISSUE #128: [FEA] Make cuco compilable using clang
Any project dependent on cuco cannot be compiled using clang if the headers in cuco cannot be compiled using clang.

**Describe the solution you'd like**
It would be great if cuco has a check in place verifying that commits can be compiled using clang.
I'm currently setting something like this up for RAFT (see https://github.com/rapidsai/raft/pull/424) so I'd be happy to provide something similar for cuco.

**Additional context**
In order to use something like `clang-tidy`, plus to get better compiler diagnostics, it's always good that code can be compiled using clang.

A few specific things I ran into while applying the above mentioned PR to RAFT (all applies to git tag [f0eecb203590f1f4ac4a9f1700229f4434ac64dc](https://github.com/NVIDIA/cuCollections/tree/f0eecb203590f1f4ac4a9f1700229f4434ac64dc) of this repo which RAFT depends on at the moment):
1. include/cuco/detail/static_map.inl:336:20, include/cuco/detail/static_map.inl:252:20, include/cuco/detail/static_map.inl:216:21, include/cuco/detail/static_map.inl:324:23: should use `this->[[next/initial]]_slot`
2. include/cuco/detail/static_map.inl:119:12: requires `typename` keyword before `static_map`

### LINKED ISSUE #744: [BUG]: Fix segmentation fault during linking when building benchmarks with Clang
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Describe the bug

When building cuco with Clang as the host compiler, the benchmark targets fail to link due to a segmentation fault during the linking stage. This issue needs to be investigated and resolved to support Clang builds properly.

### How to Reproduce

Build cuco with Clang as the host compiler within any of the available cuco devcontainers:
```bash
./ci/build.sh --cxx clang++
```

### Expected behavior

The build should complete without any warnings or errors.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

PointKernel:
Fixes #744
Fixes #128

This PR adds a CI task to test building with Clang as the host compiler and addresses associated build warnings and errors.


sleeepyjack:
/ok to test [ad63e43](https://github.com/NVIDIA/cuCollections/pull/743/commits/ad63e43255225d2aec7fcbc5901273d380b03b16)

.github/workflows/dispatch-build-and-test.yml
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
sleeepyjack, line 1: no changes in this file so technically we could leave the copyright year as is.

include/cuco/detail/hash_functions/murmurhash3.cuh
```diff
@@ -233,6 +233,8 @@ struct MurmurHash3_32 {
         k1 = rotl32(k1, 15);
         k1 *= c2;
         h1 ^= k1;
+        [[fallthrough]];
+      case 0: break;
```
sleeepyjack, line 237: This should also go into the CCCL port

include/cuco/detail/utility/cuda.cuh
```diff
@@ -48,7 +48,7 @@ __device__ constexpr int32_t warp_size() noexcept { return 32; }
  *
  * @return The global thread index
  */
-__device__ static index_type global_thread_id() noexcept
+[[maybe_unused]] __device__ static index_type global_thread_id() noexcept
```
sleeepyjack: Why is this needed?

.github/actions/compute-matrix/compute-matrix.py
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env python3
```
sleeepyjack: What's the difference between the bash and the python implementation?

sleeepyjack (commented):
It seems GH actions is encountering a syntax error: https://github.com/NVIDIA/cuCollections/actions/runs/16486736912/workflow

.github/actions/compute-matrix/compute-matrix.py
PointKernel: Removed as it's only for testing

include/cuco/detail/utility/cuda.cuh
PointKernel: This is related to an odd compilation warning in Clang: if a function is declared `static` but isn't used within the translation unit, Clang emits an "unused function" warning. The most reliable way I've found to make the code compile cleanly with both GCC and Clang is to add the `[[maybe_unused]]` attribute.

include/cuco/detail/utility/cuda.cuh
```diff
@@ -41,14 +41,14 @@ namespace cuco {
...
-__device__ constexpr int32_t warp_size() noexcept { return 32; }
+[[nodiscard]] __device__ constexpr int32_t warp_size() noexcept { return 32; }
 
 /**
  * @brief Returns the global thread index in a 1D scalar grid
  *
  * @return The global thread index
  */
-__device__ static index_type global_thread_id() noexcept
+[[nodiscard]] __device__ inline index_type global_thread_id() noexcept
```
PointKernel, line 51: `inline` is needed to silence the multiple definition error with clang

sleeepyjack (approved):
Thanks for going through this CI struggle. Nice work!

PR #746, opened 2025-07-31
Update device retrieve signature to pass atomic counter by reference

PointKernel:
We always pass atomic counter by pointer instead of by reference. This unblocks the mixed join migration with multiset.


sleeepyjack (approved):
What's the reason we use pass-by-pointer instead of by reference here? Do we expect the object behind this pointer to not be initialized at one point?

PointKernel:
> What's the reason we use pass-by-pointer instead of by reference here? Do we expect the object behind this pointer to not be initialized at one point?

Thanks. I've updated the PR so that atomic counters for device retrieve are always passed by reference.

sleeepyjack (approved):

PR #748, opened 2025-08-14
Pass ParentT to CG type and pass CG objects by-value

sleeepyjack:
This PR overhauls the way we use cooperative groups.

- Pass `ParentT` explicitly to the `thread_block_tile` type if possible, leading to an object size reduction from 16B->1B.
- Pass CG objects by-value if possible to avoid blocking registers for an extended amount of time.


include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -419,14 +419,15 @@ class open_addressing_ref_impl {
...
    *
    * @param group The Cooperative Group used to perform group insert
    * @param value The element to insert
    *
    * @return True if the given element is successfully inserted
    */
-  template <bool SupportsErase, typename Value>
-  __device__ bool insert(cooperative_groups::thread_block_tile<cg_size> const& group,
+  template <bool SupportsErase, typename Value, typename ParentCG>
+  __device__ bool insert(cooperative_groups::thread_block_tile<cg_size, ParentCG> group,
```
PointKernel, line 430:
```suggestion
  __device__ bool insert(cooperative_groups::thread_block_tile<cg_size, cooperative_groups::thread_block> group,
```
do we want to make it sub-CG compatible? If not, `thread_block` is probably good enough

sleeepyjack: It depends on what the user passes to the function. They might work with a warp group at some level and then decide to use a CG of size 8 of that warp to perforn the `insert`. So this is about flexibility.

PointKernel (approved):

PR #755, opened 2025-09-03
Synchronize CUDA stream before variable usage

caaatch22:
To get correct value of `h_num_successes`, we need to sync stream to ensure `h_num_successes` is updated.
Same for `h_submap_num_successes`


PointKernel (approved):
LGTM @caaatch22 thanks for the fix!

PR #758, opened 2025-09-12
Fix infinite loop in  `dynamic_map::find/contains/erase` after all keys are erased from a submap

caaatch22:
As titled.

Also, are there any plans to replace `legacy::static_map` with `static_map` in the implementation of `dynamic_map`?

### DIFF
--- a/tests/dynamic_map/find_test.cu
+++ b/tests/dynamic_map/find_test.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2025, NVIDIA CORPORATION.
+ * Copyright (c) 2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


tests/dynamic_map/find_test.cu
```diff
@@ -0,0 +1,209 @@
+/*
+ * Copyright (c) 2022-2025, NVIDIA CORPORATION.
```
PointKernel:
```suggestion
 * Copyright (c) 2025, NVIDIA CORPORATION.
```

PointKernel:
> Also, are there any plans to replace legacy::static_map with static_map in the implementation of dynamic_map?

It’s already on the roadmap, and we have an early version of the new `dynamic_map` under the `experimental` namespace. I’ll make this work a priority.

PointKernel (approved):
Looks great. 

@caaatch22 thanks again for the contributions!

PR #762, opened 2025-09-25
Add static_map heterogeneous lookup example

### LINKED ISSUE #714: [ENHANCEMENT]: Add an example for performing a heterogeneous lookup

All cuco hash tables support heterogeneous lookups, which have proven to be a powerful feature for tackling real-world problems. It would be valuable for cuco to include an example demonstrating how to perform heterogeneous lookups using its data structures.

Heterogeneous lookups are typically used together with the mapping table method demonstrated in [this example](https://github.com/NVIDIA/cuCollections/blob/dev/examples/static_set/mapping_table_example.cu). However, incorporating heterogeneous lookup directly into that example could make it harder to follow so we should do something better.

cnaples79:
## Summary
- add a static_map example that demonstrates hashing and comparing compatible key types
- show how the map can answer `contains` and `find` queries using probe keys with extra fields
- register the example with the examples CMake target so it builds alongside existing samples

## Rationale
- addresses NVIDIA/cuCollections#714 by giving users a reference for heterogeneous lookups without overloading the existing mapping-table example

## Changes
- new `static_map/heterogeneous_lookup_example.cu` with custom hash/equality for tuple probes
- build system entry so the sample is produced with the rest of the examples
- Tests not run (examples)

Fixes #714


PointKernel (approved):
Looks good!

I’ve updated the README accordingly and did a bit of cleanup with some minor fixes. Thanks for the contribution, @cnaples79.

PR #766, opened 2025-10-08
Update cuda_allocator to support stream-ordered allocation

### LINKED ISSUE #757: [ENHANCEMENT]: Support stream-ordered allocators in cuco data structures

cuco data structures are intended to support multi-stream usage e.g. building a hash table on one stream and probing it on another. However, this is not fully correct in the current implementation.

Internally, cuco always uses the allocator passed to the constructor. Today, we assume the allocator is not stream-ordered to match STL behavior. This creates problems when users wrap a non-stream-ordered allocator with an `rmm::mr::stream_adaptor` and pass it to cuco.

During construction, allocations work correctly.

During probing on a different stream, temporary allocations are still made using the allocator tied to the build stream. The probe API then immediately launches work on the probe stream, leading to errors.

We need to update cuco’s allocator handling and storage logic to fully support stream-ordered allocation and deallocation for both persistent and temporary storage in all cuco data structures.

PointKernel:
Closes #757
Depends on #765

This PR updates the `cuco::cuda_allocator` to support stream-ordered allocation and dealloation.


include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.inl
```diff
@@ -217,8 +217,8 @@ constexpr void dynamic_bitset<Allocator>::build_ranks_and_selects(
 #else
   stream.wait();
 #endif
-  std::allocator_traits<temp_allocator_type>::deallocate(
-    temp_allocator, thrust::device_ptr<char>{reinterpret_cast<char*>(d_sum)}, sizeof(size_type));
+  temp_allocator.deallocate(thrust::device_ptr<char>{reinterpret_cast<char*>(d_sum)},
```
sleeepyjack, line 220: Why do we need `thrust::device_ptr` here?

sleeepyjack (approved):
Looks good. Only one small oddity in `dynamic_bitset` which isn't directly related to this PR.

include/cuco/detail/trie/dynamic_bitset/dynamic_bitset.inl
PointKernel: `dynamic_bitset` currently uses `thrust::device_vector` as its internal container, which needs to be corrected.

PointKernel:
I need more testing with cudf to merge this PR since it currently breaks several cudf tests

PR #768, opened 2025-10-14
Add CMake tests to verify that each header is self-contained

PointKernel:
This PR adds a header test to verify that all headers are self-contained. It also fixes several issues uncovered by the test, along with some circular dependency problems.


cmake/header_testing.cmake
```diff
@@ -0,0 +1,90 @@
...
+# limitations under the License.
+#=============================================================================
+
+# For every public header, build a translation unit containing `#include <header>`
+# to let the compiler try to figure out warnings in that header if it is not otherwise
+# included in tests, and also to verify if the headers are modular enough.
+# .inl files are not globbed for, because they are not supposed to be used as public
+# entrypoints.
+
+# Meta target for all header builds:
```
PointKernel:
```suggestion
```

CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ endif()
 option(BUILD_TESTS "Configure CMake to build tests" ${default_build_option_state})
 option(BUILD_BENCHMARKS "Configure CMake to build (google) benchmarks" ${default_build_option_state})
 option(BUILD_EXAMPLES "Configure CMake to build examples" ${default_build_option_state})
+option(BUILD_HEADER_TESTS "Configure CMake to build header tests" ${default_build_option_state})
```
sleeepyjack: Can we make this part of the regular test suite instead?

sleeepyjack (commented):
Great idea!

sleeepyjack (approved):

PR #777, opened 2025-12-05
Replace cudaMemcpyAsync with cudaMemcpyBatchAsync to avoid locking

PointKernel:
This PR switches from `cudaMemcpyAsync` to `cudaMemcpyBatchAsync` to eliminate a performance regression caused by driver-side locking in the legacy memcpy path. Using the new batch-async API removes that locking overhead and restores the expected performance.


include/cuco/detail/utility/memcpy_async.cuh
```diff
@@ -0,0 +1,69 @@
...
+ * @param stream CUDA stream for the operation
+ */
+inline void memcpy_async(
+  void* dst, void const* src, size_t count, cudaMemcpyKind kind, cuda::stream_ref stream)
+{
+  if (dst == nullptr || src == nullptr || count == 0) { return; }
+
+#if CUDART_VERSION >= 12080
+  if (stream.get() == 0) {
+    CUCO_CUDA_TRY(cudaMemcpyAsync(dst, src, count, kind, stream.get()));
```
bdice: If you make this return the cudaError_t and do the CUCO_CUDA_TRY at the call site, I think it may be easier to track down where errors come from with file/line?

include/cuco/detail/utility/memcpy_async.cuh
```diff
@@ -0,0 +1,69 @@
...
+ * Uses cudaMemcpyBatchAsync for CUDA 12.8+ with proper edge case handling.
+ * Falls back to cudaMemcpyAsync for older CUDA versions or edge cases.
+ *
+ * @param dst Destination memory address
+ * @param src Source memory address
+ * @param count Number of bytes to copy
+ * @param kind Memory copy direction
+ * @param stream CUDA stream for the operation
+ */
+inline void memcpy_async(
```
bdice: I’d like to see if we can align on the exact semantics of this between cuCo and libcudf. My libcudf PR has a few subtle differences that we should iron out. I’ll study this PR and tweak mine accordingly but we may have more to share and learn.

For example, should we write this as “memcpy_batch_async” and call it from another function “memcpy_async”? That way we have the ability to do a batched copy if there is a use case in cuCo, and it matches the libcudf design.

include/cuco/detail/utility/memcpy_async.cuh
PointKernel: Good point. Updated.

include/cuco/detail/utility/memcpy_async.cuh
PointKernel: > I’d like to see if we can align on the exact semantics of this between cuCo and libcudf.

Agreed! Feel free to drop comments on this PR anytime you spot something that needs fixing or improving.

> should we write this as “memcpy_batch_async” and call it from another function “memcpy_async”?

Great question! I actually noticed this difference between your approach and the current PR, and I intentionally hid the fact that we’re using batch async memcpy under the hood. Since cuco doesn’t have any real use cases for batch memcpy, we’re only relying on it as a workaround for the driver locking issue in the legacy async memcpy. I felt it was cleaner to keep that detail internal. From a user’s perspective, it’s just an implementation detail they don’t need to deal with.

In libcudf, though, batch async memcpy is used more broadly IIRC, so exposing both the legacy and batch variants there makes a lot more sense.

sleeepyjack (approved):
LGTM
Could you share some details offline on why this fix is needed?

PointKernel:
> LGTM Could you share some details offline on why this fix is needed?

I’ve expanded the PR description with more context on the issue. In short, `cudaMemcpyAsync` incurs a costly driver lock when used across multiple streams, which leads to a noticeable performance hit. Switching to the new batch async API removes this locking behavior and resolves the regression.

bdice (changes requested):
The symbol `cudaMemcpyBatchAsync` exists for CUDA 12.8+ but this won't work with older CUDA 12 releases. We don't see this in cuCollections CI because it doesn't test minor version compatibility (cudf's CI builds with 12.9 and runs tests with 12.2).

@PointKernel and I discussed possible workarounds that would support CUDA 12.8+. All involve checking for and using the [driver API `cuMemcpyBatchAsync`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g6f1ff58e3065df3eb4b573dba77ad31f) instead of the runtime API.

- We might be able to use [`cudaGetDriverEntryPoint`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html#group__CUDART__DRIVER__ENTRY__POINT_1g0f58af55719ee7b5977d8fa3d0f4d898) (deprecated in CUDA 13.0) to get the driver symbol for [`cuMemcpyBatchAsync`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g6f1ff58e3065df3eb4b573dba77ad31f) and use that, but I think there are some sharp edges with this approach which led to the introduction of [`cudaGetDriverEntryPointByVersion`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html#group__CUDART__DRIVER__ENTRY__POINT_1gda6c90a43afbe8043128285afbaf8c04).
  - We can't use [`cudaGetDriverEntryPointByVersion`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html#group__CUDART__DRIVER__ENTRY__POINT_1gda6c90a43afbe8043128285afbaf8c04) because it requires CUDA 12.5+, which kind of defeats the purpose of supporting CUDA < 12.8.
- CCCL has a [somewhat complex solution](https://github.com/NVIDIA/cccl/blob/ed8fd7fba95182d538ab0cd9730a6367a7f15f89/libcudacxx/include/cuda/__driver/driver_api.h#L58) for dlopen'ing the driver and getting `cuProcGetAddress`, which we could then use to check for `cuMemcpyBatchAsync` support. This definitely works but it's a lot of work to get right.

For now, we concluded it's easiest to just require CUDA 13, by using the condition on `#if CUDART_VERSION > 13000`. For CUDA 12.8 and 12.9, we would just use `cudaMemcpyAsync` as in the current state.

I plan to adopt the same solution for https://github.com/rapidsai/cudf/pull/20800 until we can invest the additional engineering time to support CUDA 12.8-12.9.

bdice (approved):

bdice:
Thanks! Looks good (though I lack approval power).

PR #781, opened 2025-12-11
Implement set union and (approximate) intersection for bloom filters

### LINKED ISSUE #602: [ENHANCEMENT]: Add bloom_filter::intersect/union

Extension of #573 

Calculating the set intersection and union of two or more Bloom filters has various applications and we should support it.

Implement both host and device function that perform intersection/union of filter similar to how `distinct_count_estimator` is able to merge two filters into one (see [device](https://github.com/NVIDIA/cuCollections/blob/4454de4b878f31d41c5b7578fe6ca24bba5ea3f4/include/cuco/distinct_count_estimator_ref.cuh#L147) and [host](https://github.com/NVIDIA/cuCollections/blob/4454de4b878f31d41c5b7578fe6ca24bba5ea3f4/include/cuco/distinct_count_estimator.cuh#L165) `merge`).

_No response_

### Additional context

_No response_

wence-:
Given two bloom filters it is often useful to be able to compute their union and intersection. That is, suppose we have two sets A and B, denote by f(.) the construction of a bloom filter with a given set of parameters. Then we would like to be able to (approximate) f(A v B) and f(A ^ B) given f(A) and f(B). The union operation is exact, and obtained by the bitwise or of fingerprints, so we have f(A v B) = f(A) v f(B). In contrast the intersection, obtained by bitwise and of the fingerprints is only approximate so we have f(A ^ B) ~= f(A) ^ f(B).

- Closes #602.


include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -645,6 +646,54 @@ class bloom_filter_impl {
...
+#else
+    stream.wait();
+#endif
+  }
+
+  __host__ constexpr void merge_async(bloom_filter_impl<Key, Extent, Scope, Policy> const& other,
+                                      cuda::stream_ref stream)
+  {
+    CUCO_EXPECTS(this->block_extent() == other.block_extent(),
+                 "mismatching num_blocks in merge_async");
```
sleeepyjack, line 664: We have to extend these checks a bit. Union/intersection is only valid iff the hashing and bit pattern generation logic of the two filter object is identical. Besides checking for equal filter size, we would also need to check for equality of the `impl.policy_` objects ([here](https://github.com/NVIDIA/cuCollections/blob/b4e5eb2044a3ca700d72d9c69a99d319f6555393/include/cuco/detail/bloom_filter/bloom_filter_impl.cuh#L690)).

include/cuco/bloom_filter.cuh
```diff
@@ -142,7 +142,7 @@ class bloom_filter {
    * @brief Adds all keys in the range `[first, last)` to the filter.
    *
    * @note This function synchronizes the given stream. For asynchronous execution use
-   * `add_async`.
+   * add_async().
```
sleeepyjack, line 145: Why remove the text style?

include/cuco/bloom_filter.cuh
```diff
@@ -325,6 +325,85 @@ class bloom_filter {
...
+   *
+   * @note Modifies `this` in place.
+   * @note This function synchronizes the given stream. For asynchronous execution use
+   * intersect_async().
+   *
+   * @note This performs the set intersection of the two filters. Unlike merge(), this operation
+   * does not distribute over filter construction and therefore only approximates the bloom filter
+   * of the intersection of the input sets. In other words, let \f$f : X \to B\f$ denote the
+   * construction of a bloom filter on some set \f$X\f$, and let \f$A\f$ and \f$B\f$ be two sets,
+   * then \f$(A \cap B) \ne f(A) \cap f(B)\f$.
```
sleeepyjack: In my understanding the intersection is still exact in the sense that it doesn't introduce any false-negatives if any of the items in the (true) intersection is probed against the intersected filter. This is important as this is a fundamental guarantee of the data structure.
That said, the intersected filter might contain some "ghost" bits, i.e., bits that are not part of the true set intersection, but are "coincidentally" set in both filters. This will drive the false-positive rate up but does not break the logic.

include/cuco/bloom_filter.cuh
wence-: Now it links to the method

include/cuco/bloom_filter.cuh
wence-: Correct, I can adjust the docs a bit

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
wence-: Ugh, this is complicated because none of those objects implement equality and we don't have equality defined on the hash implementations either. Since users can provide customer hashers making a change to require equality would be breaking.

I am inclined to note in the docs that the policy objects of the two filters must be equal otherwise behaviour is undefined. What do you think?

sleeepyjack: You're right. The hasher and technically also the policy itself are user customization points. IIRC `std::hash` also doesn't provide equality operators whatsoever.
We're still doing type checking by requiring `other` to be of same type of `this`. I think this should catch most of the footguns already. You can add a note in the docs that the hasher and policy objects must match otherwise the result of this operation is undefined/invalid.

include/cuco/bloom_filter.cuh
sleeepyjack: Ha, good point. TIL

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
wence-: Added that the policies must match otherwise undefined. That covers the hasher too, because the policy holds the hasher.

sleeepyjack:
@wence- we would need some rudimentary unit tests for this feature. I can pick this up later today or tomorrow.

sleeepyjack:
/ok to test https://github.com/NVIDIA/cuCollections/commit/65d2402e0aa055a00ae9feaba90975bf441aef65

PointKernel (approved):
Looks great! @wence- thanks for the contribution!

@sleeepyjack  Feel free to merge once the unit tests are in place.

PR #784, opened 2025-12-18
Add Hyperloglog add_if_async host API

PointKernel:
This PR adds a new `add_if_async` API that allows users to conditionally perform add operations without the need for custom kernels. In addition, it resolves a small type-deduction issue when deriving the hash value type with `decltype`, ensuring that references and `const` qualifiers are properly removed.

Required by https://github.com/rapidsai/cudf/pull/20735

### DIFF
--- a/include/cuco/detail/hyperloglog/kernels.cuh
+++ b/include/cuco/detail/hyperloglog/kernels.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, NVIDIA CORPORATION.
+ * Copyright (c) 2024-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


include/cuco/detail/hyperloglog/kernels.cuh
```diff
@@ -58,29 +96,30 @@ CUCO_KERNEL void add_shmem_vectorized(typename RefType::value_type const* first,
...
 
-  // each thread processes VectorSize-many items per iteration
   vector_type vec;
   while (idx < n / VectorSize) {
     vec = *reinterpret_cast<vector_type*>(
       __builtin_assume_aligned(first + idx * VectorSize, sizeof(vector_type)));
-    for (auto const& i : vec) {
-      local_ref.add(i);
+    for (auto i = 0; i < VectorSize; ++i) {
+      if (pred(*(stencil + idx * VectorSize + i))) { local_ref.add(vec[i]); }
```
sleeepyjack, line 104: It seems the compiler is still able to unroll the loop if the stencil and pred are constants. https://godbolt.org/z/e7xaeh6sn (it's a bit hard to find the relevant kernels in the SASS view as the kernels are so long and there are so many instantiations. `Strg-F` for `add_shmem_vectorized<16` and `add_if_shmem_vectorized<16`). I can't run comparison benchmarks right now as clab has some issues but I'm fairly certain the change doesn't introduce any regressions.

include/cuco/types.cuh
```diff
@@ -58,6 +58,15 @@ CUCO_DEFINE_STRONG_TYPE(sketch_size_kb, double);
  */
 CUCO_DEFINE_STRONG_TYPE(standard_deviation, double);
 
+/**
+ * @brief A strong type wrapper `cuco::precision` for specifying the HyperLogLog precision
+ * parameter of `cuco::hyperloglog(_ref)`.
+ *
+ * @note Precision `p` determines the number of registers as `2^p`. Valid range is typically [4,
+ * 18].
+ */
```
sleeepyjack: `precision` is a quite generic term. I can't come up with something better but reserving this term exclusively for HLL feels not right.

sleeepyjack (approved):

include/cuco/types.cuh
PointKernel: Similar to we do during the CCCL migration, shall we make this being defined in the HLL ref header so that it's a HLL only term?

PointKernel: Moved to HLL ref header.

PR #788, opened 2026-01-20
Add support for roaring bitmaps with run containers and fewer than fo…

sleeepyjack:
This PR adds support for a previously omitted edge case of the Roaring bitmap format spec. If a serialized 32-bit roaring bitmap contains run containers but has fewer than 4 containers in total, the spec demands omitting the container offset array and calculate the offsets on-the-fly.

### DIFF
--- a/include/cuco/detail/roaring_bitmap/util.cuh
+++ b/include/cuco/detail/roaring_bitmap/util.cuh
@@ -66,6 +66,10 @@ template <>
 struct roaring_bitmap_metadata<cuda::std::uint32_t> {
   /// Maximum number of elements in an array container before converting to bitmap
   static constexpr cuda::std::uint32_t max_array_container_card = 4096;
+  /// Threshold for omitting container offsets in serialized format
+  static constexpr cuda::std::int32_t no_offset_threshold = 4;
+  /// Fixed size of a bitset container in bytes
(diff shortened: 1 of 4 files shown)


include/cuco/detail/roaring_bitmap/util.cuh
```diff
@@ -75,14 +79,18 @@ struct roaring_bitmap_metadata<cuda::std::uint32_t> {
   cuda::std::uint32_t run_container_bitmap = 0;
   /// Offset to key cardinality data
   cuda::std::uint32_t key_cards = 0;
-  /// Offset to container offset data
+  /// Offset to container offset data (only valid when offsets_in_serialized_data is true)
   cuda::std::uint32_t container_offsets = 0;
+  /// Computed container offsets (used when offsets are not in serialized data)
+  cuda::std::uint32_t computed_offsets[4] = {0, 0, 0, 0};
```
gerashegalov: would it make sense to use the mnemonic constant here ?
```suggestion
  cuda::std::uint32_t computed_offsets[no_offset_threshold] = {0, 0, 0, 0};
```

PointKernel (approved):
@sleeepyjack Is there a GitHub issue or discussion we can link to for reference on this fix? I don’t fully follow the details here, but I trust that your changes are valid. :)

PR #793, opened 2026-01-23
Remove the legacy multimap

PointKernel:
This PR removes the legacy multimap and promotes the experimental implementation to the default.


tests/static_multimap/heterogeneous_lookup_test.cu
```diff
@@ -31,7 +31,7 @@
 
 // insert key type
 template <typename T>
-struct key_pair {
+struct __align__(sizeof(T) * 2) key_pair {
```
sleeepyjack: I guess this is correct but I'm not sure about the implications.

examples/static_multimap/host_bulk_example.cu
```diff
@@ -52,23 +53,22 @@ int main(void)
...
-  // Finds all keys {0, 1, 2, ...} and stores associated key/value pairs into `d_results`
-  // If a key `keys_to_find[i]` doesn't exist, `d_results[i].second == empty_value_sentinel`
-  auto output_end = map.retrieve_outer(keys_to_find.begin(), keys_to_find.end(), d_results.begin());
-  auto retrieve_size = output_end - d_results.begin();
-
-  // The total number of outer matches should be `N + N / 2`
-  assert(not(output_size == retrieve_size == N + N / 2));
+  if (num_found == N / 2) {
+    std::cout << "Success! All " << N / 2 << " unique keys found in the multimap.\n";
+    std::cout << "Each key has 2 duplicate values, for a total of " << N << " pairs.\n";
```
sleeepyjack:
```suggestion
    std::cout << "Each key has 2 duplicate values, for a total of " << N << " pairs." << std::endl;
```

Use `std::endl` at the very end to flush the stream.

sleeepyjack (approved):
LGTM

Can you please add a deprecation note to our README?

tests/static_multimap/heterogeneous_lookup_test.cu
PointKernel: AI leftover, removed.

PR #795, opened 2026-02-03
Fix storage alignment

### LINKED ISSUE #782: [BUG]: cuco::bucket_storage might not align to sizeof(bucket_type)
`cuco::bucket_storage` [currently](https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/bucket_storage.cuh#L44-L45) only check's for alignment of sizeof(cuco::bucket_type), but doesn't acutally do any kind of aligned allocation.

PointKernel:
Closes https://github.com/NVIDIA/cuCollections/issues/782

This PR fixes a storage alignment issue where the alignment requirement was checked but the data was not actually aligned.

### DIFF
--- a/include/cuco/bucket_storage.cuh
+++ b/include/cuco/bucket_storage.cuh
@@ -30,6 +30,7 @@
 #include <memory>
 
 namespace cuco {
+
 /**
  * @brief Non-owning array of slots storage reference type.
  *
@@ -40,10 +41,14 @@ namespace cuco {
 template <typename T, int32_t BucketSize, typename Extent = cuco::extent<std::size_t>>
 class bucket_storage_ref {
  public:
-  static constexpr int32_t bucket_size   = BucketSize;  ///< Number of elements processed per bucket
-  using bucket_type                      = cuda::std::array<T, BucketSize>;  ///< Slot bucket type
-  static constexpr std::size_t alignment = cuda::std::min(cuda::std::bit_ceil(sizeof(bucket_type)),
-                                                          std::size_t{16});  ///< Required alignment
+  static constexpr int32_t bucket_size = BucketSize;        ///< Number of elements per bucket
+  static constexpr std::size_t max_vector_load_bytes = 16;  ///< Maximum vector load width in bytes
+
+  using bucket_type = cuda::std::array<T, BucketSize>;  ///< Slot bucket type
+
+  static constexpr std::size_t alignment =
+    cuda::std::min(cuda::std::bit_ceil(sizeof(bucket_type)),
+                   max_vector_load_bytes);  ///< Required alignment in bytes
 
   using extent_type = Extent;                            ///< Storage extent type
   using size_type   = typename extent_type::value_type;  ///< Storage size type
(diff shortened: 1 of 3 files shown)


include/cuco/detail/storage/bucket_storage.inl
```diff
@@ -98,10 +100,16 @@ bucket_storage_ref<T, BucketSize, Extent>::extent() const noexcept
...
 constexpr bucket_storage<T, BucketSize, Extent, Allocator>::bucket_storage(
   Extent size, Allocator const& allocator, cuda::stream_ref stream)
-  : extent_{size},
-    allocator_{allocator},
-    slots_{allocator_.allocate(capacity(), stream),
-           slot_deleter_type{capacity(), allocator_, stream}}
+  : extent_{size}, allocator_{allocator}, slots_{[this, &stream]() {
+      constexpr std::size_t align  = ref_type::alignment;
+      constexpr std::size_t extra  = (align - 1) / sizeof(value_type) + 1;
+      std::size_t const alloc_size = static_cast<std::size_t>(capacity()) + extra;
```
sleeepyjack, line 106: So we are always overallocating `extra` since we cannot infer the alignment from the allocator type before doing the actual allocation? I wonder if the CCCL mr interface has a way to specify the alignment. That would be much nicer IMHO.

include/cuco/bucket_storage.cuh
```diff
@@ -40,14 +40,14 @@ namespace cuco {
 template <typename T, int32_t BucketSize, typename Extent = cuco::extent<std::size_t>>
 class bucket_storage_ref {
  public:
-  static constexpr int32_t bucket_size = BucketSize;  ///< Number of elements processed per bucket
-  static constexpr std::size_t alignment =
-    cuda::std::min(sizeof(T) * bucket_size, std::size_t{16});  ///< Required alignment
+  static constexpr int32_t bucket_size   = BucketSize;  ///< Number of elements processed per bucket
+  using bucket_type                      = cuda::std::array<T, BucketSize>;  ///< Slot bucket type
+  static constexpr std::size_t alignment = cuda::std::min(cuda::std::bit_ceil(sizeof(bucket_type)),
+                                                          std::size_t{16});  ///< Required alignment
```
sleeepyjack: We should define the alignment requirement as a global variable.

tests/utility/storage_test.cu
```diff
@@ -96,4 +100,107 @@ TEMPLATE_TEST_CASE_SIG("utility storage tests",
...
+
+  SECTION("Data pointer is aligned to bucket boundary.")
+  {
+    auto s = storage_type(cuco::extent{size}, allocator, cuda::stream_ref{cudaStream_t{nullptr}});
+
+    auto const ptr       = reinterpret_cast<std::uintptr_t>(s.data());
+    auto const alignment = storage_ref_type::alignment;
+
+    REQUIRE((ptr % alignment) == 0);
+  }
```
sleeepyjack: These are transitively tested through the `_ref` class.

include/cuco/detail/storage/bucket_storage.inl
PointKernel: I agree. I’m considering non-CCCL / non-RMM memory resources or allocators that don’t expose alignment parameters; this approach guarantees correctness with any arbitrary memory resource or allocator.

sleeepyjack (approved):

PR #799, opened 2026-03-25
Bump rapids-cmake to 26.04 and update devcontainers

sleeepyjack:


PointKernel (approved):
nice! finally got new runners 🔥

PR #802, opened 2026-04-16
Replace precomputed prime table with on-the-fly Miller-Rabin computation

PointKernel:
This PR replaces the precomputed prime table lookup with an on-the-fly calculation. While the new approach is slightly slower, it adds at most ~2 µs of overhead, making the performance impact negligible, and eliminates the need to store a large precomputed prime table.


tests/static_multiset/insert_test.cu
```diff
@@ -96,7 +96,7 @@ TEMPLATE_TEST_CASE_SIG(
   constexpr size_type gold_capacity = [&]() {
     if constexpr (cuco::is_double_hashing<probe>::value) {
       return (CGSize == 1) ? 422   // 211 x 1 x 2
-                           : 412;  // 103 x 2 x 2
+                           : 404;  // 101 x 2 x 2
```
sleeepyjack, line 99: why is this change needed?

sleeepyjack (commented):
Ha, a Miller-Rabin test in the wild! My old Discrete Math 101 professor would be proud of us :)

The diff in the prime.hpp is too large for the GH web UI to open. I inspected the file manually. One thing that caught my eye is that we use 128bit integers to prevent overflow. We have a macro check `CUCO_HAS_INT128` that we should use in this case and have a fallback ready.

tests/static_multiset/insert_test.cu
PointKernel: This is required because the previous prime table had missing entries (e.g., 101), so this change fixes that bug.

sleeepyjack (approved):

Issue #805, opened 2026-04-27
Relax key/payload/slot type constraints to enable 1B and 2B types


sleeepyjack:
> I think we are too strict on the key/payload/slot type constraints. `cuda::atomic_ref` supports 1B and 2B operations. This is is to be explored in a future PR. 

 _Originally posted by @sleeepyjack in [#803](https://github.com/NVIDIA/cuCollections/pull/803/changes#r3150749887)_

ksapru:
This seems like a good first issue for me to explore this repository and learn more. Happy to take this on if it's still up for grabs. cc: @sleeepyjack

sleeepyjack:
Hi @ksapru,

Yes, this is a good fit for "good first issue." I'm fairly sure our implementation already works with key/value types smaller than 4 bytes (famous last words).

The best way to approach this is to add `uint8_t` and `uint16_t` to the type axes in our unit tests. The `static_set` tests are a good starting point. Once those pass, the rest should fall in line easily, since the other data structures share the same implementation in `detail/open_addressing_ref_impl.cuh` and `detail/open_addressing_impl.cuh`.

Let me know if you need any help getting started.

sleeepyjack:
Next steps after #812 : (1) Add (almost) exhaustive test coverage to all OA-based data structures and (2) extend benchmarks.

amacharla15:
Hi @sleeepyjack  I'd be interested in picking up the remaining test-coverage work after #812.

I can start by extending the 1B/2B (`uint8_t` / `uint16_t`) type coverage across the remaining open-addressing data structures and operations, and keep the benchmark extension separate if that makes the review easier.

Is that still available to work on?

PR #807, opened 2026-04-29
Fix dangling allocator reference in storage deleters

ksapru:
## Description
This PR resolves intermittent CI failures (specifically row count mismatches in `rapidsmpf` pipelines) caused by a dangling allocator reference in `cuco` storage deleters.

### Root Cause
The `custom_deleter` (in `storage_base.cuh`) and `aligned_deleter` (in `bucket_storage.cuh`) were capturing the `Allocator` by reference (`Allocator&`). When a container (e.g., `static_multimap`) was moved, the new container's deleter held a reference to the `allocator_` of the original, destroyed container. 

Subsequent destruction of the new container invoked `deallocate` on a dangling reference. In asynchronous environments like `rapidsmpf`, this corruption resulted in "falsely freed" memory being reused by concurrent CUDA streams, directly causing the observed row count mismatches and flakiness.

### Changes
- Updated `custom_deleter` and `aligned_deleter` to store the `Allocator` by **value** instead of by reference.
- Updated constructors to take `Allocator const&` to ensure proper copying.
- This ensures the allocator's lifetime is correctly tied to the `unique_ptr` that utilizes it, making all core containers safely movable.

## Checklist
- [x] I have verified the fix logic with a standalone C++ reproduction of the move-destruction pattern.
- [x] PR is ready for review.

### DIFF
--- a/include/cuco/bucket_storage.cuh
+++ b/include/cuco/bucket_storage.cuh
@@ -252,10 +252,10 @@ class bucket_storage {
   struct aligned_deleter {
     value_type* raw_ptr_;
     std::size_t size_;
-    allocator_type& allocator_;
+    allocator_type allocator_;
     cuda::stream_ref stream_;
 
-    void operator()(value_type*) const { allocator_.deallocate(raw_ptr_, size_, stream_); }
+    void operator()(value_type*) { allocator_.deallocate(raw_ptr_, size_, stream_); }
   };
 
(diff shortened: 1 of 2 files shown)


PointKernel (approved):
@ksapru Thanks for the fix.

I've edited the PR description as this work will not resolve #805.

PR #808, opened 2026-04-30
Parametric Sectorized Bloom filter policy

sleeepyjack:
Reworks `cuco::bloom_filter` around a parametric Sectorized Bloom Filter (SBF), following "Optimizing Bloom Filters for Modern GPU Architectures" (arXiv:2512.15595).

[Benchmark results](https://github.com/NVIDIA/cuCollections/pull/808#issuecomment-4794402051)


sleeepyjack:
/ok to test [20be4e3](https://github.com/NVIDIA/cuCollections/pull/808/commits/20be4e3c5b2232d731556b246a0f17b5d9ea4454)

benchmarks/bloom_filter/add_bench.cu
```diff
@@ -35,132 +38,235 @@ using namespace cuco::benchmark;  // defaults, dist_from_state, rebind_hasher_t,
...
-  using size_type   = std::uint32_t;
-  using policy_type = cuco::default_filter_policy<rebind_hasher_t<Hash, Key>,
+template <bool ExcludeIO,
+          typename Key,
+          typename Word,
+          nvbench::int32_t BlockBits,
+          nvbench::int32_t PatternBits,
+          nvbench::int32_t HorizontalLayout,
+          nvbench::int32_t VerticalLayout>
+void pfp_bloom_filter_add_impl(nvbench::state& state,
```
sleeepyjack: We don't need the `pfp_` prefix. Applies to all benchmarks

benchmarks/bloom_filter/contains_bench.cu
```diff
@@ -35,146 +39,279 @@ using namespace cuco::benchmark;  // defaults, dist_from_state, rebind_hasher_t,
 using namespace cuco::utility;    // key_generator, distribution
 
 /**
- * @brief A benchmark evaluating `cuco::bloom_filter::contains_async` performance
+ * @brief Implementation of `cuco::bloom_filter::contains_async` with
+ * `parametric_filter_policy`
```
sleeepyjack: no need to mention pfp here or in any other benchmark

include/cuco/detail/bloom_filter/parametric_filter_policy.cuh
```diff
@@ -0,0 +1,412 @@
...
+  // Set bits in the pattern array using salts starting from SaltIndex.
+  template <uint32_t SaltIndex,
+            uint32_t SaltEndIndex,
+            uint32_t PatternArrayIndex,
+            class PatternArrayT>
+  __device__ constexpr void set_bits(uint32_t hash, PatternArrayT& pattern_array) const
+  {
+    if constexpr (SaltIndex < SaltEndIndex) {
+      // Select top bit_index_width bits from salted hash to determine the bit index.
+      // if (threadIdx.x == 0) { printf("Salt Idx: %u\n", SaltIndex); }
```
sleeepyjack: Remove this

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -250,26 +52,43 @@ class bloom_filter_impl {
   using size_type   = typename extent_type::value_type;
   using policy_type = Policy;
   using word_type   = typename policy_type::word_type;
+  // uint64_t may be unsigned long, but atomicOr requires unsigned long long
+  using atomic_word_type = typename cuda::std::
+    conditional_t<cuda::std::is_same_v<word_type, unsigned long>, unsigned long long, word_type>;
```
sleeepyjack: I don't think this check is correct. The filter policy also allows for specifying different word types which would break here.

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -250,26 +52,43 @@ class bloom_filter_impl {
...
+  struct tuning {
+    static constexpr bool use_invoke_one                       = true;
+    static constexpr bool use_early_exit                       = false;
+    static constexpr bool use_cub_kernels                      = true;
+    static constexpr bool use_warp_cooperative_add_kernel      = true;
+    static constexpr bool use_warp_cooperative_contains_kernel = true;
+    static constexpr bool use_work_stealing_add_kernel         = false;
+    static constexpr bool use_work_stealing_contains_kernel    = false;
+    static constexpr bool use_cuda_atomic_ref                  = false;
+  };
```
sleeepyjack: Would be great to not having to touch the source but we can defer this change for now.

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -313,283 +148,519 @@ class bloom_filter_impl {
...
+    auto const num_keys = cuco::detail::distance(first, last);
+    for (decltype(num_keys) i = 0; i < num_keys; ++i) {
+      this->add(*(first + i));
+    }
+  }
+
+  // Device-side range add (cooperative). When the tile size matches `add_horizontal_layout`,
+  // each batch loads/hashes one key per lane in parallel, then the tile cooperatively processes
+  // them via `add_coop`. Otherwise the tile parallelizes across the key range with each lane
+  // scalar-inserting its own keys.
```
sleeepyjack: I don't think the doxygen check will like this but we can fix it later

sleeepyjack (commented):
Self review

PointKernel:
@srinivasyadav18 could use your help reviewing this PR as well

sleeepyjack:
/ok to test [8b1e995](https://github.com/NVIDIA/cuCollections/pull/808/commits/8b1e995bec088c9096ed6f8c41da2ff2458958b4)

sleeepyjack:
/ok to test [eae3049](https://github.com/NVIDIA/cuCollections/pull/808/commits/eae3049a652748d642c9349070ae8decd7941a80)

benchmarks/bloom_filter/add_bench.cu
PointKernel, line 1: @sleeepyjack Could you please share a performance comparison between the baseline and the current implementation? It would be helpful to have those numbers documented for future reference.

benchmarks/bloom_filter/contains_bench.cu
```diff
@@ -24,157 +23,143 @@
...
+//                     nvbench::enum_type_list<64, 128, 256, 512, 1024>,         ///< BlockBits
+//                     nvbench::enum_type_list<8, 16>,                           ///< PatternBits
+//                     nvbench::enum_type_list<1, 2, 4, 8, 16>,                  ///<
+//                     HorizontalLayout nvbench::enum_type_list<1, 2, 4, 8, 16> ///< VerticalLayout
+//                     ))
+//   .set_name("bloom_filter_contains_full_sweep_u64")
+//   .set_type_axes_names(
+//     {"Key", "Word", "BlockBits", "PatternBits", "HorizontalLayout", "VerticalLayout"})
+//   .add_int64_axis("NumInputs", {defaults::BF_N})
+//   .add_int64_axis("FilterSizeMB", defaults::BF_SIZE_MB_RANGE_CACHE);
```
PointKernel, line 167: shall we remove it since unused?

include/cuco/detail/bloom_filter/arrow_filter_policy.cuh
PointKernel, line 1: nice cleanup

include/cuco/bloom_filter.cuh
```diff
@@ -36,13 +36,13 @@
 namespace cuco {
 
 /**
- * @brief A GPU-accelerated Blocked Bloom Filter.
+ * @brief A GPU-accelerated Bloom filter.
```
PointKernel, line 39:
```suggestion
 * @brief A GPU-accelerated Bloom Filter.
```

PointKernel, line 39: It would be helpful to add a brief section describing the underlying algorithm, along with a reference to the original paper. I noticed the paper is referenced in the policy document, but it doesn't appear to be mentioned here.

include/cuco/detail/bloom_filter/parametric_filter_policy.cuh
```diff
@@ -0,0 +1,341 @@
...
+/**
+ * @brief Sectorized Bloom filter policy with multiplicative-hashing fingerprint generation.
+ *
+ * Implements the Sectorized Bloom Filter (SBF) variant from "Optimizing Bloom Filters for Modern
+ * GPU Architectures" (arXiv:2512.15595). Distributes `PatternBits` set bits across `WordsPerBlock`
+ * words using compile-time salt-based multiplicative hashing. The hash result is split into upper
+ * 32 bits (block selection via multiply-shift) and lower 32 bits (pattern generation), so a 64-bit
+ * hash function is required by design.
+ *
+ * @tparam Hash 64-bit hash functor whose return type satisfies `is_same_v<hash_result_type,
```
PointKernel, line 52: optional: we could enforce this with a static assert in ctor

include/cuco/detail/bloom_filter/kernels.cuh
```diff
@@ -15,42 +15,325 @@
...
+  __shared__ uint64_t bar;
+  int phase = 0;
 
-  constexpr auto tile_size = cuco::detail::warp_size();
+  auto const block = cg::this_thread_block();
 
-  auto const tile_idx       = cuco::detail::global_thread_id() / tile_size;
-  auto const n_tiles        = gridDim.x * BlockSize / tile_size;
-  auto const items_per_tile = cuco::detail::int_div_ceil(n, n_tiles);
+  cg::invoke_one(block, [&]() { ptx::mbarrier_init(&bar, 1); });
```
PointKernel: The `invoke_one` issue is only fixed in CTK 13.2 and later. Should we continue using the simple lane-0 implementation until CTK 13.2 becomes the minimum supported CUDA Toolkit version?

And `ptx::mbarrier_init(&bar, 1);` 👀 oh my, what is this

include/cuco/detail/bloom_filter/kernels.cuh
```diff
@@ -15,42 +15,325 @@
...
 
-  auto const tile_start = tile_idx * items_per_tile;
-  if (tile_start >= n) { return; }
-  auto const tile_stop = (tile_start + items_per_tile < n) ? tile_start + items_per_tile : n;
+  int bx = blockIdx.x;
 
-  auto const tile = cg::tiled_partition<tile_size, cg::thread_block>(cg::this_thread_block());
+  // Work-stealing loop:
```
PointKernel: Compared to the naive kernel, how much could we save with this work stealing setup?

include/cuco/detail/bloom_filter/kernels.cuh
```diff
@@ -15,42 +15,325 @@
...
+
+  int bx = blockIdx.x;
+
+  // Work-stealing loop:
+  while (true) {
+    // Protect result from overwrite in the next iteration,
+    // (also ensure barrier initialization at 1st iteration):
+    block.sync();
+
+    cg::invoke_one(block, [&]() {
```
PointKernel: the whole work stealing machinery is shared by both add and contains kernel and could use some dedup effort

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -317,151 +160,218 @@ class bloom_filter_impl {
...
+                                cuda::stream_ref stream)
   {
-    auto const hash_value = policy_.hash(key);
-    this->add_impl(hash_value, policy_.block_index(hash_value, num_blocks_));
+    this->merge_async(other, stream);
+#if CCCL_MAJOR_VERSION > 3 || (CCCL_MAJOR_VERSION == 3 && CCCL_MINOR_VERSION >= 1)
+    stream.sync();
+#else
+    stream.wait();
+#endif
```
PointKernel:
```suggestion
    stream.sync();
```
this should be no longer needed as we fetch CCCL 3.3+ now

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -250,26 +54,47 @@ class bloom_filter_impl {
...
+#else
+      false;  // cg::invoke_one_broadcast requires CTK >= 12.1
+#endif
+    static constexpr bool use_early_exit                       = false;
+    static constexpr bool use_cub_kernels                      = true;
+    static constexpr bool use_warp_cooperative_add_kernel      = true;
+    static constexpr bool use_warp_cooperative_contains_kernel = true;
+    static constexpr bool use_work_stealing_add_kernel         = false;
+    static constexpr bool use_work_stealing_contains_kernel    = false;
+    static constexpr bool use_cuda_atomic_ref                  = false;
```
PointKernel: Is this just a place holder for the future tuning work?

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -250,26 +54,47 @@ class bloom_filter_impl {
   using size_type   = typename extent_type::value_type;
   using policy_type = Policy;
   using word_type   = typename policy_type::word_type;
+  static_assert(sizeof(word_type) == 4 || sizeof(word_type) == 8,
+                "word_type must be 4 or 8 bytes wide for atomicOr");
+  // atomicOr overloads resolve on canonical 32- and 64-bit unsigned integer types.
+  // Normalize by size so any policy-provided word_type (uint32_t, uint64_t, unsigned long, ...)
+  // resolves to a matching overload via the reinterpret_cast in atomic_or().
+  using atomic_word_type =
+    cuda::std::conditional_t<sizeof(word_type) == 8, unsigned long long, unsigned int>;
```
PointKernel, line 63: could we use fixed size types instead?

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -279,26 +104,44 @@ class bloom_filter_impl {
...
     : words_{filter}, num_blocks_{num_blocks}, policy_{policy}
   {
+    NV_IF_TARGET(NV_IS_HOST, (l2_cache_size_ = static_cast<size_t>(cuco::detail::l2_cache_size());))
   }
 
   template <class CG>
   __device__ constexpr void clear(CG group)
   {
-    for (int i = group.thread_rank(); i < num_blocks_ * words_per_block; i += group.size()) {
+    // TODO optimize this
```
PointKernel: Is this TODO still valid

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -317,151 +160,218 @@ class bloom_filter_impl {
...
+    auto const block_index              = policy_.block_index(upper_hash, num_blocks_);
 
-      auto const worker_group  = cg::tiled_partition<worker_num_threads, CG>(group);
-      auto const worker_offset = worker_num_threads * worker_group.meta_group_rank();
+#pragma unroll num_threads
+    for (int i = 0; i < num_threads; ++i) {
+      add_patterns<ConditionalAtomic, 0>(
+        group.shfl(block_index, i), group.shfl(lower_hash, i), group.thread_rank());
+    }
+  }
```
PointKernel, line 240: will `static_for` provide better codegen?

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -317,151 +160,218 @@ class bloom_filter_impl {
...
 
-  template <class CG, class HashValue, class BlockIndex>
-  __device__ void add_impl(CG group, HashValue const& hash_value, BlockIndex block_index)
+  template <class CG, class InputIt>
+  __device__ void add(CG group, InputIt first, InputIt last)
   {
-    constexpr auto num_threads = tile_size_v<CG>;
+    auto num_keys = cuco::detail::distance(first, last);
+    if constexpr (tile_size_v<CG> == add_horizontal_layout && add_horizontal_layout > 1) {
+      auto constexpr num_threads = static_cast<decltype(num_keys)>(tile_size_v<CG>);
```
PointKernel: could we use an explicit type instead of `decltype` throughout this func?

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -636,107 +602,201 @@ class bloom_filter_impl {
...
-      [] __device__(word_type a, word_type b) { return a | b; },
-      stream.get()));
-  }
+  // TODO
+  // [[nodiscard]] __host__ double occupancy() const;
+  // [[nodiscard]] __host__ double expected_false_positive_rate(size_t unique_keys) const
+  // [[nodiscard]] __host__ __device__ static uint32_t optimal_pattern_bits(size_t num_blocks)
+  // template <typename CG, cuda::thread_scope NewScope = thread_scope>
+  // [[nodiscard]] __device__ constexpr auto make_copy(CG group, word_type* const
+  // memory_to_use, cuda_thread_scope<NewScope> scope = {}) const noexcept;
```
PointKernel: still relevant?

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -636,107 +602,201 @@ class bloom_filter_impl {
...
-#else
-    stream.wait();
-#endif
+    // The block storage is aligned to `alignment()`, but a per-lane load at offset `index` is
+    // only guaranteed to be aligned to `min(NumWords * sizeof(word_type), alignment())`. Hand the
+    // compiler the alignment that's actually delivered, not the block-level maximum.
+    constexpr auto load_alignment =
+      cuda::std::min<size_t>(NumWords * sizeof(word_type), alignment());
+    return *reinterpret_cast<cuda::std::array<word_type, NumWords>*>(
+      __builtin_assume_aligned(words_ + index, load_alignment));
```
PointKernel, line 505: IIRC, there is an CCCL util we could use for the vector load?

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
```diff
@@ -636,107 +602,201 @@ class bloom_filter_impl {
...
 
-  // TODO
-  // [[nodiscard]] __host__ double occupancy() const;
-  // [[nodiscard]] __host__ double expected_false_positive_rate(size_t unique_keys) const
-  // [[nodiscard]] __host__ __device__ static uint32_t optimal_pattern_bits(size_t num_blocks)
-  // template <typename CG, cuda::thread_scope NewScope = thread_scope>
-  // [[nodiscard]] __device__ constexpr auto make_copy(CG group, word_type* const
-  // memory_to_use, cuda_thread_scope<NewScope> scope = {}) const noexcept;
+  template <bool ConditionalAtomic>
+  __device__ constexpr void atomic_or(word_type* word_ptr, word_type pattern) const
```
PointKernel, line 553: @sleeepyjack could you elaborate a bit on why we need this custom `atomic_or`?

PointKernel (commented):
The actual code changes are not that big, but the use of work stealing definitely caught my attention. 

@sleeepyjack, could you please review all files touched by this PR and make sure the copyright years are updated where necessary?

benchmarks/bloom_filter/add_bench.cu
sleeepyjack: I'm working on a small ablation study testing all those different tuning knobs. This will also help answer some of your other comments.

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
sleeepyjack: During development we found that `cuda::atomic_ref::fetch_or` sometimes leads to suboptimal codegen so we added a tuning flag to switch between the CCCL atomics and the plain CUDA `atomicOr`. This function is the wrapper around that tuning knob.

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
sleeepyjack: Yes. Although, I would defer this to a follow-up PR so we can measure it in isolation and confirm there's no regression before changing it.

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
sleeepyjack: Possibly. The loop is already compile-time unrolled via the recursive template, so any static_for change is a codegen tradeoff rather than a correctness fix. I'd rather defer this to a follow-up PR so we can measure it in isolation and confirm there's no regression before changing it.

benchmarks/bloom_filter/contains_bench.cu
sleeepyjack: I'm thinking about adding a flag to enable more extensive benchmarks, since compile- and runtime for these setups can be quite long. Maybe in a follow-up PR?

sleeepyjack:
Regarding the tuning knobs, I (or better Codex) did an ablation study:

## Bloom filter tuning sweep summary

I ran a tuning sweep on `sleeepyjack/bloom-filter-release` / PR #808 head
`c571c33f34631fcfa05b0807188a5c12bbfda617` in the default cuCollections
devcontainer (`cuda13.1-gcc14`, CTK 13.1) on:

- GPU: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition
- L2 cache: 128 MiB
- Characterization sizes: `{1, 16, 32, 64, 128, 256, 512}` MiB
- Metrics collected: throughput, time, DRAM throughput, L2 hit rate, FPR
- FPR stayed stable at ~0.00131 for contains variants, so the differences below are performance-only.

### Overall recommendation

The only clear default change suggested by this run is:

- For vertical `contains`, prefer the custom kernel path over the CUB path:
  - `use_cub_kernels = false`

Everything else should stay at the current default unless we want a very small size-specific horizontal-contains optimization.

### Per-knob findings

#### `use_warp_cooperative_add_kernel`

Recommendation: keep enabled.

This is strongly beneficial for `add`.

Screen-stage main effect:

- `on` vs `off`: `+77.1%`

When disabled, add throughput dropped substantially at the screen sizes. This knob should remain enabled.

#### `use_cuda_atomic_ref`

Recommendation: keep disabled.

Enabling `cuda::atomic_ref` hurt `add`.

Characterization, compared to the matching non-`cuda_atomic_ref` path:

- Average: `-3.44%`
- `<= 128 MiB`: about `-3.6%`
- `256–512 MiB`: about `-0.16%`

So the current non-`cuda::atomic_ref` path is better.

#### `use_invoke_one`

Recommendation: neutral; current default is fine.

For the useful add/horizontal-contains configurations, toggling `invoke_one` was essentially noise-level.

Observed deltas:

- Add with warp-coop on and `cuda_atomic_ref=0`: `-0.02%`
- Horizontal contains with warp-coop on: about `-0.02%`

It does not look like a meaningful tuning lever for this workload.

#### `use_cub_kernels`

Recommendation: disable for vertical `contains`.

This was the largest actionable win.

Vertical contains, `cub_kernels=false` vs baseline `cub_kernels=true`:

- Average: `+14.5%`
- `<= 128 MiB`: `+15–16%`
- `256–512 MiB`: `+1.2–1.5%`

Average throughput:

- Baseline vertical contains: `61.29 G elem/s`
- `cub_kernels=false, early_exit=false`: `70.19 G elem/s`

This suggests the custom vertical contains kernel should be preferred over the CUB transform path on this setup.

#### `use_early_exit`

Recommendation: keep disabled.

Early exit did not help in this sweep.

With `cub_kernels=false`:

- `early_exit=true` was slightly worse overall: about `-0.1%`

With `cub_kernels=true`:

- Difference was noise-level, about `+0.02%`

No reason to enable it by default from these data.

#### `use_warp_cooperative_contains_kernel`

Recommendation: keep enabled.

This is critical for horizontal contains.

Horizontal contains with warp-coop disabled:

- Average: `-42.2%`
- `<= 128 MiB`: about `-47.2%`
- `256–512 MiB`: about `-4.2%`

Disabling this increased L2 hit rate somewhat but throughput collapsed, so it should remain enabled.

#### `use_work_stealing_add_kernel`

Recommendation: keep disabled.

Work stealing hurt add throughput.

`work_stealing_add=true` vs baseline add:

- Average: `-1.76%`
- `<= 128 MiB`: `-1.82%`
- `256–512 MiB`: `-0.30%`

Per-size deltas for `{1,16,32,64,128,256,512}` MiB:

- `-1.97%, -2.74%, -2.84%, -0.19%, -0.92%, -0.35%, -0.24%`

No evidence this helps add on the tested workload.

#### `use_work_stealing_contains_kernel`

Recommendation: generally keep disabled; possibly consider only for small horizontal contains.

Vertical contains with `cub_kernels=false`:

- `work_stealing=true, early_exit=false`: `-0.69%` average
- `work_stealing=true, early_exit=true`: `-0.44%` average

So work stealing does not help vertical contains.

Horizontal contains with warp-coop enabled:

- Average with `invoke_one=true`: `+0.78%`
- `<= 128 MiB`: `+0.90%`
- `256–512 MiB`: `-0.13%`

Per-size deltas for `{1,16,32,64,128,256,512}` MiB:

- `+2.02%, +1.15%, +0.63%, +0.54%, -0.28%, -0.16%, -0.11%`

So work stealing has a small niche benefit for small/cache-resident horizontal contains, but it turns slightly negative for larger filters. I would not enable it globally.

### Suggested default policy from this run

For the tested Blackwell + CTK 13.1 setup:

```cpp
use_invoke_one                       = true;  // neutral / current default OK
use_early_exit                       = false;
use_cub_kernels                      = false; // for vertical contains; biggest win
use_warp_cooperative_add_kernel      = true;
use_warp_cooperative_contains_kernel = true;
use_work_stealing_add_kernel         = false;
use_work_stealing_contains_kernel    = false; // unless specializing small horizontal contains
use_cuda_atomic_ref                  = false;
```

### Code paths that could be removed if we want to simplify

If the goal is to keep the implementation lean rather than preserve all experimental tuning paths, the sweep suggests the following removal candidates.

#### Good removal candidates

- `use_cuda_atomic_ref` alternate path:
  - The `cuda::atomic_ref` path was consistently slower for `add`.
  - Suggested simplification: remove the `use_cuda_atomic_ref` knob and keep the current faster non-`cuda::atomic_ref` atomic OR path only.

- `use_early_exit` path:
  - Early exit did not improve vertical contains and was slightly negative with the recommended `cub_kernels=false` path.
  - Suggested simplification: remove the `use_early_exit` knob and simplify the compare recursion to always evaluate the full pattern.

- `use_work_stealing_add_kernel` path:
  - Work stealing was slower for add at every characterized size.
  - Suggested simplification: remove `add_work_stealing_n_impl`, `add_work_stealing_n`, and the host-side launch branch guarded by `use_work_stealing_add_kernel`.

- Vertical-contains use of `use_work_stealing_contains_kernel`:
  - Work stealing was slower with the recommended vertical contains path (`cub_kernels=false`).
  - Suggested simplification: do not route vertical contains through the work-stealing kernel.

- `use_cub_kernels` contains path:
  - The CUB `DeviceTransform` path for vertical contains was substantially slower than the custom contains kernel.
  - Suggested simplification: remove the `use_cub_kernels` knob for contains and always use the custom contains kernel.
  - Note: this does **not** imply removing CUB from the bloom filter implementation entirely, because other operations still use CUB utilities.

#### Probably keep, or remove only with care

- `use_warp_cooperative_add_kernel`:
  - Keep this path. It is strongly beneficial.
  - If simplifying, remove the non-warp-cooperative add path instead, not the warp-cooperative path.

- `use_warp_cooperative_contains_kernel`:
  - Keep this path. It is critical for horizontal contains.
  - If simplifying, remove the non-warp-cooperative horizontal contains path instead, not the warp-cooperative path.

- `use_invoke_one`:
  - Performance impact was effectively neutral.
  - I would keep this as a compatibility/implementation knob unless we want to simplify aggressively.
  - If removing it, keep the current default behavior (`invoke_one` enabled when available) and remove the fallback branch only where toolkit support guarantees it.

- Horizontal `use_work_stealing_contains_kernel`:
  - This has a small benefit for small/cache-resident horizontal contains (`+0.5–2.0%` up to 64 MiB, `+0.9%` through 128 MiB), but turns slightly negative for larger filters.
  - I would not enable it by default. Removal is reasonable if we do not want a size-specialized path; otherwise keep it only behind a size/architecture heuristic.

### Caveat

These results are from a uniform benchmark workload on one Blackwell GPU. The clearest and most robust conclusions are:

1. keep warp-cooperative add/contains enabled,
2. keep `cuda_atomic_ref` disabled,
3. disable CUB for vertical contains,
4. keep work stealing disabled by default.

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
PointKernel: Any CCCL issue we could track this down?

sleeepyjack:
tl;dr here is a summary and my suggestion on what we should do with each tuning knob/ code path:

- `use_invoke_one`: No performance benefit in using it so we can safely remove any use of `invoke_one` as well as the tuning knob.
- `use_early_exit`: Although it didn't show any perf impact in the tested scenarios, it might become relevant for when the match rate of the filter is very low. Kevin had a usecase that was showing significant improvements from this knob. We should even expose this as a tparam of the policy instead of hiding it inside the implementation.
- `use_cub_kernels`: I'd suggest we keep this knob for now, set it to `false` for `contains` and `true` for `add`.
- `use_warp_cooperative*`: Keep this code path. Remove the tuning knobs and make warp cooperative kernels the default.
- `use_work_stealing*`: Not much benefit with the tested configuration but in the paper there were some scenarios on B200 where it was slightly better. I'd say it's not worth the complexity keeping it in our codebase right now. Instead, I would rather wait for a neat CCCL abstraction and then reintroduce it.
- `use_atomic_ref`: `atomicOr` is consistently faster than `cuda::atomic_ref::fetch_or` so I suggest to remove this knob and use `atomicOr` as default. This is fine since we control the type of the underlying atomic aka `word_type`.

What do you think?

PointKernel:
> tl;dr here is a summary and my suggestion on what we should do with each tuning knob/ code path:

Too late. I've already gone through the whole lengthy AI-generated report. 😉 

> What do you think?

All looks valid to me. Several points:

- `use_cub_kernels`: this is surprising. Worth bringing this up to CCCL?
- `use_atomic_ref`: this aligns with my observations duing the new hash table design as well, should report it to CCCL
- `use_work_stealing`: Bloom filter operations tend to have fairly uniform costs compared to hash table operations, so I don't expect work stealing to provide much benefit here. Still, it was great to explore.

sleeepyjack:
## Benchmark default policy `upstream/dev` vs PR

### Setup

- GPU: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition, L2 = 128 MiB
- Devcontainer: `cuda13.1-gcc14`, CTK 13.1
- Axes: `FilterSizeMB={1,16,32,64,128,256,512}`, `NumInputs=1e9`
- NVBench args for both runs: `--min-samples 10 --min-time 0.05 --max-noise 5 --timeout 120 --throttle-threshold 0`

### Add throughput

Geomean speedup: `+12.85%`. Arithmetic mean: `+13.57%`.

| Filter size | Dev throughput | PR throughput | Speedup | Dev time | PR time | Dev noise | PR noise |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 MiB | 48.78 G elem/s | 63.68 G elem/s | +30.54% | 20.499 ms | 15.703 ms | 3.99% | 2.65% |
| 16 MiB | 47.73 G elem/s | 60.53 G elem/s | +26.81% | 20.951 ms | 16.522 ms | 4.01% | 4.40% |
| 32 MiB | 47.73 G elem/s | 59.99 G elem/s | +25.69% | 20.952 ms | 16.669 ms | 3.81% | 4.30% |
| 64 MiB | 36.43 G elem/s | 40.64 G elem/s | +11.55% | 27.453 ms | 24.609 ms | 1.03% | 0.17% |
| 128 MiB | 8.82 G elem/s | 8.81 G elem/s | -0.05% | 113.404 ms | 113.460 ms | 0.06% | 0.03% |
| 256 MiB | 5.36 G elem/s | 5.38 G elem/s | +0.38% | 186.582 ms | 185.872 ms | 0.02% | 0.03% |
| 512 MiB | 4.77 G elem/s | 4.77 G elem/s | +0.09% | 209.773 ms | 209.591 ms | 0.01% | 0.01% |

### Contains throughput and FPR

Geomean throughput speedup: `+17.77%`. Arithmetic mean: `+18.94%`.

| Filter size | Dev throughput | PR throughput | Speedup | Dev FPR | PR FPR | FPR reduction | Dev time | PR time | Dev noise | PR noise |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 MiB | 78.39 G elem/s | 105.98 G elem/s | +35.20% | 0.013022 | 0.001308 | 10.0x | 12.757 ms | 9.436 ms | 4.13% | 2.19% |
| 16 MiB | 73.37 G elem/s | 97.73 G elem/s | +33.19% | 0.029271 | 0.001313 | 22.3x | 13.629 ms | 10.232 ms | 4.93% | 3.53% |
| 32 MiB | 73.23 G elem/s | 97.05 G elem/s | +32.54% | 0.029418 | 0.001315 | 22.4x | 13.656 ms | 10.304 ms | 4.98% | 3.51% |
| 64 MiB | 73.01 G elem/s | 96.07 G elem/s | +31.59% | 0.046045 | 0.001315 | 35.0x | 13.697 ms | 10.409 ms | 5.00% | 3.42% |
| 128 MiB | 38.73 G elem/s | 38.74 G elem/s | +0.03% | 0.060881 | 0.001316 | 46.3x | 25.819 ms | 25.812 ms | 0.05% | 0.06% |
| 256 MiB | 18.81 G elem/s | 18.82 G elem/s | +0.04% | 0.067084 | 0.001316 | 51.0x | 53.167 ms | 53.149 ms | 0.01% | 0.01% |
| 512 MiB | 16.41 G elem/s | 16.41 G elem/s | +0.02% | 0.067727 | 0.001315 | 51.5x | 60.940 ms | 60.930 ms | 0.01% | 0.01% |


**Note:** The huge FPR gap comes from a problem with the old algorithm. Block indexing and fingerprint generation reuse the same hash bits. As the filter gets larger, more bits are consumed by the block index, leaving fewer effective bits for the fingerprint, which inflates FPR. The new algorithm splits those bits, enabling orders-of-magnitude smaller filters at comparable FPR — likely small enough to fit in L2 and run faster.

sleeepyjack:
/ok to test [cc6675b](https://github.com/NVIDIA/cuCollections/pull/808/commits/cc6675be12610407ce008d7e1673a039fbbad260)

include/cuco/detail/bloom_filter/bloom_filter_impl.cuh
PointKernel: Is `unsigned long long` required here because it's the type supported by atomicOr?

PointKernel (approved):
One small but not blocking. Let's ship it.

PR #810, opened 2026-05-01
Revert PR #580 streaming workaround (CCCL #1422 resolved)

PointKernel:
This PR reverts the #580 streaming workaround as large size type is now supported by CUB.

### DIFF
--- a/include/cuco/detail/open_addressing/open_addressing_impl.cuh
+++ b/include/cuco/detail/open_addressing/open_addressing_impl.cuh
@@ -829,60 +829,48 @@ class open_addressing_impl : private open_addressing_compatible<Key, Value, Prob
     using temp_allocator_type =
       typename std::allocator_traits<allocator_type>::template rebind_alloc<char>;
 
-    cuco::detail::index_type constexpr stride = std::numeric_limits<std::int32_t>::max();
-
-    cuco::detail::index_type h_num_out{0};
     auto temp_allocator = temp_allocator_type{this->allocator()};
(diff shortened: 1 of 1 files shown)


include/cuco/detail/open_addressing/open_addressing_impl.cuh
```diff
@@ -829,60 +829,48 @@ class open_addressing_impl : private open_addressing_compatible<Key, Value, Prob
...
+    auto d_temp_storage = temp_allocator.allocate(temp_storage_bytes, stream);
+
+    CUCO_CUDA_TRY(cub::DeviceSelect::If(d_temp_storage,
+                                        temp_storage_bytes,
+                                        begin,
+                                        output_begin,
+                                        d_num_out,
+                                        this->capacity(),
+                                        is_filled,
+                                        stream.get()));
```
sleeepyjack, line 863: nit: Instead of the classical two-phase approach, we could use CUB's new single-phase API that takes an allocator: https://nvidia.github.io/cccl/unstable/cub/api_docs/device_wide.html#environment-api-single-phase

sleeepyjack (approved):

include/cuco/detail/open_addressing/open_addressing_impl.cuh
PointKernel: Just confirmed: the feature is in the CCCL main, but it isn’t included in the 3.3.0 version we’re fetching via `rapids-cmake` yet.

PR #811, opened 2026-05-01
Replace custom size kernel with cub::DeviceReduce::TransformReduce

PointKernel:
This PR replaces the custom size kernel with the corresponding cub util and the performance has been improved significantly with large capacities.

Microbenchmark timing of `static_map::size()` only. NVIDIA GH200 480GB, CUDA 13.1 / GCC 14.3. Map populated with `NumInputs` unique I64 keys at 0.5 occupancy (so capacity = 2 × NumInputs).  
   
  | Capacity (slots) | Before (custom kernel) | After (`cub::DeviceReduce::TransformReduce`) | Speedup |                                                                                                       
  |---:|---:|---:|---:|                                     
  |     2,000,000 |     47.03 us |     43.74 us | 1.08x |                                                                                                                                                      
  |    20,000,000 |    232.11 us |    122.31 us | **1.90x** |                                                                                                                                                  
  |   200,000,000 |   2,360 us   |    869.70 us | **2.71x** |
  | 1,000,000,000 |  11,927 us   |  4,212 us    | **2.83x** |

### DIFF
--- a/include/cuco/detail/open_addressing/kernels.cuh
+++ b/include/cuco/detail/open_addressing/kernels.cuh
@@ -707,43 +707,6 @@ CUCO_KERNEL void retrieve(InputProbeIt input_probe,
   }
 }
 
-/**
- * @brief Calculates the number of filled slots for the given bucket storage.
- *
- * @tparam BlockSize Number of threads in each block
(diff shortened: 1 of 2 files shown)


sleeepyjack (approved):
Nice catch!

Issue #804, opened 2026-04-27
[BUG]: atomic_storage_test fails in Release builds on CUDA < 13.1


sleeepyjack:
## Symptom

`tests/static_set/atomic_storage_test` spuriously fails in Release builds on CUDA < 13.1, missing one or more inserts:

```
tests/static_set/atomic_storage_test.cu:74: FAILED:
REQUIRE( count == num_keys )
with expansion:
99999 (0x1869f) == 100000 (0x186a0)
```

Reproduces in CI on CUDA 12.0. Uses cuco::pair<uint32_t, int32_t> (8 B), hitting the packed_cas path.

## Root cause

The test was added in #706 as a minimal reproducer for the CCCL `__cuda_is_local / isspacep.local` miscompile (originally rapidsai/cudf#18584, rapidsai/cudf#18587, NVIDIA/spark-rapids#12586). NVIDIA/cccl#4586 introduced `_CCCL_ATOMIC_UNSAFE_AUTOMATIC_STORAGE` to bypass the broken check, with this auto-define in `libcudacxx/include/cuda/std/__internal/atomic.h`:

```cpp
// Enable bypassing automatic storage checks ... when using CTK 12.2 and below and if NDEBUG is defined.
#ifndef _CCCL_ATOMIC_UNSAFE_AUTOMATIC_STORAGE
#  if _CCCL_CUDACC_BELOW(13, 1) && !defined(NDEBUG)
#    define _CCCL_ATOMIC_UNSAFE_AUTOMATIC_STORAGE
#  endif
#endif
```

With the current code the bypass fires only in debug builds. Cuco CI runs `CMAKE_BUILD_TYPE=Release`, `NDEBUG` is defined, thus the bug shows up.

## Fix

Define the macro for cuco regardless of build mode, but gate it on CTK so it auto-disappears once the upstream nvcc bug is fixed in 13.1+. Either at the header level in `include/cuco/detail/__config`:

```cpp
#if !defined(_CCCL_ATOMIC_UNSAFE_AUTOMATIC_STORAGE)             \
    && defined(__CUDACC_VER_MAJOR__)                            \
    && ((__CUDACC_VER_MAJOR__ < 13)                             \
        || (__CUDACC_VER_MAJOR__ == 13 && __CUDACC_VER_MINOR__ < 1))
#  define _CCCL_ATOMIC_UNSAFE_AUTOMATIC_STORAGE
#endif
```

## Reproduction

```
./ci/build.sh -tests
[...]

PR #812, opened 2026-05-01
Add small key type tests for static_set and static_map::find

ksapru:
Extends the `static_set` unit-test type matrices to cover sub-4-byte key types (`uint8_t` and `uint16_t`), as requested in the good first issue.

**Changes:**
- Added `uint8_t` and `uint16_t` to the `TEMPLATE_TEST_CASE_SIG` type axes in all 8 templated `static_set` test files
- Added a compile-time `num_keys` guard (`sizeof(Key) == 1 ? 100 : 400`) to keep key sequences within each type's representable range, leaving room for the `0xFF` / `0xFFFF` empty sentinel
- Adjusted hardcoded `gold_capacity` values in `unique_sequence_test` and `retrieve_all_test` to be conditional on `num_keys`
- Parameterized expected sum values in `for_each_test` to match the reduced key count for `uint8_t`

The underlying implementation (`packed_cas` via `cuda::atomic_ref`) already supports 1B and 2B atomic operations — these tests verify that correctness holds end-to-end for small key types.

Part of #805


sleeepyjack:
CI passes. Nice work! I'll give it a review on Monday or Tuesday depending on my workload. In the meantime, we can prepare follow up PRs: (1) replicate the same test extensions to all data structures, also sparsely testing combinations of differenty sized key/value types (e.g. 1B keys and 4B payloads), (2) go over the docs, readme, and asserts to make sure they reflect the new type constraints.

ksapru:
Thanks @sleeepyjack. Is there an easier way to reproduce CI locally in the future?

tests/static_set/atomic_storage_test.cu
```diff
@@ -69,6 +69,7 @@ TEST_CASE("atomic_storage_test", "")
                                                   cuda::proclaim_return_type<Key>(build_fn{}));
 
   set.insert_async(keys_begin, keys_begin + num_keys);
+  CUCO_CUDA_TRY(cudaDeviceSynchronize());
```
PointKernel: question: why is this needed?

include/cuco/detail/pair/helpers.cuh
```diff
@@ -47,6 +47,22 @@ struct packed {
   using type = void;  ///< `void` type by default
 };
 
+/**
+ * @brief Denotes the packed type when the size of the object is 1.
+ */
+template <>
+struct packed<sizeof(uint8_t)> {
+  using type = uint8_t;  ///< Packed type as `uint8_t` if the size of the object is 1
+};
```
PointKernel, line 56: We pack paired data (e.g., a 32-bit key and 32-bit value) into a single 64-bit unit for CAS, used only in map scenarios. Therefore, the smallest CAS we need is 8 bits (1B), and the smallest packed type is 2B, making the 1B overload unnecessary.
```suggestion
```

include/cuco/detail/pair/helpers.cuh
```diff
@@ -47,6 +47,22 @@ struct packed {
...
+
+/**
+ * @brief Denotes the packed type when the size of the object is 2.
+ */
+template <>
+struct packed<sizeof(uint16_t)> {
+  using type = uint16_t;  ///< Packed type as `uint16_t` if the size of the object is 2
+};
```
PointKernel, line 64: could we add a 2B key-value pair in the map unit test to exercise this code path? One test case is sufficient.

PointKernel:
> Thanks @sleeepyjack. Is there an easier way to reproduce CI locally in the future?

@ksapru All the CI environment is shared via [devcontainers](https://github.com/NVIDIA/cuCollections/tree/dev/.devcontainer). You could use them to repro the issue.

sleeepyjack:
> Thanks @sleeepyjack. Is there an easier way to reproduce CI locally in the future?

@ksapru , yes, @PointKernel 's suggestion is the right way to go. Note that if CI fails, the associated test/build log will tell you at the end how to spin up the corresponding devcontainer environment to reproduce the bug. If you don't have access to a GPU system, you can still use the base images of the devcontainers (e.g., [this image](https://github.com/NVIDIA/cuCollections/blob/b748b9d9b1ef87f8e9eeedb1c35227c60bc99991/.devcontainer/cuda12.0-gcc11/devcontainer.json#L3)) directly to at least compile the code locally by running the `ci/build.sh` script.

tests/static_set/atomic_storage_test.cu
sleeepyjack: It shouldn't be required since it is running on the default stream IIUC. It's probably a red herring trying to fix open bug #804.

tests/static_set/for_each_test.cu
```diff
@@ -68,13 +70,19 @@ void test_for_each(Set& set, size_type num_keys)
...
 TEMPLATE_TEST_CASE_SIG(
   "static_set for_each tests",
   "",
   ((typename Key, cuco::test::probe_sequence Probe, int CGSize), Key, Probe, CGSize),
+  (uint8_t, cuco::test::probe_sequence::double_hashing, 1),
+  (uint8_t, cuco::test::probe_sequence::linear_probing, 1),
+  (uint16_t, cuco::test::probe_sequence::double_hashing, 1),
+  (uint16_t, cuco::test::probe_sequence::double_hashing, 2),
+  (uint16_t, cuco::test::probe_sequence::linear_probing, 1),
+  (uint16_t, cuco::test::probe_sequence::linear_probing, 2),
```
sleeepyjack: I wonder if using signed integers (as in the rest of the tests) would make sense for consistency. Plus, it would probably also catch some unintended overflow/wrap-around errors. Just a suggestion though - no hard requirement.

include/cuco/detail/pair/helpers.cuh
ksapru: Acknowledged, thanks for the feedback.

tests/static_set/for_each_test.cu
ksapru: Added them, I think this is a good addition for consistency as well. Thanks Daniel.

sleeepyjack:
/ok to test [9e5899f](https://github.com/NVIDIA/cuCollections/pull/812/commits/9e5899fbad302bf49a4106db5bbc90f797727b0b)

sleeepyjack:
@ksapru do you want to extend the tests for the other open addressing based data structures in this PR or in a follow up PR?

PointKernel (approved):
I’ve cleaned up the code to avoid unnecessary CI iterations. In the process, I realized my previous statement was incorrect: the `packed<1>` code path is actually the desired behavior, as `set` will use that specific path as well. Sorry for the confusion, and thanks again for the contribution @ksapru.

tests/static_set/retrieve_if_test.cu
```diff
@@ -110,6 +112,7 @@ __global__ void test_retrieve_if_all_true_kernel(
 TEMPLATE_TEST_CASE_SIG("static_set retrieve_if",
                        "",
                        ((typename Key), Key),
+                       (int16_t),
```
sleeepyjack, line 116:
```suggestion
                       (int8_t),
                       (int16_t),
```

tests/static_set/shared_memory_test.cu
```diff
@@ -47,6 +48,8 @@ __global__ void shared_memory_test_kernel(Ref* sets,
   auto insert_ref = sets[set_id].make_copy(g, sm_buffer, cuco::thread_scope_block);
   auto find_ref   = insert_ref.rebind_operators(cuco::op::find);
 
+  __syncthreads();
```
sleeepyjack:
```suggestion
  g.sync();
```

tests/static_set/shared_memory_test.cu
```diff
@@ -67,6 +70,7 @@ __global__ void shared_memory_test_kernel(Ref* sets,
 TEMPLATE_TEST_CASE_SIG("static_set shared memory tests",
                        "",
                        ((typename Key), Key),
+                       (int16_t),
```
sleeepyjack, line 74:
```suggestion
                       (int8_t),
                       (int16_t),
```

tests/static_set/stream_test.cu
```diff
@@ -30,9 +30,12 @@
 
 #include <catch2/catch_template_test_macros.hpp>
 
+#include <cstdint>
+
 TEMPLATE_TEST_CASE_SIG("static_set: operations on different stream than constructor",
                        "",
                        ((typename Key), Key),
+                       (int16_t),
```
sleeepyjack, line 39:
```suggestion
                       (int8_t),
                       (int16_t),
```

tests/static_map/find_test.cu
```diff
@@ -146,6 +150,10 @@ TEMPLATE_TEST_CASE_SIG(
   (int64_t, int64_t, cuco::test::probe_sequence::double_hashing, 1),
   (int64_t, int32_t, cuco::test::probe_sequence::double_hashing, 2),
   (int64_t, int64_t, cuco::test::probe_sequence::double_hashing, 2),
+  (int8_t, int8_t, cuco::test::probe_sequence::linear_probing, 1),
+  (int8_t, int8_t, cuco::test::probe_sequence::linear_probing, 2),
```
sleeepyjack, line 155:
```suggestion
  (int8_t, int8_t, cuco::test::probe_sequence::linear_probing, 2),
  (int8_t, int16_t, cuco::test::probe_sequence::linear_probing, 2),
```

tests/static_map/find_test.cu
```diff
@@ -138,6 +138,10 @@ TEMPLATE_TEST_CASE_SIG(
    Value,
    Probe,
    CGSize),
+  (int8_t, int8_t, cuco::test::probe_sequence::double_hashing, 1),
+  (int8_t, int8_t, cuco::test::probe_sequence::double_hashing, 2),
```
sleeepyjack, line 142:
```suggestion
  (int8_t, int8_t, cuco::test::probe_sequence::double_hashing, 2),
  (int8_t, int16_t, cuco::test::probe_sequence::double_hashing, 2),
```

Cover non-packable code path

tests/static_set/shared_memory_test.cu
```diff
@@ -75,8 +79,12 @@ TEMPLATE_TEST_CASE_SIG("static_set shared memory tests",
 #endif
 )
 {
-  constexpr std::size_t number_of_sets  = 1000;
-  constexpr std::size_t elements_in_set = 500;
+  // For int16_t: sentinel = -1, so usable key range is -32768..32767 (excluding -1).
+  // For int16_t: sentinel = -1, so usable key range is -32768..32767 (excluding -1).
```
sleeepyjack:
```suggestion
  // For int8_t: sentinel = -1, so usable key range is -128...127 (excluding -1).
  // For int16_t: sentinel = -1, so usable key range is -32768..32767 (excluding -1).
```

sleeepyjack (changes requested):

sleeepyjack:
/ok to test [0d61796](https://github.com/NVIDIA/cuCollections/pull/812/commits/0d6179619a9c84fb2420f3a93f88db5705bb48c4)

sleeepyjack (approved):
Thanks for your contribution @ksapru

ksapru:
Thanks @sleeepyjack and @PointKernel — really appreciate the reviews, fixes, and clarification here.

I’ll handle the broader test extensions and docs/assert updates in follow-up PRs. Thanks again for helping get this across the line!

PR #821, opened 2026-06-10
Enforce east const via clang-format QualifierAlignment

PointKernel:
This PR updates the clang-format file so that east const is now enforced by formatting.


sleeepyjack (approved):
instant approve, lol

PR #826, opened 2026-07-03
Add Bloom filter persisting L2 access

sleeepyjack:
Depends on #827.

- Add optional persisting-L2 Bloom filter accesses using `cuda::associate_access_property`
- Add benchmarks toggles for persisting-L2 access in `add` and `contains`
- Add a documented CUDA runtime example that sets aside L2, runs `add`/`contains`, then resets the set-aside
- Add tests pinned vs. unpinned equivalency


examples/bloom_filter/persisting_l2_example.cu
```diff
@@ -0,0 +1,114 @@
...
+   * and slow unrelated kernels until the persisting cache is reset.
+   *
+   * NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition, 128 MiB L2, 200M input items:
+   *
+   * filter size | add, default | add, persisting L2 | contains, default | contains, persisting L2
+   * ----------- | ------------ | ------------------ | ----------------- | ------------------------
+   * 64 MiB      | 4.880 ms     | 3.405 ms           | 2.444 ms          | 2.427 ms
+   * 80 MiB      | 8.276 ms     | 3.380 ms           | 4.706 ms          | 2.443 ms
+   * 96 MiB      | 11.004 ms    | 3.476 ms           | 4.874 ms          | 2.575 ms
+   * 128 MiB     | 22.602 ms    | 9.097 ms           | 5.545 ms          | 5.371 ms
```
sleeepyjack, line 51: Benchmark results

examples/bloom_filter/persisting_l2_example.cu
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+#include <cuco/bloom_filter.cuh>
+#include <cuco/detail/error.hpp>
+
+#include <thrust/device_vector.h>
+#include <thrust/sequence.h>
+
+#include <cuda_runtime_api.h>
+
```
PointKernel:
```suggestion
```
not needed with nvcc, is it required by clang? I don't recall.

include/cuco/bloom_filter_policy.cuh
```diff
@@ -0,0 +1,84 @@
...
+          std::uint32_t WordsPerBlock            = 32 / sizeof(Word),
+          std::uint32_t PatternBits              = WordsPerBlock,
+          std::uint32_t AddHorizontalLayout      = WordsPerBlock,
+          std::uint32_t AddVerticalLayout        = 1,
+          std::uint32_t ContainsHorizontalLayout = 1,
+          std::uint32_t ContainsVerticalLayout   = WordsPerBlock,
+          bool ConditionalAdd                    = false,
+          bool EarlyExitContains                 = false,
+          bool PersistingL2Access                = false>
+using bloom_filter_policy = detail::bloom_filter_policy<Hash,
```
PointKernel, line 72: oh nice, could we have an alias named `parametric_filter_policy` so that we don't break cudf use cases?

PointKernel: Just to make sure, this won't break existing use cases once we have the `parametric_filter_policy` alias right?

PointKernel (approved):
Looks good. Just some small non-blocking questions

include/cuco/bloom_filter_policy.cuh
sleeepyjack: addressed in #827

astroC86:
this is amazing!

PR #827, opened 2026-07-03
Refactor Bloom filter policy API

sleeepyjack:
- Rename `parametric_filter_policy` to `bloom_filter_policy` and rename the public/detail headers
- Replace `default_filter_policy` with direct use of `cuco::bloom_filter_policy<Key>`
- Make the public policy key-based with an explicit concrete hash parameter
- Add defaults for the word/block shape, add/contains layouts, and tuning booleans


include/cuco/detail/bloom_filter/bloom_filter_policy.cuh
```diff
@@ -68,16 +73,16 @@ namespace cuco::detail {
...
-class parametric_filter_policy {
+          class Word                        = uint32_t,
+          uint32_t WordsPerBlock            = 32 / sizeof(Word),
+          uint32_t PatternBits              = WordsPerBlock,
+          uint32_t AddHorizontalLayout      = WordsPerBlock,
+          uint32_t AddVerticalLayout        = 1,
+          uint32_t ContainsHorizontalLayout = 1,
+          uint32_t ContainsVerticalLayout   = WordsPerBlock,
+          bool ConditionalAdd               = false,
+          bool EarlyExitContains            = false>
```
sleeepyjack: These defaults are duplicated in the public header. Can we have them in the public header only and remove here?

sleeepyjack: Done in 9aa0018321ef1bf7d5cb326c917f3577e403f512 — the detail policy no longer owns template defaults; they live in the public alias.

PointKernel (approved):
Could add an `parametric_filter_policy` alias for smooth cudf transition otherwise looks good to me.

sleeepyjack:
> Could add an parametric_filter_policy alias for smooth cudf transition otherwise looks good to me.

You mean as a shim until cudf adopts the new policy name? Ideally this is a one line change/rename (`parametric_filter_policy -> bloom_filter_policy`) in cudf whenever we decide to bump the cuco version in rapids-cmake again.

PointKernel:
> You mean as a shim until cudf adopts the new policy name? 

Yes

> Ideally this is a one line change/rename

A one-line breaking change is still a breaking change. It's not a big deal, though.

benchmarks/bloom_filter/add_bench.cu
```diff
@@ -67,19 +67,19 @@ void bloom_filter_add(nvbench::state& state,
...
+                                                                          cuco::xxhash_64<Key>,
+                                                                          Word,
+                                                                          words_per_block,
+                                                                          PatternBits,
+                                                                          HorizontalLayout,
+                                                                          VerticalLayout,
+                                                                          contains_horizontal_layout,
+                                                                          contains_vertical_layout,
+                                                                          false,
+                                                                          false>;
```
PointKernel, line 82: Oh, I forgot to mention: this could use strong types to improve readability and help avoid misuse. It's not a change request in this PR and we can revisit it later if we think it's worthwhile.

sleeepyjack: Yes, plain bools in public interface are bad. This policy is going to be the environment/tuning in the cudax port. Ideally we chose a design that is similar to these.

PR #831, opened 2026-08-03
Add bounds-checked roaring bitmap constructor

PointKernel:
This PR adds a `cuda::std::span` overload to `cuco::experimental::roaring_bitmap` and validates serialized bitmap bounds before host-side metadata parsing, allocation, and device copies. The bounded parser validates 32-bit container metadata, offsets, and payload sizes, as well as 64-bit bucket counts and nested bitmaps.

The existing pointer overload remains source-compatible and is documented as deprecated and unchecked.


include/cuco/roaring_bitmap.cuh
mhaseeb123, line 68: Let's deprecate this one until libcudf can switch over to the new constructor and then delete it once we are no longer using it?

PointKernel: exactly

mhaseeb123: Going to open a draft PR in cudf and temporarily point cuco to this PR until it merges

PointKernel: we also need a rapids-cmake PR bumping the cuco version

PR #832, opened 2026-08-12
Stop inspecting `atomic_ref::fetch_or` in Bloom filter checks

PointKernel:
Per [NVIDIA/cccl#10727](https://github.com/NVIDIA/cccl/pull/10727), `bloom_filter_impl` no longer takes the address of `cuda::atomic_ref::fetch_or` to validate `word_type`. That check depends on the exact standard-library function form and breaks when P3323R1 is backported with constrained member templates. Since the Bloom filter uses native `atomicOr`, this PR checks the actual requirement instead: an integral 4- or 8-byte word type.

Validated against the current head of NVIDIA/cccl#10727; all 46 Bloom-filter test cases and 122 assertions pass.


sleeepyjack:
Superseeded by #833

PR #833, opened 2026-08-13
Replace Bloom filter `Word` with `WordBytes`

sleeepyjack:
This PR replaces the user-specified `Word` type in `bloom_filter_policy` with a `WordBytes` parameter supporting 4- and 8-byte words. The policy selects the corresponding native `atomicOr` type internally, preventing users from supplying incompatible word types and removing the associated validation and conversion machinery. By removing the stale `cuda::atomic_ref::fetch_or` inspection, this also fixes the cuCollections compatibility issue exposed by [NVIDIA/cccl#10727](https://github.com/NVIDIA/cccl/pull/10727) and supersedes #832.


include/cuco/bloom_filter_policy.cuh
```diff
@@ -70,53 +70,4 @@ using bloom_filter_policy = detail::bloom_filter_policy<Hash,
...
-          std::uint32_t WordsPerBlock,
-          std::uint32_t PatternBits,
-          std::uint32_t AddHorizontalLayout,
-          std::uint32_t AddVerticalLayout,
-          std::uint32_t ContainsHorizontalLayout,
-          std::uint32_t ContainsVerticalLayout,
-          bool ConditionalAdd,
-          bool EarlyExitContains,
-          bool PersistingL2Access = false>
-using parametric_filter_policy = detail::bloom_filter_policy<Hash,
```
PointKernel, line 110: Let’s do it separately. I just realized that the new filter policy hasn’t landed in `rapids-cmake` yet, and we’ve already started removing the parametric one.

PointKernel (commented):
cuDF is still using the parametric policy because the current `rapids-cmake` fetches a Git tag from one commit before the new filter policy was introduced. Could we keep the existing parametric policy as-is while also exposing the new policy with `WordBytes`?

cuDF CI has been quite fragile recently, so introducing a breaking change could cause additional issues and potentially block the cuco version bump in `rapids-cmake`.

include/cuco/bloom_filter_policy.cuh
sleeepyjack: Done. Sorry I force-pushed by accident.

include/cuco/bloom_filter_policy.cuh
```diff
@@ -107,16 +101,17 @@ template <class Hash,
...
-                                                             AddHorizontalLayout,
-                                                             AddVerticalLayout,
-                                                             ContainsHorizontalLayout,
-                                                             ContainsVerticalLayout,
-                                                             ConditionalAdd,
-                                                             EarlyExitContains,
-                                                             PersistingL2Access>;
+using parametric_filter_policy =
+  detail::bloom_filter_policy<Hash,
+                              static_cast<std::uint32_t>(sizeof(Word)),
```
PointKernel, line 106: nice

PointKernel (approved):
Looks great. Thanks

PR #835, opened 2026-08-19
Delete default constructor and move assignment on hash tables

### LINKED ISSUE #612: [BUG]: Fix issues with hash table constructors and assignment operators.
### Is this a duplicate?

### Type of Bug

Silent Failure

### Describe the bug

In the new hash table data structures, move constructors and move assignment operators are mistakenly allowed. To avoid misuse, we should disable these features along with the default constructors for the same reason.

### How to Reproduce

https://godbolt.org/z/3ss7cTPjj 

### Expected behavior

https://godbolt.org/z/3ss7cTPjj 

The above sample code should produce a build error indicating that the corresponding operator has been deleted.

amarkdotdev:
The owning hash tables (`static_map`, `static_set`, `static_multimap`, `static_multiset`, `dynamic_map`) defaulted both move operations. This deletes the default constructor and move assignment, and keeps move construction.

Move assignment is the hazard: `impl_ = std::move(rhs.impl_)` frees the still-named destination's storage on its construction stream, so refs handed out from it silently dangle with no host side signal. Move construction just transfers a `unique_ptr`, so refs stay valid and downstream code (e.g. cudf `json_tree.cu`) relies on it.

Fixes #612

### DIFF
--- a/include/cuco/dynamic_map.cuh
+++ b/include/cuco/dynamic_map.cuh
@@ -60,7 +60,7 @@ class dynamic_map {
   dynamic_map()                              = delete;
   dynamic_map(dynamic_map const&)            = delete;
   dynamic_map& operator=(dynamic_map const&) = delete;
-  dynamic_map(dynamic_map&&)                 = delete;
+  dynamic_map(dynamic_map&&) noexcept        = default;  ///< Move constructor
   dynamic_map& operator=(dynamic_map&&)      = delete;
   ~dynamic_map()                             = default;
 
--- a/include/cuco/dynamic_map.cuh
+++ b/include/cuco/dynamic_map.cuh
@@ -57,18 +57,12 @@ class dynamic_map {
   using hasher      = typename map_type::hasher;      ///< Hash function type
   using mapped_type = T;                              ///< Payload type
 
+  dynamic_map()                              = delete;
   dynamic_map(dynamic_map const&)            = delete;
   dynamic_map& operator=(dynamic_map const&) = delete;
-
-  dynamic_map(dynamic_map&&) = default;  ///< Move constructor
-
-  /**
-   * @brief Replaces the contents of the container with another container.
-   *
-   * @return Reference of the current map object
-   */
-  dynamic_map& operator=(dynamic_map&&) = default;
-  ~dynamic_map()                        = default;
+  dynamic_map(dynamic_map&&)                 = delete;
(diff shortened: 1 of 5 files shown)


sleeepyjack:
/ok to test [043b2a0](https://github.com/NVIDIA/cuCollections/pull/835/commits/043b2a0aeb0127a048606f43cad3560d0ecd7ae7)

PointKernel (approved):
@amarkdotdev Thanks for the contribution

Approving. Checked cudf: every cuco table is held behind a pimpl `unique_ptr` or as a local passed by reference, so nothing downstream breaks.

Pushed `dynamic_map` to your branch, which the description mentions but the diff missed.

Note `bloom_filter` and `hyperloglog` still default their move ops. Same argument applies, worth a follow-up.

sleeepyjack:
Was #807 considered here? It fixed the dangling allocator references specifically so that core containers are safely movable. Given that the current containers transfer pointer-owned storage without relocating it, what remaining invariant requires deleting their move operations?

PointKernel:
Pushed a change restoring the move constructor, keeping the deleted default ctor and deleted move assignment.

To answer the question directly: for move construction there is no remaining invariant. All five containers hold `std::unique_ptr<impl_type>`, so a move transfers a pointer, `bucket_storage` never relocates, and refs taken beforehand stay valid. #807 removed the last self referential state, so you are right that half of this was already solved.

Move assignment is the exception. `impl_ = std::move(rhs.impl_)` is `reset(release())`, which frees the destination's storage through `aligned_deleter` on its construction stream while the destination stays named and usable, so refs previously handed out from it silently point at freed memory. That is the only operation here with no host side signal, and it is what is worth deleting. This lands on the same shape as `cudf::table` and `cudf::column`.

Deleting the move constructor also broke cudf at `cpp/src/io/json/json_tree.cu:694`, `:704` and `:712`, where 694 and 704 fail even without an explicit `std::move` because copy elision does not reach through the `std::pair` constructor parameter. My earlier approval note was wrong on that point, sorry. Every downstream use I found is move construction and I found no move assignment of a cuco container in any consumer or in our own tests, so this shape costs nobody a line.

Two description fixes for a follow up: `X() = delete` is a no op today since every ctor has two or more non defaulted leading parameters, worth keeping only as a guard against a future default argument, cf. `hyperloglog.cuh:61`. And the `unique_ptr` in `tests/static_map/shared_memory_test.cu:92` documents the absence of move ops, not a hazard from moving.

PR #836, opened 2026-08-20
Fix open-addressing index generation and wraparound

### LINKED ISSUE #834: [BUG]: static_set with a signed extent reads OOB when hash is 0x80000000
### Is this a duplicate?

### Type of Bug

Runtime Error

### Describe the bug

When `cuco::static_set` uses a signed extent type such as `cuco::extent<int32_t>`, a valid hash value of `0x80000000u` can produce a negative initial probe position.

The affected path narrows the hash to the signed extent type and applies `cuda::std::abs`. The narrowed value is `INT_MIN`, whose positive magnitude is not representable by `int32_t`; the signed absolute-value operation therefore has undefined behavior. In the tested CUDA build, the observable result remains `INT_MIN`. `linear_probing` then applies signed remainder, producing a negative slot index.

The public `static_set::insert` call in the reproducer below reads 32 bytes before its slots allocation. It does not depend on cuDF, RMM, or an application framework.

Reproduced with cuCollections revision
[`f517bbb1277753b1852dfd388993383e401eaa38`](https://github.com/NVIDIA/cuCollections/tree/f517bbb1277753b1852dfd388993383e401eaa38).
The two functions involved are unchanged on `dev` as of this report.

## Reproducer
1. Save this as `repro.cu` in a checkout of that revision:

```cuda
// A valid 0x80000000 hash makes static_set probe before its slots allocation
// when the extent type is int32_t.

#include <cuco/static_set.cuh>
#include <cuda/std/functional>
#include <cuda_runtime_api.h>

#include <cstdint>
#include <cstdio>

#define CUDA_TRY(expr)                                                            \
  do {                                                                            \
    auto const status = (expr);                                                   \
    if (status != cudaSuccess) {                                                  \
      std::fprintf(stderr, "%s failed: %s\n", #expr, cudaGetErrorString(status)); \
      return 1;                                                                   \
    }                                                                             \
  } while (false)

struct high_bit_hash {
  std::uint32_t const* value;

  __device__ std::uint32_t operator()(std::int32_t) const noexcept
  {
    // A device load prevents compile-time folding of the INT_MIN case.
    return *value;
  }
};

int main()
{
  using key_type    = std::int32_t;
  using extent_type = cuco::extent<std::int32_t>;
  using probe_type  = cuco::linear_probing<1, high_bit_hash>;
  using set_type    = cuco::static_set<key_type,
                                       extent_type,
                                       cuda::thread_scope_device,
                                       cuda::std::equal_to<key_type>,
                                       probe_type>;

  constexpr std::uint32_t raw_hash = 0x80000000u;
  constexpr key_type key           = 7;

  std::uint32_t* device_hash{};
  key_type* device_key{};
  CUDA_TRY(cudaMalloc(reinterpret_cast<void**>(&device_hash), sizeof(raw_hash)));
  CUDA_TRY(cudaMalloc(reinterpret_cast<void**>(&device_key), sizeof(key)));
  CUDA_TRY(cudaMemcpy(device_hash, &raw_hash, sizeof(raw_hash), cudaMemcpyHostToDevice));
  CUDA_TRY(cudaMemcpy(device_key, &key, sizeof(key), cudaMemcpyHostToDevice));

  // A non-power-of-two capacity is important: INT_MIN % 8 happens to be zero.
  set_type set{extent_type{10},
               cuco::empty_key<key_type>{-1},
               cuda::std::equal_to<key_type>{},
               probe_type{high_bit_hash{device_hash}}};
  std::printf("hash=0x%08x capacity=%d\n", raw_hash, static_cast<int>(set.capacity()));
  std::fflush(stdout);

  set.insert(device_key, device_key + 1);
  CUDA_TRY(cudaDeviceSynchronize());
  return 0;
}
```

2. Build directly against the cuCollections headers:

```bash
nvcc -std=c++20 --expt-extended-lambda -lineinfo -arch=sm_100 \
  -I include repro.cu -o repro
```

Change `sm_100` to the architecture of the test GPU if necessary.

3. Run it. The bare run already fails; Compute Sanitizer is only needed to localize the access:

```text
$ ./repro
hash=0x80000000 capacity=10
terminate called after throwing an instance of 'cuco::cuda_error'
  what():  CUDA error at: include/cuco/detail/storage/counter_storage.cuh97: cudaErrorIllegalAddress an illegal memory access was encountered
Aborted (core dumped)
```

4. Run it under Compute Sanitizer:

```bash
compute-sanitizer \
  --tool memcheck \
  --error-exitcode 86 \
  --destroy-on-device-error context \
  --print-limit 5 \
  ./repro
```

Observed output from the command above:

```text
========= COMPUTE-SANITIZER
hash=0x80000000 capacity=10
========= Invalid __global__ read of size 4 bytes
=========     at cuco::bucket_storage_ref<int, (int)1, cuco::valid_extent<int, (unsigned long)18446744073709551615>>::operator [](int) const+0x3e0 in bucket_storage.inl:76
=========     by thread (0,0,0) in block (0,0,0)
=========     Access to 0x5f3fffffe0 is out of bounds
=========     and is 32 bytes before the nearest allocation at 0x5f40000000 of size 44 bytes
=========         Device Frame: ... open_addressing_ref_impl<...>::insert<int>(T1)+0x330 in open_addressing_ref_impl.cuh:389
=========         Device Frame: ... operator_impl<cuco::op::insert_tag, ...>::insert<int>(T1)+0xa0 in static_set_ref.inl:394
=========         Device Frame: ... cuco::detail::open_addressing_ns::insert_if_n<...>(...)+0xa0 in kernels.cuh:83
...
========= ERROR SUMMARY: 3 errors
```

Compute Sanitizer exits with status 86. The first error is the invalid slot read; the remaining
two are `cudaErrorLaunchFailure` fallout after the CUDA context is destroyed.

## Root-cause analysis

[`include/cuco/detail/utils.cuh`](https://github.com/NVIDIA/cuCollections/blob/f517bbb1277753b1852dfd388993383e401eaa38/include/cuco/detail/utils.cuh#L38-L46)
narrows the hash before taking its signed absolute value:

```cpp
template <typename SizeType, typename HashType>
__host__ __device__ constexpr SizeType to_positive(HashType hash)
{
  if constexpr (cuda::std::is_signed_v<SizeType>) {
    return cuda::std::abs(static_cast<SizeType>(hash));
  } else {
    return static_cast<SizeType>(hash);
  }
}
```

`sanitize_hash` forwards to `to_positive`, and
[`linear_probing::make_iterator`](https://github.com/NVIDIA/cuCollections/blob/f517bbb1277753b1852dfd388993383e401eaa38/include/cuco/detail/probing_scheme/probing_scheme_impl.inl#L107-L115)
then performs the remainder in the signed extent type:

```cpp
size_type const init = cuco::detail::sanitize_hash<size_type>(hash_(probe_key)) %
                       (upper_bound / BucketSize) * BucketSize;
```

Evaluating `cuco::detail::sanitize_hash<int32_t>(0x80000000u)` in a device kernel on this build returns `-2147483648`, confirming that `abs` does not make the value positive.

The reproducer uses only the public `static_set::insert` API. It does not call `cuco::detail::sanitize_hash` directly.

The same signed remainder appears in the cooperative-group `linear_probing::make_iterator` overload and in both `double_hashing::make_iterator` overloads, so those paths are likely affected as well.

### Expected behavior

Every value in a hasher's unsigned result range should produce an in-bounds probe position. The one-key insertion should complete without a Compute Sanitizer error.

### Reproduction link

_No response_

### Operating System

Ubuntu 24.04.1

### nvidia-smi output

GPU: NVIDIA GB200 NVL4, sm_100

### NVCC version

Driver: 580.105.08
CUDA compiler: 13.2.78

### LINKED ISSUE #837: [ENHANCEMENT]: Avoid expensive modulo when advancing open-addressing probe iterators
## Is your feature request related to a problem? Please describe.

`cuco::detail::probing_iterator::operator++` currently advances a probe with:

```cpp
curr_index_ = (curr_index_ + step_size_) % upper_bound_;
```

The plain `%` here is expensive.
For dynamic extents, `upper_bound_` uses `fast_int`, so every additional probe executes the
fast division/remainder path. This is still more expensive than necessary in a hot loop, especially
for unsuccessful lookups and high-occupancy tables where probe chains are longer.

The addition may also overflow the extent's integer type before the remainder is evaluated, even
though the mathematically reduced index is representable.

Use the probing iterator invariants:

```text
0 <= curr_index_ < upper_bound_
0 < step_size_ <= upper_bound_
```

Their sum can cross `upper_bound_` at most once, so the update can be expressed without modulo or a
potentially overflowing addition:

```cpp
auto const remaining = upper_bound_ - curr_index_;
curr_index_ = step_size_ >= remaining ? step_size_ - remaining
                                      : curr_index_ + step_size_;
```

## Additional context

PR #836 contains a prototype of this optimization while fixing the index-safety bug in #834.

A focused NVBench comparison of `upstream/dev` against that PR on an NVIDIA RTX PRO 6000 Blackwell
Max-Q (`sm_120`, CUDA 13.1.115, GCC 14.3, Release, 20 million `int32_t` keys) produced:

| Configuration group | Cases | Faster | Unchanged | Slower |
|---|---:|---:|---:|---:|
| Default extent, double hashing | 15 | 14 | 1 | 0 |
| `extent<int32_t>`, double hashing | 21 | 21 | 0 | 0 |
| `extent<int32_t>`, linear probing | 12 | 9 | 3 | 0 |
| **Total** | **48** | **44** | **4** | **0** |

Representative improvements:

- default-extent `contains` at 0.8 occupancy: 8.9-10.3%;
- default-extent `find` at 0.8 occupancy: 7.6-11.6%;
- `extent<int32_t>` linear-probing lookups at 0.8-0.9 occupancy: 11.2-12.3%;
- insert at 0.9 occupancy: approximately 5%.

The largest improvements occur for high occupancy and non-matching lookups, which execute the most
iterator increments. Low-occupancy linear-probing cases were statistically unchanged.

sleeepyjack:
Closes #834
Closes #837

- make hash-to-index reduction defined across the supported hash and extent ranges
- avoid signed and unsigned overflow when advancing probing iterators
- reject capacities that cannot be represented after stride or prime rounding
- add regression coverage for scalar and cooperative-group linear/double probing and signed/unsigned extent boundaries

### DIFF
--- a/include/cuco/detail/extent/extent.inl
+++ b/include/cuco/detail/extent/extent.inl
@@ -14,9 +14,48 @@
 
 #include <cuda/std/type_traits>
 
+#include <cmath>
 #include <cstdint>
 
 namespace cuco {
+namespace detail {
+
+constexpr std::uint64_t extent_div_ceil(std::uint64_t dividend, std::uint64_t divisor)
+{
+  return dividend / divisor + static_cast<std::uint64_t>(dividend % divisor != 0);
+}
+
+template <typename SizeType>
+constexpr std::uint64_t max_extent_value()
+{
+  static_assert(cuda::std::is_integral_v<SizeType>);
+  static_assert(sizeof(SizeType) <= sizeof(std::uint64_t));
+  return static_cast<std::uint64_t>(cuda::std::numeric_limits<SizeType>::max());
+}
+
+template <typename SizeType, std::size_t N>
+constexpr bool is_static_extent_representable()
+{
+  if constexpr (sizeof(SizeType) > sizeof(std::size_t) ||
+                (sizeof(SizeType) == sizeof(std::size_t) && cuda::std::is_unsigned_v<SizeType>)) {
+    return true;
+  } else {
+    return N <= static_cast<std::size_t>(cuda::std::numeric_limits<SizeType>::max());
+  }
+}
+
+template <typename SizeType>
+constexpr std::uint64_t normalize_extent(SizeType size)
+{
+  if constexpr (cuda::std::is_signed_v<SizeType>) {
+    return size > 0 ? static_cast<std::uint64_t>(size) : 1ull;
+  } else {
+    return size == 0 ? 1ull : static_cast<std::uint64_t>(size);
+  }
+}
+
+}  // namespace detail
+
 template <typename SizeType, std::size_t N>
 struct valid_extent {
   using value_type = SizeType;  ///< Extent value type
@@ -102,17 +141,26 @@ struct valid_extent<SizeType, dynamic_extent> : cuco::utility::fast_int<SizeType
 template <int32_t CGSize, int32_t BucketSize, typename SizeType, std::size_t N>
 [[nodiscard]] auto constexpr make_valid_extent(extent<SizeType, N> ext)
 {
-  auto constexpr stride = CGSize * BucketSize;
-  auto const size       = cuco::detail::int_div_ceil(
-    cuda::std::max(static_cast<SizeType>(ext), static_cast<SizeType>(1)), stride);
+  static_assert(CGSize > 0);
+  static_assert(BucketSize > 0);
+
+  constexpr auto stride     = static_cast<std::uint64_t>(CGSize) * BucketSize;
+  constexpr auto max_groups = detail::max_extent_value<SizeType>() / stride;
 
   if constexpr (N == dynamic_extent) {
-    return valid_extent<SizeType, dynamic_extent>{
-      static_cast<SizeType>(cuco::detail::next_prime(static_cast<std::uint64_t>(size)) * stride)};
+    auto const requested = detail::normalize_extent(static_cast<SizeType>(ext));
+    auto const groups    = detail::extent_div_ceil(requested, stride);
+    auto const prime     = cuco::detail::next_prime(groups, max_groups);
+    if (prime == 0) { CUCO_FAIL("Requested extent exceeds the representable capacity"); }
+    return valid_extent<SizeType, dynamic_extent>{static_cast<SizeType>(prime * stride)};
   } else {
-    return valid_extent<SizeType,
-                        static_cast<std::size_t>(
-                          cuco::detail::next_prime(static_cast<std::uint64_t>(size)) * stride)>{};
+    static_assert(detail::is_static_extent_representable<SizeType, N>(),
+                  "Static extent must be representable by its size type");
+    constexpr auto requested = N == 0 ? 1 : N;
+    constexpr auto groups    = detail::extent_div_ceil(requested, stride);
+    constexpr auto prime     = cuco::detail::next_prime(groups, max_groups);
+    static_assert(prime != 0, "Requested extent exceeds the representable capacity");
+    return valid_extent<SizeType, static_cast<std::size_t>(prime * stride)>{};
   }
 }
 
@@ -130,16 +178,27 @@ template <typename ProbingScheme, typename Storage, typename SizeType, std::size
   if constexpr (cuco::is_double_hashing<ProbingScheme>::value) {
     return make_valid_extent<ProbingScheme::cg_size, Storage::bucket_size, SizeType, N>(ext);
   } else {
-    auto constexpr stride = ProbingScheme::cg_size * Storage::bucket_size;
-    auto const size =
-      cuco::detail::int_div_ceil(
-        cuda::std::max(static_cast<SizeType>(ext), static_cast<SizeType>(1)), stride) +
-      static_cast<SizeType>(ext == 0);
+    static_assert(ProbingScheme::cg_size > 0);
+    static_assert(Storage::bucket_size > 0);
+
+    constexpr auto stride =
+      static_cast<std::uint64_t>(ProbingScheme::cg_size) * Storage::bucket_size;
+    constexpr auto max_groups = detail::max_extent_value<SizeType>() / stride;
 
     if constexpr (N == dynamic_extent) {
-      return valid_extent<SizeType, dynamic_extent>{size * stride};
+      auto const value = static_cast<SizeType>(ext);
+      auto groups      = detail::extent_div_ceil(detail::normalize_extent(value), stride);
+      if (value == 0) { ++groups; }
+      if (groups > max_groups) { CUCO_FAIL("Requested extent exceeds the representable capacity"); }
+      return valid_extent<SizeType, dynamic_extent>{static_cast<SizeType>(groups * stride)};
     } else {
-      return valid_extent<SizeType, size * stride>{};
+      static_assert(detail::is_static_extent_representable<SizeType, N>(),
+                    "Static extent must be representable by its size type");
+      constexpr auto requested = N == 0 ? 1 : N;
+      constexpr auto groups =
+        detail::extent_div_ceil(requested, stride) + static_cast<std::uint64_t>(N == 0);
+      static_assert(groups <= max_groups, "Requested extent exceeds the representable capacity");
+      return valid_extent<SizeType, static_cast<std::size_t>(groups * stride)>{};
     }
   }
 }
@@ -151,8 +210,16 @@ template <typename ProbingScheme, typename Storage, typename SizeType>
   CUCO_EXPECTS(desired_load_factor > 0., "Desired occupancy must be larger than zero");
   CUCO_EXPECTS(desired_load_factor <= 1., "Desired occupancy must be no larger than one");
 
-  auto const temp = cuda::std::ceil(static_cast<double>(SizeType{ext}) / desired_load_factor);
-  if (temp > static_cast<double>(cuda::std::numeric_limits<SizeType>::max())) {
+  auto const value = static_cast<SizeType>(ext);
+  if constexpr (cuda::std::is_signed_v<SizeType>) {
+    if (value <= 0) { return make_valid_extent<ProbingScheme, Storage>(ext); }
+  } else {
+    if (value == 0) { return make_valid_extent<ProbingScheme, Storage>(ext); }
+  }
+
+  auto const temp =
+    std::ceil(static_cast<long double>(value) / static_cast<long double>(desired_load_factor));
+  if (temp > static_cast<long double>(cuda::std::numeric_limits<SizeType>::max())) {
     CUCO_FAIL(
       "Invalid load factor: requested extent divided by load factor exceeds maximum representable "
       "value");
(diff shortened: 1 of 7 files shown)


sleeepyjack:
### Benchmark results

Environment:

- GPU: NVIDIA RTX PRO 6000 Blackwell Max-Q, `sm_120`
- CUDA: 13.1.115
- Host compiler: GCC 14.3
- Build type: Release
- Input: 20 million `int32_t` keys
- Baseline: `upstream/dev` at `4b26118c`
- Candidate: `e6894e31`

The main performance change is the probe iterator advancement. Previously, each collision advanced
the iterator using:

```cpp
(current + step) % upper_bound
```

For dynamic extents, this invokes the `fast_int` division/modulo path on every additional probe.
The new implementation relies on the iterator invariants (`current < upper_bound` and
`step <= upper_bound`) and performs at most one wrap using a comparison and add/subtract.

As a result, the improvement scales with probe-chain length:

- Higher occupancies produce more collisions and therefore larger improvements.
- Non-matching lookups traverse longer probe sequences than successful lookups and benefit most.
- Linear probing shows the largest gains at high occupancy because clustering creates long chains.
- Insert gains are smaller because successful inserts often terminate earlier and include atomic
  insertion costs that are unaffected by this change.
- Capacity validation runs on the host during construction and does not affect timed kernels.

Across the focused matrix:

| Configuration group | Cases | Faster | Unchanged | Slower |
|---|---:|---:|---:|---:|
| Default extent, double hashing | 15 | 14 | 1 | 0 |
| `extent<int32_t>`, double hashing | 21 | 21 | 0 | 0 |
| `extent<int32_t>`, linear probing | 12 | 9 | 3 | 0 |
| **Total** | **48** | **44** | **4** | **0** |

#### Default extent: contains by matching rate

Occupancy is fixed at `0.8`.

| Key | Distribution | NumInputs | Occupancy | MatchingRate | Ref Time | Ref Noise | Cmp Time | Cmp Noise | Diff | %Diff | Status |
|:---:|:---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|:---:|
| I32 | UNIQUE | 20000000 | 0.8 | 0 | 1.377 ms | 0.42% | 1.235 ms | 0.32% | -141.887 us | **-10.30%** | FAST |
| I32 | UNIQUE | 20000000 | 0.8 | 0.5 | 1.140 ms | 0.38% | 1.027 ms | 0.29% | -113.240 us | **-9.93%** | FAST |
| I32 | UNIQUE | 20000000 | 0.8 | 1 | 826.834 us | 0.48% | 753.128 us | 0.44% | -73.706 us | **-8.91%** | FAST |

#### Default extent: find by matching rate

Occupancy is fixed at `0.8`.

| Key | Distribution | NumInputs | Occupancy | MatchingRate | Ref Time | Ref Noise | Cmp Time | Cmp Noise | Diff | %Diff | Status |
|:---:|:---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|:---:|
| I32 | UNIQUE | 20000000 | 0.8 | 0 | 1.442 ms | 0.50% | 1.275 ms | 0.40% | -166.867 us | **-11.57%** | FAST |
| I32 | UNIQUE | 20000000 | 0.8 | 0.5 | 1.261 ms | 0.35% | 1.137 ms | 0.45% | -124.080 us | **-9.84%** | FAST |
| I32 | UNIQUE | 20000000 | 0.8 | 1 | 963.968 us | 0.36% | 890.785 us | 0.34% | -73.183 us | **-7.59%** | FAST |

#### Default extent: insert by occupancy

| Key | Distribution | NumInputs | Occupancy | Ref Time | Ref Noise | Cmp Time | Cmp Noise | Diff | %Diff | Status |
|:---:|:---:|---:|---:|---:|---:|---:|---:|---:|---:|:---:|
| I32 | UNIQUE | 20000000 | 0.9 | 1.577 ms | 0.33% | 1.486 ms | 0.42% | -90.445 us | **-5.74%** | FAST |
| I32 | UNIQUE | 20000000 | 0.8 | 1.536 ms | 1.48% | 1.481 ms | 1.12% | -54.285 us | **-3.53%** | FAST |
| I32 | UNIQUE | 20000000 | 0.5 | 2.829 ms | 0.06% | 2.828 ms | 0.06% | -1.494 us | -0.05% | SAME |

#### Signed extent, double hashing: contains

| NumInputs | Occupancy | MatchingRate | Ref Time | Ref Noise | Cmp Time | Cmp Noise | Diff | %Diff | Status |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|:---:|
| 20000000 | 0.9 | 0.5 | 1.502 ms | 0.41% | 1.406 ms | 0.25% | -96.631 us | **-6.43%** | FAST |
| 20000000 | 0.8 | 0.5 | 1.063 ms | 0.54% | 994.997 us | 0.29% | -67.614 us | **-6.36%** | FAST |
| 20000000 | 0.9 | 0 | 1.966 ms | 0.40% | 1.849 ms | 0.19% | -117.584 us | **-5.98%** | FAST |
| 20000000 | 0.9 | 1 | 858.481 us | 0.44% | 809.869 us | 0.37% | -48.612 us | **-5.66%** | FAST |
| 20000000 | 0.8 | 0 | 1.269 ms | 0.44% | 1.199 ms | 0.29% | -70.282 us | **-5.54%** | FAST |
| 20000000 | 0.8 | 1 | 774.082 us | 0.50% | 732.542 us | 0.50% | -41.540 us | **-5.37%** | FAST |
| 20000000 | 0.5 | 0 | 1.017 ms | 2.13% | 987.446 us | 1.46% | -29.280 us | **-2.88%** | FAST |
| 20000000 | 0.5 | 0.5 | 954.535 us | 0.25% | 928.963 us | 0.39% | -25.571 us | **-2.68%** | FAST |
| 20000000 | 0.5 | 1 | 863.941 us | 0.19% | 851.916 us | 0.22% | -12.025 us | **-1.39%** | FAST |

#### Signed extent, linear probing: representative high-occupancy cases

| Operation | Occupancy | MatchingRate | Ref Time | Cmp Time | %Diff |
|---|---:|---:|---:|---:|---:|
| Contains | 0.9 | 0.5 | 3.985 ms | 3.495 ms | **-12.28%** |
| Contains | 0.9 | 0 | 5.084 ms | 4.469 ms | **-12.10%** |
| Contains | 0.8 | 0 | 1.366 ms | 1.211 ms | **-11.35%** |
| Contains | 0.8 | 0.5 | 1.119 ms | 993.131 us | **-11.23%** |
| Insert | 0.9 | - | 1.526 ms | 1.450 ms | **-4.97%** |

PointKernel (approved):

PR #839, opened 2026-08-29
Add GPU construction for 32-bit Roaring bitmaps from raw indices

sleeepyjack:
- add `from_indices`, `from_sorted_indices`, and `from_sorted_unique_indices` factories for constructing portable 32-bit Roaring bitmaps from device-accessible ranges of raw indices
- use CUB for sorting, deduplication, container discovery, and workspace reuse, with custom array/bitset serialization kernels and one synchronization for exact output sizing
- add `from_serialized` for consistency with the other factories as an alternative to the existing ctor which will be removed in the future, fix empty bitmap parsing and cached references after moves, and add tests, examples, and NVBench coverage

Deferred features:
- no run container support since it is up-to 24x slower during construction and is optional
- additional optimizations for small bitmaps aka run the whole pipeline in a single CTA
- no batch build yet (TBD)


include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
```diff
@@ -266,16 +253,84 @@ class roaring_bitmap_storage<cuda::std::uint32_t, Allocator> {
...
+
+  /**
+   * @brief Move constructor.
+   *
+   * Rebuilds the cached reference because small run-container bitmaps store computed offsets
+   * directly in the metadata object.
+   *
+   * @param other Storage to move from
+   */
+  roaring_bitmap_storage(roaring_bitmap_storage&& other) noexcept
```
PointKernel: This rebuilds `ref_` around the self-referential `container_offsets_` instead of removing it, so every other by-value copy of the ref (the impl copy, the `[*this]` device capture, the 64-bit bucket memcpy) still points into a dead object. Drop `container_offsets_` (compute it in the accessor) and the cached `ref_`, return a fresh `ref_type{data_.get(), metadata_}` from `ref()`, and make these `= default`. The new move test also passes on the base branch, so it does not guard this.

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
```diff
@@ -266,16 +253,84 @@ class roaring_bitmap_storage<cuda::std::uint32_t, Allocator> {
...
+    ref_       = ref_type{data_.get(), metadata_};
+    return *this;
+  }
+
+  /**
+   * @brief Returns a mutable pointer to serialized storage
+   *
+   * @return Pointer to serialized storage
+   */
+  cuda::std::byte* data() noexcept { return data_.get(); }
```
PointKernel: nit: `[[nodiscard]]` on `data()`, `allocator()`, and `ref()`, as in `bucket_storage.cuh`.

include/cuco/roaring_bitmap.cuh
```diff
@@ -49,6 +57,102 @@ class roaring_bitmap {
...
+   *       immediately by work submitted to the same stream; use an explicit dependency before
+   *       accessing it from another stream.
+   *
+   * @param bitmap Pointer to the beginning of the serialized bitmap in host memory
+   * @param alloc Allocator used to allocate device-accessible storage
+   * @param stream CUDA stream used for device memory operations during construction
+   *
+   * @return Bitmap containing a copy of the serialized input
+   */
+  [[nodiscard]] static roaring_bitmap from_serialized(cuda::std::byte const* bitmap,
```
PointKernel, line 76: `from_serialized` is a `uint32_t`-only forward to the ctor it replaces, which already works for `uint64_t`; drop the `static_assert` and construct via `storage_type{bitmap, alloc, stream}` so `roaring_bitmap<uint64_t>` keeps a non-deprecated path. Prefer `cuda::std::span<cuda::std::byte const>` over a bare pointer, as `hyperloglog` does, so `parse_serialized` can bound-check. If the ctor is meant to be deprecated, use `[[deprecated]]`; otherwise drop the word from the note.

include/cuco/roaring_bitmap.cuh
```diff
@@ -19,20 +19,21 @@ namespace cuco::experimental {
...
- * @tparam T Key type. Must be `cuda::std::uint32_t` or `cuda::std::uint64_t`.
+ * @tparam T Index type. Must be `cuda::std::uint32_t` or `cuda::std::uint64_t`.
  * @tparam Allocator Allocator type used to manage device-accessible storage for the serialized
  *                   bytes.
  */
 template <class T, class Allocator = cuco::cuda_allocator<cuda::std::byte>>
 class roaring_bitmap {
  public:
-  using value_type     = T;                                             ///< Key type
+  using value_type     = T;                                             ///< Index type
```
PointKernel, line 36: The rename to indices stops here: the `roaring_bitmap_ref` docs and the build state (`num_keys`, `KeyIt keys`) still say keys, and in `write_roaring_bitmap_header` `key` also means the 16-bit container key. Rename the build state to `num_indices` and `IndexIt indices`, and update the ref docs.

include/cuco/roaring_bitmap.cuh
```diff
@@ -49,6 +57,102 @@ class roaring_bitmap {
...
+   *
+   * @param first Beginning of the sequence of indices
+   * @param last End of the sequence of indices
+   * @param alloc Allocator used for permanent and temporary device storage
+   * @param stream CUDA stream used for device memory operations and kernel launches
+   *
+   * @return Bitmap containing the unique input indices
+   */
+  template <class InputIt>
+  [[nodiscard]] static roaring_bitmap from_indices(InputIt first,
```
PointKernel, line 103: `from_serialized` throws `cuco::logic_error` on a malformed header; document it with `@throw` as `bloom_filter` and `hyperloglog` do.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+#pragma once
+
+#include <cuco/detail/roaring_bitmap/util.cuh>
+#include <cuco/detail/utility/cuda.cuh>
+
+#include <cuda/std/cstddef>
+#include <cuda/std/cstdint>
+#include <cuda/std/memory>
+
+namespace cuco::experimental::detail {
```
PointKernel, line 16: Missing `CUCO_SUPPRESS_KERNEL_WARNINGS` after the namespace; every other kernel header has it.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+
+  auto* const key_cards = bitmap + 2 * sizeof(cuda::std::uint32_t);
+  misaligned_store(key_cards + index * 2 * sizeof(cuda::std::uint16_t), key);
+  misaligned_store(key_cards + (index * 2 + 1) * sizeof(cuda::std::uint16_t), card_minus_one);
+
+  auto* const offsets      = key_cards + state.num_containers * 2 * sizeof(cuda::std::uint16_t);
+  auto const payload_begin = 2 * sizeof(cuda::std::uint32_t) +
+                             static_cast<cuda::std::uint32_t>(state.num_containers) *
+                               (2 * sizeof(cuda::std::uint16_t) + sizeof(cuda::std::uint32_t));
+  reinterpret_cast<cuda::std::uint32_t*>(offsets)[index] = payload_begin + payload_offsets[index];
```
PointKernel: Raw aligned `uint32_t` store while every other store in this file is `misaligned_store`; with a 2-byte-aligned allocator this throws `cudaErrorMisalignedAddress`. Use `misaligned_store` and make the `size_t` to `uint32_t` narrowing explicit.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+   * @return Payload size in bytes, or zero for an unused slot
+   */
+  __device__ cuda::std::uint32_t operator()(cuda::std::int64_t index) const noexcept
+  {
+    using metadata_type = roaring_bitmap_metadata<cuda::std::uint32_t>;
+
+    auto const num_containers = state->num_containers;
+    if (index >= num_containers) { return 0; }
+
+    auto const begin = container_starts[index];
```
PointKernel: The `begin`/`end` container-bounds derivation appears six times, the array/bitset payload-size ternary twice, the no-run header formula three times, and `>> 16` for the container key is now in four places including `roaring_bitmap_impl.cuh`. One `container_bounds()` helper here plus `constexpr` layout helpers on `roaring_bitmap_metadata<uint32_t>` would replace all of them, `from_no_run_build` and `empty_build_state` included.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+    auto const end   = index + 1 < num_containers ? container_starts[index + 1] : state->num_keys;
+    auto const cardinality = static_cast<cuda::std::uint32_t>(end - begin);
+    return cardinality <= metadata_type::max_array_container_card
+             ? cardinality * sizeof(cuda::std::uint16_t)
+             : metadata_type::bitset_container_bytes;
+  }
+};
+
+template <class ContainerStartIt>
+CUCO_KERNEL void compute_container_payload_sizes(cuda::std::uint32_t* payload_sizes,
```
PointKernel: This kernel only materializes `container_payload_size` for the in-place `ExclusiveSum`. Scan `cuda::make_transform_iterator(cuda::counting_iterator<int64_t>{0}, container_payload_size{...})` directly, as `open_addressing_impl.cuh` does, and delete it. Also, none of the five kernels has Doxygen while the two functors above do.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+                                                 roaring_bitmap_build_state const* state)
+{
+  auto const index = cuco::detail::global_thread_id();
+  if (index >= num_container_slots) { return; }
+  // Slots beyond the selected container count contribute zero to the fixed-size scan.
+  payload_sizes[index] = container_payload_size{container_starts, state}(index);
+}
+
+template <class ContainerStartIt, class PayloadOffsetIt>
+CUCO_KERNEL void compute_roaring_bitmap_build_size(roaring_bitmap_build_state* state,
```
PointKernel: A `<<<1,1>>>` launch for work the last container's thread in `collect_container_indexes` can do once `payload_offsets` is final; the `num_containers == 0` branch is unreachable because `build()` short-circuits empty input.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+  } else {
+    // Bitset indexes grow backward so both work lists share one num_container_slots-sized buffer.
+    auto const output_index = atomicAdd(&state->num_bitset_containers, 1);
+    container_indexes[num_container_slots - output_index - 1] =
+      static_cast<cuda::std::uint32_t>(index);
+  }
+}
+
+template <class KeyIt>
+__device__ cuda::std::int64_t lower_bound_low_bits(KeyIt keys,
```
PointKernel: could we use `cuda::std::lower_bound` here?

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+    auto const container_index = static_cast<cuda::std::int64_t>(bitset_containers[bitset_index]);
+    auto const begin           = container_starts[container_index];
+    auto const end             = container_index + 1 < state.num_containers
+                                   ? container_starts[container_index + 1]
+                                   : state.num_keys;
+    auto* const container      = bitmap + payload_begin + payload_offsets[container_index];
+    auto const word_begin      = word * 64;
+    auto const word_end        = word_begin + 64;
+    auto const first           = lower_bound_low_bits(keys, begin, end, word_begin);
+    auto const last            = lower_bound_low_bits(keys, first, end, word_end);
```
PointKernel: Each of the 1024 threads per bitset container runs two full bisections, but `last` of thread t is `first` of thread t+1. Search once per thread, exchange `first` through shared memory, and let only the last thread search `last`.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+                               (2 * sizeof(cuda::std::uint16_t) + sizeof(cuda::std::uint32_t));
+  reinterpret_cast<cuda::std::uint32_t*>(offsets)[index] = payload_begin + payload_offsets[index];
+}
+
+template <cuda::std::uint32_t BlockSize,
+          class KeyIt,
+          class ContainerStartIt,
+          class PayloadOffsetIt,
+          class ContainerIndexIt>
+CUCO_KERNEL void write_roaring_containers(cuda::std::byte* bitmap,
```
PointKernel: could add `__launch_bounds__(BlockSize)` like the other `BlockSize` kernels.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+                                          ContainerStartIt container_starts,
+                                          PayloadOffsetIt payload_offsets,
+                                          ContainerIndexIt array_containers,
+                                          ContainerIndexIt bitset_containers)
+{
+  using metadata_type = roaring_bitmap_metadata<cuda::std::uint32_t>;
+
+  constexpr cuda::std::uint32_t warps_per_block = BlockSize / cuco::detail::warp_size();
+  constexpr cuda::std::uint32_t bitset_words =
+    metadata_type::bitset_container_bytes / sizeof(unsigned long long);
```
PointKernel: `unsigned long long` here but `sizeof(cuda::std::uint64_t)` in the builder for the same word. Add `bitset_word_type = cuda::std::uint64_t` and `bitset_container_words` to the metadata and derive both from it.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+    auto const cardinality = static_cast<cuda::std::uint32_t>(end - begin);
+    return cardinality <= metadata_type::max_array_container_card
+             ? cardinality * sizeof(cuda::std::uint16_t)
+             : metadata_type::bitset_container_bytes;
+  }
+};
+
+template <class ContainerStartIt>
+CUCO_KERNEL void compute_container_payload_sizes(cuda::std::uint32_t* payload_sizes,
+                                                 cuda::std::int64_t num_container_slots,
```
PointKernel: Sibling kernels use `cuco::detail::index_type` for counts; `int64_t` here and `size_t` for `num_container_slots_` force a conversion at every launch.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+  if (index >= state->num_containers || index >= num_container_slots) { return; }
+
+  auto const begin = container_starts[index];
+  auto const end =
+    index + 1 < state->num_containers ? container_starts[index + 1] : state->num_keys;
+  auto const cardinality = static_cast<cuda::std::uint32_t>(end - begin);
+  if (cardinality <= metadata_type::max_array_container_card) {
+    // The queue does not need to preserve container order: writers use the stored container index
+    // to find the final payload offset.
+    auto const output_index         = atomicAdd(&state->num_array_containers, 1);
```
PointKernel, line 140: note: this is the only raw `atomicAdd` in cuco

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+    auto const begin = container_starts[index];
+    auto const end   = index + 1 < num_containers ? container_starts[index + 1] : state->num_keys;
+    auto const cardinality = static_cast<cuda::std::uint32_t>(end - begin);
+    return cardinality <= metadata_type::max_array_container_card
+             ? cardinality * sizeof(cuda::std::uint16_t)
+             : metadata_type::bitset_container_bytes;
+  }
+};
+
+template <class ContainerStartIt>
```
PointKernel: `ContainerStartIt`, `PayloadOffsetIt`, and `ContainerIndexIt` are always raw pointers from `allocate_temporary_buffer`; we could take `int64_t const*` / `uint32_t const*` and keep only `KeyIt` generic.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
```diff
@@ -0,0 +1,282 @@
...
+ */
+
+#pragma once
+
+#include <cuco/detail/roaring_bitmap/util.cuh>
+#include <cuco/detail/utility/cuda.cuh>
+
+#include <cuda/std/cstddef>
+#include <cuda/std/cstdint>
+#include <cuda/std/memory>
```
PointKernel: `<cuda/std/memory>` is unused here.

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
```diff
@@ -0,0 +1,450 @@
...
+      first_, container_starts.get(), payload_offsets.get(), state.get(), workspace.get());
+  }
+
+  template <class SourceIt>
+  [[nodiscard]] sorted_indices_result sort_indices(SourceIt first,
+                                                   cuda::std::uint32_t* indices_a,
+                                                   cuda::std::uint32_t* indices_b,
+                                                   cuda::std::byte* workspace) const
+  {
+    CUCO_CUDA_TRY(cub::DeviceTransform::Transform(
```
PointKernel: For contiguous input this identity copy is about 22% of the sort-phase bytes for nothing; on `thrust::is_contiguous_iterator_v<InputIt>` use the `SortKeys(const KeyT*, KeyT*, ...)` overload on the unwrapped pointer. Note it raises peak temporary memory from 8N to 12N.

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
```diff
@@ -0,0 +1,450 @@
...
+                                  0,
+                                  stream_.get()>>>(
+      storage.data(), first, host_state, container_starts, payload_offsets);
+
+    if (host_state.num_containers > 0) {
+      constexpr cuda::std::uint32_t block_size      = 256;
+      constexpr cuda::std::uint32_t warps_per_block = block_size / 32;
+      constexpr cuda::std::uint32_t bitset_blocks_per_container =
+        metadata_type::bitset_container_bytes / sizeof(cuda::std::uint64_t) / block_size;
+      auto const array_blocks =
```
PointKernel, line 417: could use `cuco::detail::int_div_ceil`

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
```diff
@@ -0,0 +1,450 @@
...
+  }
+
+  [[nodiscard]] storage_type build_sorted_unique()
+  {
+    auto container_starts = allocate_temporary_buffer<cuda::std::int64_t>(num_container_slots_);
+    auto payload_offsets  = allocate_temporary_buffer<cuda::std::uint32_t>(num_container_slots_);
+    auto state            = allocate_temporary_buffer<roaring_bitmap_build_state>(1);
+    auto workspace        = allocate_temporary_buffer<cuda::std::byte>(workspace_bytes_);
+
+    CUCO_CUDA_TRY(cudaMemcpyAsync(&state->num_keys,
```
PointKernel: IIRC, we have a detail `memcpy_async`

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
```diff
@@ -0,0 +1,450 @@
...
+   * @param alloc Allocator used for temporary and final storage
+   * @param stream Stream used for all allocation and build work
+   */
+  explicit roaring_bitmap_builder(InputIt first,
+                                  InputIt last,
+                                  roaring_bitmap_builder_input_order input_order,
+                                  Allocator const& alloc,
+                                  cuda::stream_ref stream)
+    : first_{first},
+      num_indices_{std::max<cuda::std::int64_t>(0, cuco::detail::distance(first, last))},
```
PointKernel: Clamping `distance(first, last)` to zero hides a reversed range; every other cuco bulk API treats that as a precondition violation. Drop the clamp and the reversed-range test.

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
```diff
@@ -0,0 +1,450 @@
...
+
+    if (input_order_ == roaring_bitmap_builder_input_order::unsorted) {
+      required_bytes = 0;
+      cub::DoubleBuffer<cuda::std::uint32_t> indices{index_buffer_a, index_buffer_b};
+      CUCO_CUDA_TRY(cub::DeviceRadixSort::SortKeys(nullptr,
+                                                   required_bytes,
+                                                   indices,
+                                                   num_indices_,
+                                                   0,
+                                                   sizeof(cuda::std::uint32_t) * 8,
```
PointKernel: `sizeof(T) * CHAR_BIT` or `cuda::std::numeric_limits<uint32_t>::digits`, as elsewhere in cuco.

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
```diff
@@ -0,0 +1,450 @@
...
+#include <cuda/std/iterator>
+#include <cuda/std/type_traits>
+#include <cuda/stream_ref>
+
+#include <algorithm>
+#include <memory>
+
+namespace cuco::experimental::detail {
+
+enum class roaring_bitmap_builder_input_order { unsorted, sorted, sorted_unique };
```
PointKernel, line 37: Doc comment for the enum, like the other detail enums.

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
```diff
@@ -0,0 +1,450 @@
...
+    roaring_bitmap_build_state const* state) const
+  {
+    using metadata_type = roaring_bitmap_metadata<cuda::std::uint32_t>;
+
+    roaring_bitmap_build_state host_state{};
+    CUCO_CUDA_TRY(cuco::detail::memcpy_async(
+      &host_state, state, sizeof(host_state), cudaMemcpyDeviceToHost, stream_));
+    // The serialized allocation size depends on device-computed container cardinalities. This is
+    // the only synchronization required before final storage can be allocated.
+#if CCCL_MAJOR_VERSION > 3 || (CCCL_MAJOR_VERSION == 3 && CCCL_MINOR_VERSION >= 1)
```
PointKernel, line 373: Fourth verbatim copy of this CCCL-version sync switch (three in `static_multimap.inl`); a `cuco::detail::stream_sync(cuda::stream_ref)` helper would centralize it.

include/cuco/detail/roaring_bitmap/roaring_bitmap.inl
```diff
@@ -18,6 +23,62 @@ roaring_bitmap<T, Allocator>::roaring_bitmap(cuda::std::byte const* bitmap,
...
+  return roaring_bitmap{bitmap, alloc, stream};
+}
+
+template <class T, class Allocator>
+roaring_bitmap<T, Allocator>::roaring_bitmap(storage_type&& storage)
+  : storage_{cuda::std::move(storage)}
+{
+}
+
+template <class T, class Allocator>
```
PointKernel, line 39: The three factory bodies are identical apart from the enumerator; a `detail::build_roaring_bitmap_storage<T>(first, last, order, alloc, stream)` helper reduces each to one line. Related: why three public factories rather than one with a tag or enum, given the builder is already keyed on `roaring_bitmap_builder_input_order`?

tests/roaring_bitmap/build_test.cu
```diff
@@ -0,0 +1,449 @@
...
+
+  std::vector<index_type> absent;
+  absent.reserve(cardinalities.size());
+  for (cuda::std::size_t container = 0; container < cardinalities.size(); ++container) {
+    absent.push_back((static_cast<index_type>(container) << 16) | cardinalities[container]);
+  }
+  require_contains(bitmap, absent, std::vector<bool>(absent.size(), false));
+}
+
+TEST_CASE("roaring_bitmap selects array and bitset containers at the format threshold",
```
PointKernel, line 289: No test covers index 0 or `0xFFFFFFFF`, key `0xFFFF`, a full container, the 65536-container cap, duplicates straddling a container boundary, a bitset container via `from_sorted_indices`, or a non-empty `from_serialized` round trip; the largest key here is 10. They pass on this head, so these are coverage gaps.

tests/roaring_bitmap/build_test.cu
```diff
@@ -0,0 +1,449 @@
...
+}
+
+TEST_CASE("roaring_bitmap matches RoaringFormatSpec no-run serialization", "[roaring_bitmap]")
+{
+#ifndef CUCO_ROARING_DATA_DIR
+  SKIP(
+    "CUCO_ROARING_DATA_DIR is not defined. Configure with -DCUCO_DOWNLOAD_ROARING_TESTDATA=ON to "
+    "run this test.");
+#else
+  std::vector<index_type> host_indices;
```
PointKernel: Third hand-typed copy of the `bitmapwithoutruns.bin` key set (also `contains_test.cu` and the example); share it from a small test header.

benchmarks/roaring_bitmap/build_bench.cu
```diff
@@ -0,0 +1,145 @@
...
+void roaring_bitmap_build(nvbench::state& state, nvbench::type_list<Dist>)
+{
+  using index_type  = cuda::std::uint32_t;
+  using bitmap_type = cuco::experimental::roaring_bitmap<index_type>;
+
+  auto const num_inputs = state.get_int64("NumInputs");
+  thrust::device_vector<index_type> indices(num_inputs);
+
+  [[maybe_unused]] key_generator generator{};
+  if constexpr (Mode == build_mode::sorted_unique_indices) {
```
PointKernel, line 48: All generic inputs are dense in `[0, N/multiplicity)`, so the sorted factories only ever produce bitset containers; add a container-cardinality axis (for example `{1, 8, 64, 512, 4096, 8192, 32768}`) via the tabulate generator so the array writer is measured too.

include/cuco/detail/roaring_bitmap/roaring_bitmap_storage.cuh
PointKernel: not sure if it's legit review from AI

include/cuco/roaring_bitmap.cuh
sleeepyjack: I've now unblocked the 64-bit code path. 
I would keep the pointer API for now. A span requires callers to know the serialized bitmap size, which they may not have, and without that size we cannot meaningfully bounds-check anyway. This is in line with the decision we've made for now on the related nvbug.

include/cuco/detail/roaring_bitmap/roaring_bitmap_kernels.cuh
sleeepyjack: I would keep atomicAdd here for now. The relevant CCCL atomics API is currently being revised. I would wait for the update rather than mess with codegen.

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
sleeepyjack: Lets open an issue on that: #840

include/cuco/detail/roaring_bitmap/roaring_bitmap.inl
sleeepyjack: I would keep the three named factories. The input ordering guarantee is part of the API, and a public enum or tag makes that precondition easier to misuse. The small wrappers are intentional; the detail builder already centralizes the actual implementation, so I do not think another free helper buys enough to justify another layer. Also, adding another specialized enum to the public API feels unnecessary.

examples/roaring_bitmap/host_bulk_from_serialized_example.cu
```diff
@@ -94,8 +94,14 @@ bool check(std::string const& bitmap_file_path)
   file.close();
 
   // Create roaring bitmap from the file
-  cuco::experimental::roaring_bitmap<KeyType> roaring_bitmap(
-    thrust::raw_pointer_cast(buffer.data()));
+  auto roaring_bitmap = [&] {
+    auto const* data = thrust::raw_pointer_cast(buffer.data());
+    if constexpr (cuda::std::is_same_v<KeyType, cuda::std::uint32_t>) {
```
PointKernel: `from_serialized` now supports `uint64_t` too, so this `if constexpr` can collapse to a single `roaring_bitmap<KeyType>::from_serialized(data)` call.

include/cuco/detail/roaring_bitmap/roaring_bitmap_builder.cuh
```diff
@@ -0,0 +1,453 @@
...
+  [[nodiscard]] storage_type build_sorted_unique()
+  {
+    auto container_starts = allocate_temporary_buffer<cuco::detail::index_type>(
+      static_cast<cuda::std::size_t>(num_container_slots_));
+    auto payload_offsets = allocate_temporary_buffer<cuda::std::uint32_t>(
+      static_cast<cuda::std::size_t>(num_container_slots_));
+    auto state     = allocate_temporary_buffer<roaring_bitmap_build_state>(1);
+    auto workspace = allocate_temporary_buffer<cuda::std::byte>(workspace_bytes_);
+
+    CUCO_CUDA_TRY(cuco::detail::memcpy_async(
```
PointKernel, line 264: This copies from pageable host memory. It is safe only because `read_build_state` syncs the stream before the builder goes out of scope. Worth a one-line comment so nobody moves the sync later.

benchmarks/roaring_bitmap/build_bench.cu
```diff
@@ -0,0 +1,171 @@
...
+void roaring_bitmap_build(nvbench::state& state, nvbench::type_list<Dist>)
+{
+  using index_type = cuda::std::uint32_t;
+
+  auto const num_inputs = state.get_int64("NumInputs");
+  thrust::device_vector<index_type> indices(num_inputs);
+
+  [[maybe_unused]] key_generator generator{};
+  if constexpr (Mode == build_mode::sorted_unique_indices) {
+    thrust::sequence(indices.begin(), indices.end());
```
PointKernel, line 49: nit: this branch ignores `Dist`, so the `Distribution` and `Multiplicity` axes on `roaring_bitmap_from_sorted_unique_indices` are decorative. Same for the `Multiplicity {1}` axis on `roaring_bitmap_from_indices_unique`, which the unique generator ignores.

PointKernel (approved):
All previous points addressed, no new correctness issues. A few nits inline.

PR #841, opened 2026-09-05
Add byte-budget construction for Bloom filters

### LINKED ISSUE #829: Add byte-oriented sizing and validation utilities for `bloom_filter`
## Is your feature request related to a problem? Please describe.

`cuco::bloom_filter` is currently sized only by `num_blocks` (a raw count of filter blocks). Callers that think in terms of a storage budget in bytes, which is the natural unit for memory allocation and for distributed work, have to convert bytes to blocks themselves and re-derive the policy's constraints: a filter size must be a positive multiple of the block size (`words_per_block * sizeof(word_type)`) and no greater than the policy maximum (`max_filter_blocks`).

This came up in [rapidsai/cudf#23067](https://github.com/rapidsai/cudf/pull/23067), where libcudf_streaming's `device_bloom_filter` wrapper re-implements this sizing logic: a byte-sized constructor, an `aligned_size(bytes)` helper that rounds a byte count down to the largest valid filter size, and a `max_size()` accessor for the policy's upper bound. This is general-purpose logic that every cuco bloom filter user needs, not something specific to cudf, so it belongs in cuco with the corresponding validation checks rather than being re-derived by each caller.

Expose byte-oriented sizing utilities directly on `cuco::bloom_filter` / its policy, mirroring the convenience constructors HyperLogLog already provides (`cuco::sketch_size_kb`, `cuco::standard_deviation`). Concretely:

- A way to construct or size a filter from a target storage size in bytes, alongside the existing block-count path.
- An `aligned_size`-style helper that returns the largest valid filter size not exceeding a requested byte count (a positive multiple of the block size, capped at the policy maximum).
- An accessor for the maximum storage size supported by the filter policy.
- The associated validation (positive, multiple of block size, within the policy limit) built into these utilities so callers get consistent error checking.

## Describe alternatives you've considered

Keeping the conversion in each downstream project, as libcudf_streaming does today. This duplicates the block-size and maximum-size rules across users and risks them drifting from the policy's actual constraints.

## Additional context

Origin discussion: https://github.com/rapidsai/cudf/pull/23067#discussion_r3562252707. The relevant policy constants (`words_per_block`, `max_filter_blocks`) live here: https://github.com/NVIDIA/cuCollections/blob/0883368d39296f3bef3a058033141bcc642c5c54/include/cuco/detail/bloom_filter/parametric_filter_policy.cuh#L101-L119

yuweih205:
Fixes #829.

Adds `cuco::bloom_filter_bytes` construction for dynamic-extent Bloom filters. The value is an upper-bound storage budget: allocation is rounded down to whole filter blocks and capped at `max_size()`, so it never exceeds the requested bytes. Budgets smaller than one block fail before allocation.

The existing block-count constructor remains unchanged. Coverage is folded into the existing Bloom filter policy test rather than a separate test target.

Local validation: repository formatting and copyright hooks, plus `git diff --check`.

### DIFF
--- a/include/cuco/bloom_filter.cuh
+++ b/include/cuco/bloom_filter.cuh
@@ -8,6 +8,7 @@
 #include <cuco/bloom_filter_policy.cuh>
 #include <cuco/bloom_filter_ref.cuh>
 #include <cuco/detail/storage/storage_base.cuh>
+#include <cuco/detail/utility/strong_type.cuh>
 #include <cuco/extent.cuh>
 #include <cuco/hash_functions.cuh>
 #include <cuco/utility/allocator.hpp>
@@ -23,6 +24,11 @@
 #include <memory>
 
 namespace cuco {
+
+/**
+ * @brief A strong type wrapper for specifying an exact Bloom filter size in bytes.
+ */
+CUCO_DEFINE_STRONG_TYPE(bloom_filter_size_bytes, std::size_t)
 
 /**
  * @brief A GPU-accelerated Bloom filter.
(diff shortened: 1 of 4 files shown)


sleeepyjack:
Hi @yuweih205,

Thanks for contributing to cuco!
@PointKernel and I discussed the topic offline and settled on a simpler approach:
Instead of allowing the user to go both ways aka provide a lower-bound or upper-bound size in bytes, we only really need the upper-bound case which can be handled through the new ctor overload.

In short, 'bloom_filter(bloom_filter_bytes size)' allocates a filter that is _always smaller or equal_ to 'size'; never exceeding it.
This means we can also remove the 'aligned_size' helper and the extent factory, making the API much simpler.

One small nit: I would suggest to rename the strong type for the filter size to `bloom_filter_bytes' as it is a bit shorter.

Do you agree with these changes? Happy to discuss and thanks again for your work on this!

yuweih205:
Agreed—updated. `bloom_filter_bytes` is now an upper-bound budget: the constructor rounds down to whole blocks, caps at `max_size()`, and is available only for dynamic extents. I removed `aligned_size`, the extent factory, and the standalone sizing test, folding the minimal coverage into the existing policy test.

sleeepyjack:
/ok to test https://github.com/NVIDIA/cuCollections/pull/841/commits/c80884c25614d43320e5bd134daf1c95cede815f

PR #844, opened 2026-09-13
[REVIEW] Add brace-initializer overload to static_map insert and insert_and_find

### LINKED ISSUE #410: [ENHANCEMENT]: Add insert overloads that take `value_type` as input

Firstly proposed by @sleeepyjack in https://github.com/NVIDIA/cuCollections/pull/407#discussion_r1431913449

The current heterogeneous insert feature requires users to pass an explicit pair type when invoking the device insert functions. So the legacy behavior like:
```cuda
  set_ref.insert({key, payload});
```
is no longer supported.

Similar to STL [insert](https://godbolt.org/z/azjc36EG1), it would be nice to add an overload of insert taking `value_type` as input thus the [sample code](https://godbolt.org/z/YhW8E65Wj) can work as well.

MaybeWilli:
Fixes #410.

Added two overloads for static_map insert and insert_and_find, so that legacy functions like
```
  set_ref.insert({key, payload});
```
is supported again.


include/cuco/detail/static_map/static_map_ref.inl
```diff
@@ -445,6 +445,19 @@ class operator_impl<
...
   }
 
+  /**
+   * @brief Inserts an element.
+   *
+   * @param value The element to insert
+   *
+   * @return True if the given element is successfully inserted
+   */
+  __device__ bool insert(value_type value) noexcept
```
PointKernel, line 455: Could we add `value_type` overloads for cooperative `insert` and `insert_and_find` too? Brace-initialized group calls still fail deduction.

include/cuco/detail/open_addressing/open_addressing_ref_impl.cuh
```diff
@@ -406,6 +406,15 @@ class open_addressing_ref_impl
...
   }
 
+  /**
+   * @brief Inserts an element.
+   *
+   * @param value The element to insert.
+   *
+   * @return True if the given element is successfully inserted
+   */
+  __device__ bool insert(value_type value) noexcept { return insert<value_type>(value); }
```
PointKernel: These appear unnecessary: `static_map_ref` already passes a typed value that the existing implementation templates accept.

tests/static_map/insert_overload_test.cu
```diff
@@ -0,0 +1,99 @@
...
+  *found_value = iter->second;
+}
+
+TEMPLATE_TEST_CASE_SIG("static_map insert and insert_and_find value_type overloads",
+                       "",
+                       ((typename Key, typename Value), Key, Value),
+                       (int32_t, int32_t),
+                       (int32_t, int64_t),
+                       (int64_t, int32_t),
+                       (int64_t, int64_t))
```
PointKernel: Could we use one `int`/`int` case covering both APIs with CG sizes 1 and 2? Existing tests cover the width matrix. Direct result checks would also simplify this. Wider cases need the existing pre-Volta guard.

PointKernel (changes requested):
Please address the inline comments.

PointKernel (approved):
Looks great. Thanks @MaybeWilli for the contribution!

PR #845, opened 2026-09-16
Remove fast_int inheritance from dynamic valid extents

### LINKED ISSUE #838: [ENHANCEMENT]: Remove fast_int storage from dynamic valid extents
## Is your feature request related to a problem?

Blocked by #837.

Dynamic `valid_extent` currently inherits from `cuco::utility::fast_int`, storing the extent value plus precomputed division metadata.

After #837 removes modulo from probe advancement, the extent is no longer used as a runtime divisor. The additional state becomes unnecessary but is still constructed and copied through storage references and probing iterators.

Store only the underlying `SizeType` in dynamic `valid_extent`, preserving its existing value access, conversions, and required arithmetic operations.

The standalone public `cuco::utility::fast_int` utility must remain available and unchanged to avoid breaking its API.

MaybeWilli:
Fixes #838.

Removed dynamic valid_extent specialization, because as of #836 / #837, fast_int's modulo utility is no longer needed by valid_extent. Made valid_extent inherit from extent, allowing valid_extent to reuse extent's existing static/dynamic behavior. valid_extent.value() is inlined with no additional instructions (verified by PTX analysis). Changed struct valid_extent to class valid_extent.


include/cuco/detail/extent/extent.inl
```diff
@@ -110,12 +110,17 @@ struct valid_extent {
 };
 
 template <typename SizeType>
-struct valid_extent<SizeType, dynamic_extent> : cuco::utility::fast_int<SizeType> {
-  using value_type =
-    typename cuco::utility::fast_int<SizeType>::fast_int::value_type;  ///< Extent value type
+struct valid_extent<SizeType, dynamic_extent> {
```
PointKernel: Thanks for the cleanup! Could we also remove the unused `fast_int.cuh` include and the stale `TODO use fast_int operator` in `probing_scheme_impl.inl`? The PR description should reference #836 / #837 instead of #827.

include/cuco/detail/extent/extent.inl
```diff
@@ -85,37 +84,20 @@ struct valid_extent {
...
-  {
-    return lhs % rhs.value();
-  }
 };
 
 template <typename SizeType>
-struct valid_extent<SizeType, dynamic_extent> : cuco::utility::fast_int<SizeType> {
-  using value_type =
-    typename cuco::utility::fast_int<SizeType>::fast_int::value_type;  ///< Extent value type
+struct valid_extent<SizeType, dynamic_extent> {
```
sleeepyjack: I think we can simplify the implementation even more:
Now that the `fast_int` member has been removed, `valid_extent` is functionally the same as `extent`, but only allows for private construction through friend factories (`make_valid_extent`). There's also no longer a need for disallowing implicit conversions to `SizeType`. 

The cleanest option is `valid_extent` inheriting from the corresponding `extent`directly:
```cpp
template <typename SizeType, std::size_t N>
class valid_extent : public extent<SizeType, N> {
  using base_type = extent<SizeType, N>;

 public:
  using value_type = typename base_type::value_type;

  constexpr value_type value() const noexcept
  {
    return base_type::operator value_type();
  }

 private:
  explicit constexpr valid_extent(value_type value = {}) noexcept
    : base_type{value}
  {
  }

  // Factory friends
};
```

This would:

  - Delete the dynamic `valid_extent` specialization.
  - Reuse `extent`’s static/dynamic storage behavior.
  - Preserve private construction and explicit conversion.
  - Preserve empty-base optimization for static extents.
  - Keep the shared arithmetic operators unchanged.

